PR #32, opened 2020-11-26
Tom/add tutorial

TomWildenhain-Microsoft:


xadupre (approved):

wenbingl:
@TomWildenhain-Microsoft , can we create a simple example like only converting a existing tf-op without a onnx corresponding, and run with py-op? Then a comprehensive case of what you are doing.

BTW, can we remove the tensorflow CustomModule from the tutorial which looks irrelevant to ONNX custom op and it may mislead the audience the ONNX custom ops can only works with tf custom ops.

TomWildenhain-Microsoft:
@wenbingl , fair point about the CustomModule.  I'll remove it.  By "only converting a existing tf-op without a onnx corresponding, and run with py-op" do you mean what I was calling "Case 2" or "Case 1"?

wenbingl:
> @wenbingl , fair point about the CustomModule. I'll remove it. By "only converting a existing tf-op without a onnx corresponding, and run with py-op" do you mean what I was calling "Case 2" or "Case 1"?

Yes, it could be two cases, case 1 (simple): the missing tf-op already is implemented so there are only few lines to show tf2onnx and ort-customops could be called, case 2(full): a full example to demonstrate how to write the custom op with Python and C++.

TomWildenhain-Microsoft:
> Yes, it could be two cases, case 1 (simple): the missing tf-op already is implemented so there are only few lines to show tf2onnx and ort-customops could be called, case 2(full): a full example to demonstrate how to write the custom op with Python and C++.

Right.  I might be misunderstanding, but I think I already have that case specifically covered as "Case 1".  Do you want me to make it a separate notebook file?

wenbingl:
> > Yes, it could be two cases, case 1 (simple): the missing tf-op already is implemented so there are only few lines to show tf2onnx and ort-customops could be called, case 2(full): a full example to demonstrate how to write the custom op with Python and C++.
> 
> Right. I might be misunderstanding, but I think I already have that case specifically covered as "Case 1". Do you want me to make it a separate notebook file?

sorry, it's my fault and I review the case too fast, and we are on the same page. Just removing tf custommodule and everything will looks fine with me.

wenbingl (approved):

PR #42, opened 2021-01-18
Operator Schemas of text processing ops

Toudsour:
This PR includes:
 
- operator schema for text processing ops

We need a deep discussion of the operator schema. Any suggestion and issues are welcome~

### DIFF
--- a/docs/custom_text_ops.md
+++ b/docs/custom_text_ops.md
@@ -21,9 +21,9 @@
 
 |**Operator**|**Support State**|
 |------------|-----------------|
-|GPT2Tokenizer| Supported      |
-|WordPieceTokenizer| In next version |
-|BertTokenizer | In next version |
+|GPT2Tokenizer| Supported       |
+|BertTokenizer| In next version |
+|XLNetTokenizer| In next version |
 
 
 ### Decoder 
@@ -83,7 +83,7 @@ axes = np.array([0, 1], dtype=np.int64)
 steps = np.array([1, 1], dtype=np.int64)
 
 expect(node, inputs=[x, starts, ends, axes, steps], outputs=[y],
-       name='test_slice')
+       name='test_string_slice')
 ```
 </details>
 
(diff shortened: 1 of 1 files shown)


docs/custom_text_ops.md
```diff
@@ -0,0 +1,365 @@
...
+|XLNetTokenizer| In next version |
+
+
+## Auxiliary String Operator
+
+[TODO: Add existing operators]
+
+### <a name="StringSlice"></a><a name="StringSlice">**StringSlice**</a>
+Do the slice operation to each string element in input tensor. Similar to string slice in python
+```python
```
xadupre, line 36: Shoud we parse these examples to run all of them in a unit test?

Toudsour: I agree with you that those examples should be added to the unit test, but I think we do not need to parse them. The reason for parsing the doc is we are afraid of the consistency between op schema and our code. But in my point of view,  the schema should not be changed too often since it establishes.

I will add those examples into the unit test when I am implementing those processors.

docs/custom_text_ops.md
```diff
@@ -0,0 +1,365 @@
...
+|**Operator**|**Support State**|
+|------------|-----------------|
+|StringEqual |  Supported        |
+|StringHash  |  Supported        |
+|StringToHashBucketFast|Supported|
+|StringJoin  | Supported         |
+|StringRegexReplace| Supported  |
+|StringSplit | Supported       |
+|StringUpper  | Supported     |
+|StringSlice | In next version |
```
wenbingl, line 14: We don't set up the version yet, so 'in next version' could be 'ongoing'?

docs/custom_text_ops.md
```diff
@@ -0,0 +1,365 @@
...
+***data: tensor(string)***
+<dd>String tensor to extract slices from.</dd>
+
+***starts: tensor(int64/int32)***
+<dd>The tensor of starting indices of corresponding string in data, which has same dimension of data.</dd>
+
+***ends: tensor(int64/int32)***
+<dd>The tensor of ending indices of corresponding string in data, which has same dimension of data.</dd>
+
+***steps(optional): tensor(int64/int32)***
```
wenbingl, line 52: Due to ORT API limit, the optional input cannot be supported. Alternatively, we could assign a default value for the node.

docs/custom_text_ops.md
Toudsour: Sure, fixed.

wenbingl (approved):

PR #44, opened 2021-01-18
Run python unit tests on every platform + fix all string operators

xadupre:
Signed-off-by: xavier dupré <xavier.dupre@gmail.com>

### DIFF
--- a/ci_build/azure-pipelines/mshost.yaml
+++ b/ci_build/azure-pipelines/mshost.yaml
@@ -41,10 +41,10 @@ jobs:
   - script: python -m pip install -r requirements-dev.txt
     displayName: Install requirements-dev.txt
 
-  - script: python -u test/test_string_ops.py
+  - script: python -u test/test_string_ops.py --verbose
     displayName: Run test_string_ops.py
 
-  - script: python -u test/test_gpt2tok.py
+  - script: python -u test/test_gpt2tok.py --verbose
     displayName: Run test_gpt2tok.py
 
   - script: python -m pytest test
--- a/ci_build/azure-pipelines/mshost.yaml
+++ b/ci_build/azure-pipelines/mshost.yaml
@@ -54,28 +54,6 @@ jobs:
     displayName: Run test_string_ops.py
     continueOnError: True
 
-  - script: python -u -X faulthandler test/test_string_ops.py --verbose -k test_string_to_hash_bucket_cc
-    displayName: Run test_string_ops.py
-    continueOnError: True
-
-  - script: python -u -X faulthandler test/test_string_ops.py --verbose -k test_string_equal_cc
-    displayName: Run test_string_ops.py
-    continueOnError: True
-
-  - script: python -u -X faulthandler test/test_string_ops.py --verbose -k test_string_join_cc
-    displayName: Run test_string_ops.py
-    continueOnError: True
-
-  - script: python -u -X faulthandler test/test_string_ops.py --verbose -k test_string_split_cc
-    displayName: Run test_string_ops.py
-    continueOnError: True
-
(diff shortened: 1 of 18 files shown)


test/test_gpt2tok.py
```diff
@@ -10,7 +10,8 @@
     enable_custom_op,
     PyCustomOpDef,
     expand_onnx_inputs,
-    get_library_path as _get_library_path)
+    get_library_path as _get_library_path,
```
xadupre, line 13: remove

includes/onnxruntime/onnxruntime_cxx_api.h
```diff
@@ -599,6 +599,7 @@ struct CustomOpApi {
   OrtValue* KernelContext_GetOutput(OrtKernelContext* context, _In_ size_t index, _In_ const int64_t* dim_values, size_t dim_count);
 
   void ThrowOnError(OrtStatus* result);
+  const OrtApi& Api() { return api_; }
```
wenbingl, line 602: These headers are directly copied for ORT release, which suppose not be changed.

xadupre: Retrieving input string tensors is not easy. The function used for the other types fail on Linux, there is no KernelContext_* methods to do so. The only way is go handle them directly from the API. I need to raise an issue on onnxruntime. This is a temporary solution.

test/test_gpt2tok.py
xadupre: done

includes/onnxruntime/onnxruntime_cxx_api.h
xadupre: I removed it by using a trick which should be replaced by something cleaner when the C API is updated.

includes/ocos.h
```diff
@@ -6,6 +6,7 @@
 #define ORT_API_MANUAL_INIT
 #define EXCLUDE_REFERENCE_TO_ORT_DLL
 #include "onnxruntime_cxx_api.h"
+#include "onnxruntime_cxx_inline.h"
```
wenbingl, line 9: the *_inline.h is included at the footer of *_cxx_api.h, this change is unnecessary.

includes/onnxruntime/onnxruntime_cxx_inline.h
```diff
@@ -7,6 +7,8 @@
 // These are the inline implementations of the C++ header APIs. They're in this separate file as to not clutter
 // the main C++ file with implementation details.
 
+#pragma once
```
wenbingl, line 10: not needed.

includes/ocos.h
xadupre: done

includes/onnxruntime/onnxruntime_cxx_inline.h
xadupre: done

ci_build/azure-pipelines/mshost.yaml
```diff
@@ -31,26 +35,26 @@ jobs:
...
 
-  - script: |
-      python -m pip install -r requirements-dev.txt
+  - script: python -m pip install -r requirements-dev.txt
     displayName: Install requirements-dev.txt
 
-  - script: |
-      # FIXME: need check the CI environment for the failure.
-      # python -m pytest test
+  - script: python -m pytest test --verbose --verbose
```
wenbingl, line 44: two verbose?

wenbingl (approved):

includes/onnxruntime/onnxruntime_cxx_inline.h
xadupre: I prefer to keep it that way. When python crashes, there is no way to tell which test failed unless we keep that verbosity.

PR #56, opened 2021-02-03
Enable C++ end to end test with onnxruntime

wenbingl:
The change includes
1. Enable C++ end to end test with onnxruntime
2. Make all operator configurable in cmake option list
3. Some housekeeping works

BTW, C++ test end to end to checked in another PR which needs more tries on Azure Pipeline.


ci_build/azure-pipelines/mshost.yaml
```diff
@@ -26,15 +26,9 @@ jobs:
     displayName: Install requirements.txt
 
   - script: |
-      sh ./build.sh
       python setup.py develop
     displayName: Build the library and tests
 
-  - script: |
-      cd out/Linux
-      ./ortcustomops_test
```
xadupre, line 35: I understand that onnxruntime can run unit tests for the registered operators. What happens for the other unittests checking internal functions?

wenbingl: All unit tests static link to the internal function has been moved into ortcustomops_static_test. So there are two types of test, and the ortcustomops_test tests the whole thing in DLL with ONNXRuntime. So in CI pipeline, the test execution will be  ctest.

wenbingl:
@xadupre , any more comments, this PR needs to be merged as soon as possible to avoid the conflicts for any future PR.

xadupre:
> @xadupre , any more comments, this PR needs to be merged as soon as possible to avoid the conflicts for any future PR.

Looks good to me.

xadupre (approved):

PR #57, opened 2021-02-03
Add op: VectorToString

Toudsour:
- add VectorToString
- move utils-related file into ocos/kernels/utils


wenbingl:
The ocos/kernel will moved into the top directory soon and will be organized according to the operator category.
And ocos/pyfunc will be move into /onnxruntime_customops as well.
Then we can rename 'ocos' folder as 'util' if it was preferred.

Toudsour:
> The ocos/kernel will moved into the top directory soon and will be organized according to the operator category.
> And ocos/pyfunc will be move into /onnxruntime_customops as well.
> Then we can rename 'ocos' folder as 'util' if it was preferred.

Sure. I will do a refactor to those folders in the next PR because this change has already involved too many files.

Toudsour:
@wenbingl I can't open the detailed azure pipeline, it tells us `404 - Page not found`, the link:
https://aiinfra.visualstudio.com/04e34684-2e55-4d0f-858e-70a8fbfd3eca/_build/results?buildId=142923&view=logs&jobId=beebc4e0-08c3-587a-a9f8-8370dd8c77d0

Do you have any ideas?

wenbingl:
> @wenbingl I want to check why my pipeline failed but I found that I can't open the detailed azure pipeline, it tells me `404 - Page not found`, the link:
> https://aiinfra.visualstudio.com/04e34684-2e55-4d0f-858e-70a8fbfd3eca/_build/results?buildId=142923
> Do you have any ideas?

I have added your account in the member list, can you have a try?

Toudsour:
> > @wenbingl I want to check why my pipeline failed but I found that I can't open the detailed azure pipeline, it tells me `404 - Page not found`, the link:
> > https://aiinfra.visualstudio.com/04e34684-2e55-4d0f-858e-70a8fbfd3eca/_build/results?buildId=142923
> > Do you have any ideas?
> 
> I have added your account in the member list, can you have a try?

Succeeded.

ocos/kernels/vector_to_string.hpp
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+
+
+#include "kernels.h"
+#include "utils/string_utils.h"
+
+class VectorToStringImplBase
```
wenbingl, line 10: You are already using the template Kernel, I wasn't seeing this base classes is needed, especially the virtual compute method

tokenizer/gpt2tok.cc
```diff
@@ -20,7 +20,7 @@
 #include "nlohmann/json.hpp"
 #include "kernels/kernels.h"
 #include "kernels/string_common.h"
-#include "unicode.h"
+#include "utils/unicode.h"
```
wenbingl, line 23: this unicode source wasn't so common, and only be used in gpt2tokenizer.
Let's keep it where it was. We can move it until someday it will be reused by some other ops.

Toudsour: Sure.

ocos/kernels/vector_to_string.hpp
Toudsour: Yep, the virtual base classes is unnecessary, I used them because let Op class(non-template) could have a template class, do you have any ideas on this? Sorry for my poor C++ knowledge.

wenbingl: I didn't get your idea, why does the template has to be having a base class.
Just remove all staff of VectorToStringImplBase in your case.

test/static_test/test_base64.cc
```diff
@@ -3,7 +3,7 @@
 
 #include "gtest/gtest.h"
 #include "test_base64.h"
-#include "base64.h"
+#include "utils/base64.h"
```
wenbingl, line 6: keep where it was, unless the class is really be reused.
As we discussed before, the base64 is unnecessary and is harmful for mobile product.

Toudsour: Sure, I will move back them.

Toudsour: I delete it because base64 isn't used in our project.

Toudsour: I think we can delete base64, the only usage in our project is here
```c++
  if (base64_decode(model_as_string, model_as_bytes)) {
    model_proto.ParseFromArray(model_as_bytes.data(), model_as_bytes.size());
  } else {
    model_proto.ParseFromArray(model_as_string.c_str(), model_as_string.size());
  }
```

ocos/kernels/vector_to_string.hpp
Toudsour: I'm sorry I didn't explain myself clearly. 
I want VectorToStringImpl to support int64/float/int32 input. VectorToString needs a map<key_type,value_type> to store the mapping relationship, so in this situation, the key_type should be a template type instead of a certain type. But in C++, a non-template class could not own a template class as its member, the following code should be wrong:
```c++
class A {
   std::hash_map<T, std::string> map_;
}
```
So I created a base class and put the template class members into the template class which is inherited from the base class.
This's the way I could only think of, I don't know whether you have a better idea.


However, it doesn't matter, ort-customop can't support multi-type input now. I decided to only support int64 input now, then the base class could be removed.

wenbingl (approved):

PR #62, opened 2021-02-10
Add hook feature for the model debugging.

wenbingl:


xadupre:
If I summarize, you insert a python operator node before another one to have a look on the inputs. Is it possible to add some documentation in the function hook_model_op?

wenbingl:
> If I summarize, you insert a python operator node before another one to have a look on the inputs. Is it possible to add some documentation in the function hook_model_op?

Good reminder, without more python API, we should have them be documented.
I added the existing ones

xadupre (approved):

PR #63, opened 2021-02-10
Refactor C++ tests about operators

xadupre:


wenbingl:
Are you going to add more C++ test for the kernel?
The existing test cases and code come from the ONNXRuntime, and the purpose is to identify the issues caused the ONNXRuntime API unexpected change.

xadupre:
> Are you going to add more C++ test for the kernel?
> The existing test cases and code come from the ONNXRuntime, and the purpose is to identify the issues caused the ONNXRuntime API unexpected change.

I'd like to add more tests. Having a C++ code to test an operator is easier and it is possible to debug.

wenbingl (approved):

PR #64, opened 2021-02-10
Add StringLower operator

xadupre:
StringLower works on any string on Linux, MacOs. On Windows, every non ascii string still fails. onnxruntime may have to be compiled with option utf-8. Still under investigation.


wenbingl:
Need fix the test before I approve the PR.

xadupre:
> Need fix the test before I approve the PR.

The test is fixed.

CMakeLists.txt
```diff
@@ -87,6 +87,9 @@ if (OCOS_ENABLE_SPM_TOKENIZER)
   list(APPEND TARGET_SRC ${stpiece_TARGET_SRC})
 endif()
 
+
+add_compile_options("$<$<C_COMPILER_ID:MSVC>:/utf-8>")
```
wenbingl, line 91: Did you try to save the UTF-8 with BOM? MSVC should be OK with it then.

wenbingl (approved):

CMakeLists.txt
xadupre: The BOM was added to the file including chinese. The flag utf-8 is used to change the way strings are processed: https://docs.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-160. I may try other scenario if you prefer.

PR #65, opened 2021-02-16
[WIP] Remove StringLower, StringUpper, handles utf-8 strings with wstring

xadupre:
StringLower, StringUpper are already implemented in onnxruntime with operator StringNormalizer.
In the previous implementation, StringSplit was doing an ASCII split. However split `Ĭ,b` with delimiter `,` returns something like `['?', '', '', 'b']` because character `Ĭ` is a two bytes character ending by `,`. That's what this PR fixes.

This implementation also handles the case when the delimiter is a two bytes character, tensorflow does not. That may explain some discrepencies.


ocos/kernels/string_split.cc
```diff
@@ -13,9 +13,9 @@ void KernelStringSplit::Compute(OrtKernelContext* context) {
   const OrtValue* input_sep = ort_.KernelContext_GetInput(context, 1);
   const OrtValue* input_skip_empty = ort_.KernelContext_GetInput(context, 2);
   const bool* skip_empty = ort_.GetTensorData<bool>(input_skip_empty);
-  std::vector<std::string> X, sep;
-  GetTensorMutableDataString(api_, ort_, context, input_X, X);
-  GetTensorMutableDataString(api_, ort_, context, input_sep, sep);
+  std::vector<std::wstring> X, sep;
+  GetTensorMutableDataWString(api_, ort_, context, input_X, X);
+  GetTensorMutableDataWString(api_, ort_, context, input_sep, sep);
```
Toudsour, line 18: Does it mean users should input std::wstring instead of std::string?

xadupre: Not necessarily but it is safer to do it. It matters to split strings, it does not for join. Concatenating string or wstring is the same.

PR #67, opened 2021-02-17
Improve style in python file

xadupre:
Signed-off-by: xavier dupré <xavier.dupre@gmail.com>


wenbingl:
In the .flake8, the max line length is 120.
I don't think 80 is a right decision for this project, 80 is mostly good for the console script.

xadupre:
I usually use 80 because my screen is small. I'll discard the PR.

PR #68, opened 2021-02-18
Add StringLength op

Toudsour:
- Add StringLength op

### DIFF
--- a/ocos/kernels/string_length.cc
+++ b/ocos/kernels/string_length.cc
@@ -4,6 +4,7 @@
 #include "string_length.hpp"
 #include "string_common.h"
 #include <vector>
+#include <locale>
 #include <codecvt>
 #include <algorithm>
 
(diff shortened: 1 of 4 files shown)


ocos/kernels/string_length.cc
```diff
@@ -0,0 +1,51 @@
...
+
+
+KernelStringLength::KernelStringLength(OrtApi api) : BaseKernel(api) {
+}
+
+void KernelStringLength::Compute(OrtKernelContext* context) {
+  // Setup inputs
+  const OrtValue* input = ort_.KernelContext_GetInput(context, 0);
+  std::vector<std::string> input_data;
+  GetTensorMutableDataString(api_, ort_, context, input, input_data);
```
xadupre, line 19: You should use wstring if you want to get the number of characters (see PR #65) otherwise it is the number of bytes.

Toudsour: I got your concern, but wstring is encoding with UCS2 that doesn't support wider characters like emoji 😀.

wenbingl: Don't use wstring, which is inconsistent between WIN32 and Linux. And more important, most of input is UTF-8 and the corresponding type is string, instead of wstring.

wenbingl (approved):

PR #69, opened 2021-02-19
add unicode support by u32string

Toudsour:
- Wrap u32string with ustring, in case we will use other implementation in the future

### DIFF
--- a/ocos/kernels/string_common.cc
+++ b/ocos/kernels/string_common.cc
@@ -28,5 +28,27 @@ void FillTensorDataString(const OrtApi& api, Ort::CustomOpApi& ort, OrtKernelCon
   for (size_t i = 0; i < value.size(); ++i) {
     temp[i] = value[i].c_str();
   }
-  api.FillStringTensor(output, temp.data(), value.size());
+
+  Ort::ThrowOnError(api,api.FillStringTensor(output, temp.data(), value.size()));
+}
+
+void GetTensorMutableDataString(const OrtApi& api, Ort::CustomOpApi& ort, OrtKernelContext* context,
+                                 const OrtValue* value, std::vector<ustring>& output) {
+  std::vector<std::string> utf8_strings;
+  GetTensorMutableDataString(api, ort, context, value, utf8_strings);
+
+  output.reserve(utf8_strings.size());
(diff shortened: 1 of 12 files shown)


xadupre:
Do you think this approach is better than using std::wstring?

wenbingl:
> Do you think this approach is better than using std::wstring?

@xadupre 
check this answer: https://stackoverflow.com/questions/4588302/why-isnt-wchar-t-widely-used-in-code-for-linux-related-platforms

wenbingl:
The string tests is great to help the people to understand how to use u32string. 

but I didn't see too much benefit to create a new class ustring, instead of directly using std::u32string.

xadupre:
We should probably update StringNormalizer in onnxruntime https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/core/providers/cpu/nn/string_normalizer.cc#L22 as it is using wstring.

Toudsour:
> The string tests is great to help the people to understand how to use u32string.
> 
> but I didn't see too much benefit to create a new class ustring, instead of directly using std::u32string.

I worry that we will use ICU someday, so I want to wrap it instead of using std::u32string directly. ustring has all the operators of std::u32string, user can use them in the same way with u32string.

wenbingl:
> > The string tests is great to help the people to understand how to use u32string.
> > but I didn't see too much benefit to create a new class ustring, instead of directly using std::u32string.
> 
> I worry that we will use ICU someday, so I want to wrap it instead of using std::u32string directly. ustring has all the operators of std::u32string, user can use them in the same way with u32string.

that's a good justification.

wenbingl:
> We should probably update StringNormalizer in onnxruntime https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/core/providers/cpu/nn/string_normalizer.cc#L22 as it is using wstring.

Wasn't sure if they made the best decision. If it is possible for you check the tensorflow implementation?

wenbingl (dismissed):

Toudsour:
I am sorry that I don't know pushing a new commit require another approval, please help me approve again.

wenbingl (approved):

PR #71, opened 2021-03-02
Add design doc

Toudsour:


docs/design.md
```diff
@@ -0,0 +1,84 @@
+# ONNX Custom Ops Design
```
xadupre, line 1: ONNX Custom Ops or Onnxruntime Custom Ops?

Issue #74, opened 2021-03-08
PyOp only supports string attribute


Toudsour:
Recently, I add attribute `padding_length`(type: int64_t) for `GPT2Tokenizer`. Unittest of GPT2Tokenizer tests not only the C++ version but also implementation with PyOp:
```py
        @onnx_op(op_type="GPT2Tokenizer",
                 inputs=[PyCustomOpDef.dt_string],
                 outputs=[PyCustomOpDef.dt_int64, PyCustomOpDef.dt_int64],
                 attrs=["padding_length"])
        def bpe_tokenizer(s, **kwargs):
            # The user custom op implementation here.
            TestGPT2Tokenizer.pyop_invoked = True
            max_length = kwargs["padding_length"]
            input_ids = np.array(
                [TestGPT2Tokenizer.tokenizer(st_, max_length=max_length)["input_ids"] for st_ in s])
            attention_mask = np.array(
                [TestGPT2Tokenizer.tokenizer(st_, max_length=max_length)["attention_mask"] for st_ in s])
            return input_ids, attention_mask

```

When I run the test, I got the following error:

```
onnxruntime.capi.onnxruntime_pybind11_state.RuntimeException: [ONNXRuntimeError] : 6 : RUNTIME_EXCEPTION : Exception during initialization: Unable to find attribute 'padding_length' due to 'Attibute name and type don't match'.
```

After my investigation, I found the root of the error:
```c++
  PyCustomOpKernel(OrtApi api, const OrtKernelInfo* info, uint64_t id, const std::vector<std::string>& attrs)
      : api_(api),
        ort_(api_),
        obj_id_(id) {
    size_t size;
    for (std::vector<std::string>::const_iterator it = attrs.begin(); it != attrs.end(); ++it) {
      size = 0;
      OrtStatus* status = api_.KernelInfoGetAttribute_string(info, it->c_str(), nullptr, &size);
      if (api_.GetErrorCode(status) != ORT_INVALID_ARGUMENT) {
        std::string error_message(api_.GetErrorMessage(status));
        api_.ReleaseStatus(status);
        throw std::runtime_error(MakeString(
            "Unable to find attribute '", *it, "' due to '",
            error_message, "'."));
      }
      api_.ReleaseStatus(status);
      attrs_values_[*it] = "";
      attrs_values_[*it].resize(size);
      status = api_.KernelInfoGetAttribute_string(info, it->c_str(), &(attrs_values_[*it][0]), &size);
[...]

PR #75, opened 2021-03-08
Add batch query and attention mask support for GPT2Tokenizer

Toudsour:
GPT2Tokenizer:
-  support batch query
- support attention mask 
- support specifying padding length

Disable part of test due to #74


test/test_gpt2tok.py
```diff
@@ -4,7 +4,7 @@
 
 from pathlib import Path
 from onnx import helper, onnx_pb as onnx_proto
-from transformers import GPT2Tokenizer
+from transformers import *
```
wenbingl, line 7: Don't import everything, especially for this large package.

tokenizer/gpt2_tokenizer.hpp
```diff
@@ -0,0 +1,26 @@
+#include <kernels/ustring.hpp>
```
wenbingl, line 1: What's the idea of extracting this header file?

tutorials/tf2onnx_custom_ops_tutorial.ipynb
```diff
@@ -69,10 +69,10 @@
    "source": [
     "class Model1(tf.keras.Model):\n",
     "\n",
-    "    def __init__(self, name='model1', **kwargs):\n",
```
wenbingl, line 72: It's better to separate this change into another PR.

test/test_gpt2tok.py
```diff
@@ -23,90 +23,103 @@ def _get_test_data_file(*sub_dirs):
...
+        self._run_tokenizer(["#$%^&()!@?><L:{}\\[];',./`ǠǡǢǣǤǥǦǧǨ"])
+        self._run_tokenizer(["ڠڡڢڣڤڥڦڧڨکڪګڬڭڮگ"])
+        self._run_tokenizer(["⛀⛁⛂⛃⛄⛅⛆⛇⛈⛉⛊⛋⛌⛍⛎⛏"])
+        self._run_tokenizer(["I can feel the magic, can you?", "Yes I do."])
+        self._run_tokenizer(["I can feel the magic, can you?", "Yes I do."], 100)
+
+        enable_custom_op(True)
+
+
+# def test_tokenizer_pyop(self):
```
wenbingl, line 114: No pyop support any more?

test/test_gpt2tok.py
Toudsour: Sure, thanks.

tokenizer/gpt2_tokenizer.hpp
Toudsour: - Keep the same format with other ops
- Let GPT2Tokenizer could be registered by `operator_lists`
I noticed that GPT2Tokenizer is registered by `LoadTokenizerSchemaList`, should we keep this registration way?

tutorials/tf2onnx_custom_ops_tutorial.ipynb
Toudsour: Oh, those parts of code are changed by my IDE in a mistake, I will revert it.

test/test_gpt2tok.py
Toudsour: No, it's just disabled temporarily,  due to #74

tokenizer/gpt2_tokenizer.hpp
wenbingl: > * Keep the same format with other ops
> * Let GPT2Tokenizer could be registered by `operator_lists`
>   I noticed that GPT2Tokenizer is registered by `LoadTokenizerSchemaList`, should we keep this registration way?

The operator list will bring a lot of trouble if we want to use ifdef to enable/disable the operator because of the header file dependency. We should create some functions like LoadTokenizerSchemaList to decouple all operators.

Toudsour: I think header won't bring trouble because compiler will remove those function/class are not referred by the exported functions if we set right option.

```c++
#include "GPT2Tokenizer.hpp"

class GPT2Tokenizer m_gpt2;

auto register_list = [
#ifdef INCLUDE_GPT2
 &m_gpt2
#endif
]
```
If we turn off `INCLUDE_GPT2`, the whole GPT2Tokenizer will not be included into the final binary, even it's included in a cpp. Compiler will think class GPT2Tokenizer is useless and remove it freely, because there is not any import code(exported function) refer it.

wenbingl: GCC-liked library is a little complicated than you think. And In the most case, there are a lot of dependency file included in the header file, if there is a third party library, it will introduces a lot of ifdef in the file.

So the load function is the simple way to do it and prevent the user from missing ifdef in the file. And this library, there is no need to let the operator referenced by each other, and no need for the header file at all.

Toudsour: I move the content of header into cpp.

wenbingl (approved):

wenbingl (approved):

PR #76, opened 2021-03-10
Support the pre/post processing Python and pytorch code snippets to be converted and be merged into the ONNX model.

wenbingl:


wenbingl:
@jiafatom , can you please the implementation of the pytorch function call to onnx conversion?

examples/gpt2_e2e.py
```diff
@@ -0,0 +1,120 @@
+import os
```
xadupre, line 1: We should add some comments to help users understand what the example illustrates.

examples/gpt2_e2e.py
```diff
@@ -0,0 +1,120 @@
...
+import numpy
+from transformers import AutoConfig
+from onnxruntime_customops import mytorch as torch
```
xadupre, line 4: I would not add an alias `torch`, it is difficult to know what's coming from torch or from customops.

examples/gpt2_e2e.py
```diff
@@ -0,0 +1,120 @@
...
+output_ms = inference_and_dump_full_model(input_text)
+
+# 3. Inference on the all-in-one model
+full_model = torch.op_from_model(gpt2_full_model_path)
+outputs = full_model(input_text)
+
+# 4. Test the result
+for n_, tx_ in enumerate(outputs):
+    if output_ms[n_] != outputs[n_]:
+        import warnings  # noqa
```
xadupre, line 118: Move to the top.

examples/gpt2_e2e.py
```diff
@@ -0,0 +1,120 @@
...
+        os.makedirs(cache_dir)
+    return cache_dir
+
+
+def convert_models():
+    from transformers import GPT2Tokenizer  # noqa
+    from onnxruntime.transformers.gpt2_helper import Gpt2Helper, MyGPT2LMHeadModel  # noqa
+    cache_dir = get_cache_directory()
+
+    tokenizer = GPT2Tokenizer.from_pretrained(model_name_or_path, cache_dir=cache_dir)
```
xadupre, line 32: I would move these global variables as argument to the function, it is easier to reuse.

onnxruntime_customops/_cuops.py
```diff
@@ -0,0 +1,95 @@
...
+                                     domain=default_opset_domain())
+        graph = onnx.helper.make_graph([cuop],
+                                       "og_{}_{}".format(op_type, cls.get_next_id()),
+                                       inputs,
+                                       outputs)
+        return graph
+
+    @staticmethod
+    def get_op_class(op_type):
+        return globals()[op_type]
```
xadupre, line 95: I would avoid the use of globals.

onnxruntime_customops/_ocos.py
```diff
@@ -65,6 +65,19 @@ def create(func, *args, **kwargs):
         add_custom_op(opdef._nativedef)
         return opdef
 
+    @classmethod
+    def get_opdef(cls, op_type):
```
xadupre, line 69: If there are many operators and if the method is called many times, it should be optimized with a dictionary.

onnxruntime_customops/_ocos.py
```diff
@@ -157,16 +173,16 @@ def hook_model_op(model, node_name, hook_func):
     del hkd_model.graph.node[:]
     hkd_model.graph.node.extend(repacked)
 
-    value_infos = list(hkd_model.graph.value_info)
-    type_dict = {k_: None for k_ in input_names}
-    for vi_ in value_infos:
-        if vi_.name in type_dict:
-            type_dict[vi_.name] = vi_.type.tensor_type.elem_type
+    # value_infos = list(hkd_model.graph.value_info)
```
xadupre, line 176: To be removed later?

onnxruntime_customops/_ortcustomops.pyi
```diff
@@ -0,0 +1,15 @@
+
```
xadupre, line 1: Is this file needed?

examples/gpt2_e2e.py
wenbingl: Yes, there will be some document talking about it.

examples/gpt2_e2e.py
```diff
@@ -0,0 +1,120 @@
...
+
+def get_cache_directory():
+    cache_dir = os.path.join("../..", "cache_models")
+    if not os.path.exists(cache_dir):
+        os.makedirs(cache_dir)
+    return cache_dir
+
+
+def convert_models():
+    from transformers import GPT2Tokenizer  # noqa
```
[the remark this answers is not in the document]
wenbingl: This tutorial script, which is logic clear is more important than the style.

examples/gpt2_e2e.py
wenbingl: This tutorial script, which is logic clear is more important than the style.

onnxruntime_customops/_ortcustomops.pyi
wenbingl: Yes, it's very important for C++ implemented function and class on the type verification.

onnxruntime_customops/mytorch/__init__.py
```diff
@@ -0,0 +1,41 @@
+"""
```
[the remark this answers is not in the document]
wenbingl: It's a good thinking, but I don't expect these code will be re-used by any converter. And if it will, we can do it later. Right now, putting all pieces together will be more convenient.

onnxruntime_customops/_cuops.py
wenbingl: this 'globals' is not a whole package, it's only the current file/module. And it's a good way to reflect the class, without a dictionary.

onnxruntime_customops/_ocos.py
wenbingl: not used now, removed them

onnxruntime_customops/_ocos.py
wenbingl: removed.

onnxruntime_customops/_ocos.py
wenbingl: not used now, removed.

onnxruntime_customops/_cuops.py
wenbingl: This globals only is only the scope of the current file. And this file is to be expects only contains the custom op classes.

jiafatom (dismissed):
Please address the nits.

PR #78, opened 2021-04-07
Create webassembly pipeline 

Toudsour:
- Create webassembly pipeline

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "cmake/deps/emsdk"]
+	path = cmake/deps/emsdk
+	url = https://github.com/emscripten-core/emsdk.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,0 @@
-[submodule "cmake/deps/emsdk"]
-	path = cmake/deps/emsdk
-	url = https://github.com/emscripten-core/emsdk.git
-
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "cmake/deps/emsdk"]
 	path = cmake/deps/emsdk
 	url = https://github.com/emscripten-core/emsdk.git
+
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -223,6 +223,7 @@ include(CPack)
 # test section
 if (OCOS_ENABLE_CTEST)
   # Enable CTest
+  find_library(ONNXRUNTIME onnxruntime HINTS "${ONNXRUNTIME_LIB_DIR}")
   if (NOT ONNXRUNTIME)
     message(FATAL_ERROR "Enable ctest should provide onnxruntime path, please specify the ONNXRUNTIME_LIB_DIR.")
   endif()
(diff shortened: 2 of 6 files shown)


.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "cmake/deps/emsdk"]
```
wenbingl, line 1: I intentionally avoid the git submodule to decrease this project complexity.
Can we pass the emsdk DIR in the build command argument?

ci_build/azure-pipelines/mshost.yaml
```diff
@@ -3,227 +3,269 @@
 #######
 
 jobs:
-- job: Linux
```
wenbingl, line 6: why is every line changed? it's hard to review

.gitmodules
Toudsour: Sure, we can download emsdk in the pipeline. BTW, I am curious why git submodule will increase complexity.

ci_build/azure-pipelines/mshost.yaml
Toudsour: I format this file, only added lines: 230-271

CMakeLists.txt
```diff
@@ -15,15 +15,13 @@ include(CheckLanguage)
...
 option(OCOS_ENABLE_PYTHON "Enable Python component building" OFF)
+option(OCOS_ENABLE_CTEST "Enable C++ test" ON)
 option(OCOS_ENABLE_TF_STRING "Enable String Operator Set" ON)
 option(OCOS_ENABLE_GPT2_TOKENIZER "Enable the GPT2 tokenizer building" ON)
 option(OCOS_ENABLE_SPM_TOKENIZER "Enable the SentencePiece tokenizer building" ON)
 option(OCOS_ENABLE_BERT_TOKENIZER "Enable the BertTokenizer building" ON)
 
-find_library(ONNXRUNTIME onnxruntime HINTS "${ONNXRUNTIME_LIB_DIR}")
-if ((NOT OCOS_ENABLE_PYTHON) AND (NOT ONNXRUNTIME))
-  message(FATAL_ERROR "Cannot find onnxruntime in the default library paths, please specify the ONNXRUNTIME_LIB_DIR.")
```
wenbingl, line 25: This is a pretty useful error message, is it replaced by some other message?

CMakeLists.txt
```diff
@@ -98,6 +99,7 @@ endif()
 add_compile_options("$<$<C_COMPILER_ID:MSVC>:/utf-8>")
 add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/utf-8>")
 add_library(ortcustomops_static STATIC ${TARGET_SRC})
+target_include_directories(ortcustomops_static PUBLIC tokenizer)
```
wenbingl, line 102: so the tokenizer cannot be disabled in the compiling now?

CMakeLists.txt
Toudsour: Actually, we check the onnxruntime library only when we build test, so I move them to line:226 after `OCOS_ENABLE_CTEST` condition

CMakeLists.txt
Toudsour: Emm... We can discuss this more in the meeting. You can refer this question:[How to include all objects of an archive in a shared object?](https://stackoverflow.com/questions/2763988/how-to-include-all-objects-of-an-archive-in-a-shared-object?rq=1)

wenbingl (approved):

Toudsour:
@wenbingl, there are some failed tests due to Eager model. After investigation, I can't figure out the root cause. (Maybe it's related to the new version of onnx -- 1.9)

PR #81, opened 2021-04-12
Fixes onnxruntime C API 5 forward compatibility

xadupre:
#80 does not work with onnxruntime 1.7.0.

### DIFF
--- a/includes/onnxruntime/onnxruntime_cxx_inline.h
+++ b/includes/onnxruntime/onnxruntime_cxx_inline.h
@@ -869,8 +869,10 @@ inline std::string CustomOpApi::KernelInfoGetAttribute<std::string>(_In_ const O
   OrtStatus* status = api_.KernelInfoGetAttribute_string(info, name, nullptr, &size);
 
   // The status should be ORT_INVALID_ARGUMENT because the size is insufficient to hold the string
-  if (api_.GetErrorCode(status) == ORT_INVALID_ARGUMENT) {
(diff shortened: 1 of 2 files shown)


includes/onnxruntime/onnxruntime_cxx_inline.h
```diff
@@ -869,8 +869,10 @@ inline std::string CustomOpApi::KernelInfoGetAttribute<std::string>(_In_ const O
   OrtStatus* status = api_.KernelInfoGetAttribute_string(info, name, nullptr, &size);
 
   // The status should be ORT_INVALID_ARGUMENT because the size is insufficient to hold the string
-  if (api_.GetErrorCode(status) == ORT_INVALID_ARGUMENT) {
-    api_.ReleaseStatus(status);
+  if (status == nullptr || api_.GetErrorCode(status) == ORT_INVALID_ARGUMENT) {
```
wenbingl, line 872: What's the case of status == nullptr? it happens on ONNXRuntime or this custom op library?

xadupre: It is like that in the current version of onnxruntime https://github.com/microsoft/onnxruntime/blob/master/include/onnxruntime/core/session/onnxruntime_cxx_inline.h#L889. But replacing these files by the new ones does not work with older version of onnxruntime (see #80).

wenbingl (approved):

PR #91, opened 2021-05-11
Fix an index error in KernelWordpieceTokenizer

xadupre:


TomWildenhain-Microsoft (approved):

xadupre:
> It looks the CI pipeline takes long now, should we move the BuildWheels into another pipeline.

We could also remove the job for python 3.6. Overall, it takes 48 minutes right now.

PR #107, opened 2021-06-08
Update patch file to integrate CustomOps into ONNXRuntime.

Zuwei-Zhao:
Update patch file to integrate CustomOps into ONNXRuntime.

### DIFF
--- a/ci_build/onnxruntime_integration/build_with_onnxruntime.sh
+++ b/ci_build/onnxruntime_integration/build_with_onnxruntime.sh
@@ -11,8 +11,8 @@ cd $BASE_DIR
 git clone https://github.com/microsoft/onnxruntime.git
 cd onnxruntime
 
-# there is no stable version including webassembly
-git checkout e6a3308db7c03a13e0f08b221b6770e17fc3a4ef
+# The latest commit of ONNXRuntime that has been verified for integration (#7941)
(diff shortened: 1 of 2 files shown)


ci_build/onnxruntime_integration/build_with_onnxruntime.sh
```diff
@@ -11,8 +11,8 @@ cd $BASE_DIR
 git clone https://github.com/microsoft/onnxruntime.git
 cd onnxruntime
 
-# there is no stable version including webassembly
-git checkout e6a3308db7c03a13e0f08b221b6770e17fc3a4ef
+# The latest commit of ONNXRuntime that has been verified for integration (#7941)
```
wenbingl, line 14: Can we use the rel-1.8.0 branch for this patch, which looks more stable?

Zuwei-Zhao: Hi, I found that branch: **rel-1.8.0** doesn't contain **string tensor support** in onnxruntime web, which is required in our scenario. So from my perspective, the latest commit would be better. Thanks.

wenbingl: in that case, let's goes the latest.

wenbingl (approved):

PR #129, opened 2021-08-17
Replace Re2 with std::re2

Toudsour:
- Replace Re2 with std::re2


snnn:
Would you like to help us change the source code in https://github.com/Microsoft/onnxruntime too?

Toudsour:
> Would you like to help us change the source code in https://github.com/Microsoft/onnxruntime too?

You mean replacing re2 with std::regex?

snnn:
Yes.

wenbingl:
> Yes.

Do you know which standard op needs google::re2?
if it's a contrib op, it suppose to be moved into here.

operators/text/string_regex_replace.cc
```diff
@@ -5,7 +5,11 @@
 #include <vector>
 #include <cmath>
 #include <algorithm>
+#ifdef ENABLE_RE2
```
wenbingl, line 8: I guess the reason why we replace it with std::regex is to save some binary size. 
In that case, if this operator can be fully implemented by std::regex, then we do it by std::regex.
Otherwise, the operator has to depend on google::re2. so we can remove this kind of ifdef/endif

snnn:
https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/contrib_ops/cpu/tokenizer.cc

wenbingl:
> https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/contrib_ops/cpu/tokenizer.cc
Not sure if it is used by some products, or it should be deprecated, since in my understanding, there is no generic tokenizer.

std::regex and google::re2 is difference in the syntax of regex.

snnn:
But if we can reduce a dependency ...

PR #133, opened 2021-08-20
Create wheels.yml

wenbingl:


.github/workflows/wheels.yml
```diff
@@ -0,0 +1,38 @@
...
+on:
+  schedule:
+    # Run weekly on Monday 00:00
+    - cron:  '00 00 * * MON'
+  push:
+    branches: [rel-*]
+  pull_request:
+    branches: [rel-*]
+  workflow_dispatch:
+
```
snnn, line 12: It is so nice and simple. I hope onnxruntime's packaging pipeline could be so simple.

PR #135, opened 2021-08-24
Add BertTokenizer

Toudsour:
- Add op BasicTokenizer, BertTokenizer, same interface with hugging face.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,7 +145,7 @@ endif()
 
 if (OCOS_ENABLE_BERT_TOKENIZER)
   # Bert
-  file(GLOB bert_TARGET_SRC "operators/tokenizer/wordpiece*.*")
+  file(GLOB bert_TARGET_SRC "operators/tokenizer/wordpiece*.*" "operators/tokenizer/basic_tokenizer.*")
   list(APPEND TARGET_SRC ${bert_TARGET_SRC})
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,7 +145,7 @@ endif()
 
 if (OCOS_ENABLE_BERT_TOKENIZER)
   # Bert
-  file(GLOB bert_TARGET_SRC "operators/tokenizer/wordpiece*.*" "operators/tokenizer/basic_tokenizer.*")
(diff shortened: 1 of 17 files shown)


operators/string_utils.cc
```diff
@@ -32,6 +32,35 @@ std::vector<std::string_view> SplitString(const std::string_view& str, const std
   return result;
 }
 
+bool IsChineseChar(char32_t c) {
```
wenbingl, line 35: CJK?

operators/ustring.h
```diff
@@ -25,6 +25,10 @@ class ustring : public std::u32string {
   explicit ustring(const std::u32string_view& str);
   explicit ustring(const std::u32string_view&& str);
 
+  ustring(ustring& str);
```
wenbingl, line 28: is it necessary?

wenbingl (dismissed):

CMakeLists.txt
```diff
@@ -146,7 +146,7 @@ endif()
 
 if (OCOS_ENABLE_BERT_TOKENIZER)
   # Bert
-  file(GLOB bert_TARGET_SRC "operators/tokenizer/wordpiece*.*")
+  file(GLOB bert_TARGET_SRC "operators/tokenizer/wordpiece*.*" "operators/tokenizer/basic_tokenizer.*" "operators/tokenizer/bert_tokenizer.*")
```
Zuwei-Zhao, line 149: Hi, what's the difference between new bert tokenizer and existing wordpiece tokenizer? Should they both be built using one flag?

tools/gen_selectedops.py
```diff
@@ -61,7 +61,7 @@ def gen_cmake_oplist(opconfig_file, oplist_cmake_file='_selectedoplist.cmake'):
     if len(sys.argv) == 2:
```
Zuwei-Zhao, line 61: Please remember to add a mapping to `OPMAP_TO_CMAKE_FLAGS` in this file because we will select operators to use when build together with onnxruntime.

CMakeLists.txt
Toudsour: BertTokenizer consist of BasicTokenizer and WordPieceTokenizer, here I will split them.

PR #143, opened 2021-09-02
Disable c++ exceptions in onnxruntime-extensions.

Zuwei-Zhao:
Disable c++ exceptions in onnxruntime-extensions, which will **reduce binary size**.


CMakeLists.txt
```diff
@@ -79,6 +80,27 @@ if (OCOS_ENABLE_SELECTED_OPLIST)
   message(STATUS "Looking for the _selectedoplist.cmake")
   disable_all_operators()
   include(_selectedoplist)
+
+  # Reset exceptions flag for specific operators,
+  # would remove this resetting later when we have supported no-exceptions for these operators.
+  if (OCOS_ENABLE_BLINGFIRE OR OCOS_ENABLE_SPM_TOKENIZER OR OCOS_ENABLE_WORDPIECE_TOKENIZER)
```
wenbingl, line 86: We should only enable c++ exception for the third party library only, instead of silently turning on this flag for the whole package.

Zuwei-Zhao: Hi @wenbingl , I have removed the logic here. I verified that we don't need to set additional exceptions flag for blingfire because it supports exceptions by default. So we only need to disable exceptions for other operators and no need to modify blingfire build options. Please help confirm. Thanks.

wenbingl (approved):

PR #147, opened 2021-09-09
Fix op selection

Toudsour:
SegmentExtraction was put into Math operator set, so the build flag for SegmentExtraction should be `OCOS_ENABLE_MATH` 
Inverse op need dlib but dlib cannot disable exception, so set a flag for inverse operator.


CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ option(OCOS_ENABLE_SPM_TOKENIZER "Enable the SentencePiece tokenizer building" O
 option(OCOS_ENABLE_WORDPIECE_TOKENIZER "Enable the WordpieceTokenizer building" ON)
 option(OCOS_ENABLE_BERT_TOKENIZER "Enable the BertTokenizer building" ON)
 option(OCOS_ENABLE_BLINGFIRE "Enable the Blingfire building" ON)
+option(OCOS_ENABLE_INVERSE "Enable the Inverse building" ON)
```
wenbingl, line 35: Similar to use re2_regex, this option could be OCOS_ENABLE_DLIB

Toudsour: OK

wenbingl (approved):

PR #148, opened 2021-09-09
Add new operator StringMapping

Toudsour:
- add new operator StringMapping


onnxruntime_extensions/_cuops.py
```diff
@@ -61,6 +61,32 @@ def serialize_attr(cls, attrs):
         for k_, v_ in attrs.items():
             if k_ == 'map' and isinstance(v_, dict):
                 attr_data[k_] = '\n'.join(k + "\t" + " ".join([str(i) for i in v]) for k, v in v_.items())
+            elif k_ == 'map' and isinstance(v_, str):
+                with open(v_) as map_file:
```
wenbingl, line 65: why does it have to read the file here, instead a real string, which seems to be more consistent to the native OP implementation.

Toudsour: I think putting a raw vocab string into function does not look like a 'Python' way, but we can still provide them this way.

Toudsour: I add `map_file` parameter for reading map from file. And `map` parameter can accept a read string.

wenbingl: I don't think it's a Pythonic way. The problem is to keep an consistent way with other case. If the op is in a ONNX model file, it will bring unnecessary complexity about  shipping an additional file for the model file.

wenbingl: > I add `map_file` parameter for reading map from file. And `map` parameter can accept a read string.

don't do that, it is unnecessary to support extra file in file, I don't see the requirement.
ONNX actually supports an additional file storing the tensor, in case there is a very large. But obviously, this is not the case.

Toudsour: Removed `map_file`.

wenbingl (approved):

PR #154, opened 2021-09-26
Reduce memory usage

Toudsour:
- Replace the memory through replacing ustring by std::string_view in BertTokenizerDecoder


wenbingl (approved):
so only en-US ansi char can be supported now?

Toudsour:
> so only en-US ansi char can be supported now?

No, it still support Unicode through UTF-8

PR #165, opened 2021-10-04
Correct memory indexing issue.

joburkho:
`KernelBlingFireSentenceBreaker::Compute(OrtKernelContext* context)` from operators/tokenizer/blingfire_sentencebreaker.cc currently contains the following code:
```
  int max_length = 2 * input_string.size() + 1;
  std::string output_str;
  output_str.reserve(max_length);
```
... however, `std::string::reserve` only reserves memory and does not resize `std::string` so the use of `output_str.data()` and the use of `output_str[i]` that follow are not defined ... and in a Debug build where bounds checking based on the size of `std::string` is done, `output_str[i]` fails.

Either of the following can be used instead:
```
  std::string output_str;
  output_str.resize(max_length, '\0'); // or just output_str.resize(max_length); since '\0' is implied.
```
... or:
```
std::unique_ptr<char[]> output_str = std::make_unique<char[]>(max_length);
```
... with `output_str.get()` used in the place of `output_str.data()` where needed.

This PR uses the later since it retains the semantics of how `output_str` was being used inside `KernelBlingFireSentenceBreaker::Compute(OrtKernelContext* context)`.

### DIFF
--- a/ci_build/azure-pipelines/mshost.yaml
+++ b/ci_build/azure-pipelines/mshost.yaml
@@ -129,7 +129,7 @@ jobs:
 
   - job: MacOSWheel
     pool:
-      vmImage: 'macOS-latest'
+      vmImage: 'macOS-10.15'
 
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


wenbingl (dismissed):
Thanks for the fixing

wenbingl (approved):

PR #174, opened 2021-10-20
Fix ::tolower error when locale is not 'C'

Toudsour:
- Fix ::tolower error when locale is not 'C'
- Add more test for BertTokenizer

### DIFF
--- a/test/shared_test/test_ortops_tokenizer.cc
+++ b/test/shared_test/test_ortops_tokenizer.cc
@@ -242,7 +242,8 @@ TEST(tokenizer_opertors, test_bert_tokenizer_scalar) {
   model_path /= "test_bert_tokenizer_scalar.onnx";
   TestInference(*ort_env, model_path.c_str(), inputs, outputs, GetLibraryPath());
 
-  std::locale("en_US.UTF-8");
+  // change locale to system locale
+  std::locale();
 
   inputs[0].name = "text";
   inputs[0].element_type = ONNXTensorElementDataType::ONNX_TENSOR_ELEMENT_DATA_TYPE_STRING;


operators/string_utils.cc
```diff
@@ -98,10 +98,23 @@ bool IsAccent(char32_t c)
   return c >= 0x300 && c <= 0x36F;
 }
 
+// only support latin now
+char32_t ToLower(char32_t c) {
```
snnn, line 102: Why do we need  to implement this by ourself? 
And in which case we are using UTF-32?  I thought our code is mostly based on UTF-8 and UTF-16.

Toudsour: This is because ::tolower will be affected by the locale. When locale is not 'C' then ::tolower will translate non-ASCII(such as chinese) character to strange character. Onnxruntime-extensions is not only targeted for server-side but also for client-side. In client-side we can not guarantee the locale must be 'C'.

Tokenizers need support multi-language, so something we need to remove a non-ASCII character from input string. At this time, UTF-32 is a better choice.

snnn: This is fundamentally wrong. Either you use UTF-8, or you use system locale(like mbstowcs) , or you use a popular open source library like ICU/iconv.  Chinese language has no directly relation to UTF-32. UTF-32 is never in the picture.  If you use UTF-32, how could the code work on Windows? It's a bomb.

wenbingl: Do you mean some functions like this
https://en.cppreference.com/w/cpp/locale/ispunct
?
It looks it fits the need if the locale is known.
But I heard there are some issues on the locale C++ implementation on Windows.

wenbingl: > Yes. Things like this. You can always pass a locale object in. Nothing is perfect. But we have use it for a long time in onnxruntime, and we run the tests every day. So I think it won't be a big problem.

Unfortunately, these functions sometimes returns different result in different platforms with the same locale. It's not a reliable APIs that can depends on.

The robust solution is the Unicode category table which checked in the same folder. but the table takes some binary size.

PR #182, opened 2021-11-02
Add MaskedFill

Toudsour:
- add new a operator --- MaskedFill


includes/ocos.h
```diff
@@ -60,6 +60,18 @@ struct OrtTensorDimensions : std::vector<int64_t> {
   bool IsScalar() const{
     return empty();
   }
+
+  bool IsVector() const{
+    if (size() == 1) {
+      return true;
+    }
+
+    if ((size() == 2) && (*begin() == 1)) {
```
joburkho, line 69: Is a batch parameter not supported?

operators/text/string_remove.cc
```diff
@@ -0,0 +1,81 @@
...
+void KernelStringRemove::Compute(OrtKernelContext* context) {
+  // Setup inputs
+  const OrtValue* input_string = ort_.KernelContext_GetInput(context, 0);
+  const OrtValue* input_condition = ort_.KernelContext_GetInput(context, 1);
+
+  OrtTensorDimensions string_dimensions(ort_, input_string);
+  OrtTensorDimensions condition_dimensions(ort_, input_condition);
+
+  if ((!(string_dimensions.IsScalar() || string_dimensions.IsVector())) ||
+      (!(condition_dimensions.IsVector() || condition_dimensions.IsVector()))) {
```
joburkho, line 24: Why is this `condition_dimensions.IsVector() || condition_dimensions.IsVector()` instead of `condition_dimensions.IsScalar() || condition_dimensions.IsVector()`?

includes/ocos.h
```diff
@@ -60,6 +60,18 @@ struct OrtTensorDimensions : std::vector<int64_t> {
   bool IsScalar() const{
     return empty();
   }
+
+  bool IsVector() const{
```
wenbingl, line 64: it's a little weird of this method that the vector can be 2d, though first dim is 1.

onnxruntime_extensions/_cuops.py
```diff
@@ -90,6 +90,17 @@ def serialize_attr(cls, attrs):
         return attr_data
 
 
+class StringRemove(CustomOp):
```
wenbingl, line 93: can you update the doc here: https://github.com/microsoft/onnxruntime-extensions/blob/main/docs/custom_text_ops.md ?

test/test_string_remove.py
```diff
@@ -0,0 +1,45 @@
...
+    test_dir = Path(__file__).parent
+    return str(test_dir.joinpath(*sub_dirs))
+
+
+def read_file(path):
+    with open(path) as file_content:
+        return file_content.read()
+
+
+def _run_string_remove(strings, conditions, output):
```
wenbingl, line 17: no comma needed here?

onnxruntime_extensions/_cuops.py
wenbingl, line 93: Do you know if there is any similar op in other framework/library that can be a reference?

includes/ocos.h
Toudsour: Because we usually represent vector as [n] or [1,n]. It rarely meets [1,n,1] or [1, 1, n].
Currently, `IsVector` only supports `[n]` format.

operators/text/string_remove.cc
Toudsour: Fixed, thank you for pointing out this bug.

includes/ocos.h
Toudsour: Sure, currently,` IsVector` only supports `[n]` format.

onnxruntime_extensions/_cuops.py
Toudsour: Added

test/test_string_remove.py
Toudsour: Removed

onnxruntime_extensions/_cuops.py
Toudsour: Renamed to MaskedFill

wenbingl (approved):

PR #183, opened 2021-11-04
Add android pipeline

Toudsour:
- Add Android build pipeline
- Add op_selection for MaskedFill
- Fix tensorflow_text version error


tools/utils/__init__.py
```diff
@@ -0,0 +1,6 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
wenbingl, line 1: Will it be better to move this package into onnxruntime_extensions folder to be an android sub-package?

test/shared_test/test_ortops_strings.cc
```diff
@@ -292,10 +283,7 @@ TEST(utils, test_string_join) {
   outputs[0].dims = {1};
   outputs[0].values_string = {"abc-zzz-efg"};
 
-  std::filesystem::path model_path = __FILE__;
```
wenbingl, line 295: android NDK doesn't support __FILE__ or the file cannot be accessed in the test app?

Toudsour: The model file should be accessed by relative path instead of absolute path, because the data file will be pushed to Android emulator and we can't keep the same path. So I remove `__FILE__`.

tools/utils/__init__.py
Toudsour: Those file is copied from `onnxruntime` repo and aim to run Android emulator. Currently, we need to build onnxruntime by ourself, so I leverage files in `onnxruntime` repo and delete those files.

requirements-dev.txt
```diff
@@ -1,3 +1,3 @@
 pytest
-transformers
-tensorflow_text
+transformers  == 4.9.2
```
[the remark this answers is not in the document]
Toudsour: Fixed

wenbingl (approved):

PR #193, opened 2022-01-14
[ONNX RUNTIME] [TVM EP]: rename STVM to TVM

vvchernov:
This PR was initiated by [PR](https://github.com/microsoft/onnxruntime/pull/10260) from onnx runtime. They should be approved and merged synchroniously.


vvchernov:
I observed that it is patch and this fix does not matter

PR #215, opened 2022-03-25
gpt2 end-end pre-processing script fixing with pytorch 1.11.

wenbingl:


snnn (approved):

pyproject.toml
```diff
@@ -1,4 +1,5 @@
 [project]
+name = "onnxruntime_extensions"
 # since onnxruntime havn't supported Python 3.10 yet
 requires-python = "<3.10"
```
snnn, line 4: onnxruntime has a 3.10 nightly package at https://pypi.org/project/ort-nightly/ .  It may be used for testing your 3.10 packages.

wenbingl: thanks for the info. it's not urgent and we may wait for the official release.

Issue #227, opened 2022-05-06
Building ORT with extensions fails memory leak test


shalvamist:
Hey,

I ran several build configurations on onnx and found that when adding extensions to the build one of the built-in tests will fail on memory leak.

Here are the configuration I tested - 
1. Vanila repo without extensions - CL - _$> build_ - This PASSED all tests
2. Vanila repo with extensions - CL - _$> build –-use_extensions_ - This failed on the memory leak
    Error - 
    "
    1: ----- MEMORY LEAKS: 656 bytes of memory leaked in 8 allocations
    1/3 Test #1: onnxruntime_test_all .............***Failed  336.43 sec
    "
3. Vanila repo with latest extensions repo - CL - _$> build –-use_extensions --extensions_overridden_path <path to the extension repo>_ - This failed on the memory leak as well
    Error - 
    "
    1: ----- MEMORY LEAKS: 656 bytes of memory leaked in 8 allocations
    1/3 Test #1: onnxruntime_test_all .............***Failed  336.43 sec
    "

Is this the expected behavior ?

PR #234, opened 2022-05-24
Ensure custom op domain is released

hombreola:


shared/ortcustomops.cc
```diff
@@ -1,11 +1,32 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+#include <mutex>
 #include <set>
 
 #include "onnxruntime_extensions.h"
 #include "ocos.h"
 
+struct OrtCustomOpDomainDeleter {
```
hombreola, line 10: This code snippet was taken from https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/test/testdata/custom_op_library/custom_op_library.cc

wenbingl:
thanks for the fixing.

shared/ortcustomops.cc
```diff
@@ -1,11 +1,32 @@
...
+    ort_api_->ReleaseCustomOpDomain(domain);
+  }
+
+  const OrtApi* ort_api_;
+};
+
+using OrtCustomOpDomainUniquePtr = std::unique_ptr<OrtCustomOpDomain, OrtCustomOpDomainDeleter>;
+static std::vector<OrtCustomOpDomainUniquePtr> ort_custom_op_domain_container;
+static std::mutex ort_custom_op_domain_mutex;
+
```
snnn, line 24: Please think it again, why do you need this and when the destructors of the variables will be called?

shared/ortcustomops.cc
```diff
@@ -1,11 +1,32 @@
...
+static void AddOrtCustomOpDomainToContainer(OrtCustomOpDomain* domain, const OrtApi* ort_api) {
+  std::lock_guard<std::mutex> lock(ort_custom_op_domain_mutex);
```
snnn, line 26: Why does it need be thread-safe?

shared/ortcustomops.cc
hombreola: Indeed it's testing code, but it shows how you have to take care of registering and  releasing the custom domains

shared/ortcustomops.cc
hombreola: ort_custom_op_domain_container is defined as static, so the destructor will be called when the process exits

wenbingl:
@hombreola , will this PR fix this issue #227 ?

shared/ortcustomops.cc
snnn: In which case this function might be invoked concurrently from multiple threads?

hombreola: For example, a process with multiple onnxruntime sessions running in different threads. Each session might create its own OrtSessionOptions and register the onnxruntime-extensions custom ops

shared/ortcustomops.cc
hombreola: Well if you have crash, I guess you don't care about memory leaks ;-) The OS will take care of freeing all the memory allocated by the process.
Regarding the order of destruction of global static variables, the standard says (From § 3.6.3 of the C++03 standard):

>     Destructors (12.4) for initialized objects of static storage duration (declared at block scope or at namespace scope) are called as a result of returning from main and as a result of calling exit (18.3). These objects are destroyed in the reverse order of the completion of their constructor or of the completion of their dynamic initialization. If an object is initialized statically, the object is destroyed in the same order as if the object was dynamically initialized. For an object of array or class type, all subobjects of that object are destroyed before any local object with static storage duration initialized during the construction of the sub- objects is destroyed.

shared/ortcustomops.cc
snnn: Then how many times this function will be called?

hombreola:
Not sure, I found this leak by running valgrind. I'll see if I can reproduce the errors reported in https://github.com/microsoft/onnxruntime-extensions/issues/227

shared/ortcustomops.cc
hombreola: It will be called each time the user calls RegisterCustomOpsLibrary() with "ortcustomops.dll" as argument. And this can happen concurrently from different threads in a process (although I admit it's not a common scenario)

shared/ortcustomops.cc
wenbingl: Agree, they are static so it shouldn't be accessible to any variables not in this file.

wenbingl (approved):

shared/ortcustomops.cc
snnn: @wenbingl, wait for crashes.

snnn: In my opinion, This PR may create more problems than what is solving.

hombreola: What's the risk in your opinion?
I can't see how this changes can create a problem: it's just a static vector where we keep track of all the created domains. The mutex is protecting a simple push operation that won't create any deadlock or similar. And the destructors will only be called at process exit.

snnn: OrtCustomOpDomainDeleter takes an OrtApi pointer and hold it here, which means this custom op DLL needs to take dependency on the core onnxruntime.dll. And the core onnxruntime.dll must be alive before the custom op dll is unloaded. Before your change, there wasn't such a requirement. Your change added this. And operating systems do not guarantee such an order.

snnn: So it may crash at process exit.

snnn: See my comments at https://github.com/microsoft/onnxruntime/pull/11566#issuecomment-1130767041

wenbingl: @snnn , your concern is valid. But for the custom op DLL, it should be unloaded before onnxruntime shutdown if the user wants a graceful shutdown. Not sure the onnxruntime environment will unload the DLLs on its destructor, if not, it should be added as well.

hombreola: @snnn in my opinion this specific case is safe, let me explain:
OrtApi is defined as a pointer to a "static constexpr" variable, see https://github.com/microsoft/onnxruntime/blob/006597b9b8d581c4db5e4294863a33c65cb49fb5/onnxruntime/core/session/onnxruntime_c_api.cc#L2553 and https://github.com/microsoft/onnxruntime/blob/006597b9b8d581c4db5e4294863a33c65cb49fb5/onnxruntime/core/session/onnxruntime_c_api.cc#L2274

According to the standard, static initialization is always performed before any dynamic initialization takes place (see https://stackoverflow.com/questions/24531186/is-order-of-static-initialization-implementation-defined). This implies that OrtApi will always be destroyed after all the dynamically initiazed variables are destroyed and therefore its use is perfectly safe.

wenbingl: discussed with @snnn offline, we will merge PR and do modifications if there is any concrete example causing the crash.

Issue #230, opened 2022-05-12
Add HuggingFace vocab format to BERT tokenizer


natke:

shaahji:
Fixed by #235

PR #235, opened 2022-05-25
Issue #230: Add HuggingFace vocab format to Bert tokenizer

shaahji:
Issue #230: Add HuggingFace vocab format to Bert tokenizer

HuggingFace vocab format is newline separated (unlike GPT which is
json). Newline separated is likely to be faster and doesn't require
an external library to parse it. Instead of introducing a json based
format, added support for native HuggingFace newline separated token
format.


wenbingl (approved):

natke (commented):
Looks good @shaahji. Can you please add a test for the new format?

shaahji:
> Looks good @shaahji. Can you please add a test for the new format?

On it. Also, I have requested a more elaborate test from @naraymad.

PR #240, opened 2022-06-03
Create onebranch pipeline for win32-wheel

snnn:
For compliance work. This is just the start.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -133,6 +133,9 @@ def read_version():
     return version_str
 
 
+if sys.platform == "win32":
+    load_msvcvar()
+
 ext_modules = [
     setuptools.extension.Extension(
         name=str('onnxruntime_extensions._ortcustomops'),
--- a/setup.py
+++ b/setup.py
@@ -133,9 +133,6 @@ def read_version():
     return version_str
 
 
-if sys.platform == "win32":
-    load_msvcvar()
-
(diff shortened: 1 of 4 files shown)


natke (approved):

setup.py
```diff
@@ -133,9 +133,6 @@ def read_version():
     return version_str
 
 
-if sys.platform == "win32":
-    load_msvcvar()
```
wenbingl, line 137: Why do you need to remove this line?

snnn: This project is cmake-based. You don't need to run cmake and the build commands in VC++'s  environment. You can directly run cmake in a normal cmd.exe Windows. On the other side, when you run this bat, it adds a lot of things to the PATH, not just the compilers. For example, it brings in cmake/python/...

wenbingl: In some environments like Powershell, VC batch file cannot be directly used, that's why this line was added.

snnn: Ok. let me revert it. 
To clarify, you don't need to use the VC batch file. CMake can auto find VC++ paths without the help of the VC batch file

wenbingl: Yes, the cmake as a pip package can find VC++. but there is a bug in pip version cmake sometimes prevent it working with setup.py.

wenbingl (approved):

PR #249, opened 2022-06-07
Add two opencv operators as ONNX custom ops.

wenbingl:


natke:
So the two operators are GaussianBlur and ImageReader? 

Can we add these two to a list of supported ops in the library? Or create the list if we don't already have it.

Why these two?

Do we have any usage examples?

cmake/externals/opencv.cmake
```diff
@@ -93,12 +93,15 @@ FetchContent_Declare(
         -DBUILD_SHARED_LIBS:BOOL=FALSE
         -DCMAKE_INSTALL_PREFIX:PATH=${CMAKE_CURRENT_BINARY_DIR}/opencv
         -DCV_TRACE:BOOL=FALSE
-        PATCH_COMMAND  git apply --whitespace=fix ${CMAKE_CURRENT_SOURCE_DIR}/cmake/externals/opencv-no-rtti.patch
+        PATCH_COMMAND git checkout . && git apply --whitespace=fix --reject --allow-empty ${CMAKE_CURRENT_SOURCE_DIR}/cmake/externals/opencv-no-rtti.patch
```
snnn, line 96: Try this: https://github.com/microsoft/onnxruntime/blob/master/cmake/external/abseil-cpp.cmake

I suggest avoid using git apply, use patch directly.

wenbingl:
@snnn , it looks like the BinSkim rule is too strict, which complains about the C++ warning disable in some other library. Can we disable this rule in some config.

  Tool: BinSkim: Rule: BA2007 (EnableCriticalCompilerWarnings). https://github.com/microsoft/binskim/blob/main/docs/BinSkimRules.md#rule-BA2007EnableCriticalCompilerWarnings
  'ortcustomops.dll' disables compiler warning(s) which are required by policy.

snnn:
No. It is not negotiable.

snnn:
If the warnings are from 3rd-party code that you don't have any control, you can ignore these findings. You will still need to enable the critical compiler warnings for your code.

cmake/externals/opencv.cmake
wenbingl: patch command line is only available on Unix-like systems, and git apply command is same among all platforms. so it looks like it is better.

shaahji (approved):

skottmckay:
> Why these two?

I'd like to know as well. Naively I would have expected things like gaussian blur and inverse to happen when training (creates more variations of input from a set of images) rather than when inferencing.

wenbingl:
> > Why these two?
> 
> I'd like to know as well. Naively I would have expected things like gaussian blur and inverse to happen when training (creates more variations of input from a set of images) rather than when inferencing.

@skottmckay  @natke 

There is no priority list of OpenCV operation for ONNX now, so these two operators were picked because they are pretty typical to give some concrete examples on opencv modules imgcodecs and imgproc and verify the opencv integration.

Resize could be a candidate but I verified the output from openCV are quit different from Pillow, so I don't choose it to avoid more confusion.

There will be a tutorial on how to use openCV in onnxruntime-extension in the next PR.

PR #252, opened 2022-06-16
Example to demonstrate BertTokenizer as a customop

shaahji:
A simple example to demonstrate loading of BertTokenizer as a customop
in a onnx model


tutorials/bert_customop.py
```diff
@@ -0,0 +1,69 @@
...
+# Load the onnx model as encoded string
+content = open(str(model_path), "rb").read()
+onnx_model_b64 = base64.b64encode(base64.decodebytes(content))
+
+# Load the corresponding vocabulary
+vocab_filepath = "data/bert-base-cased.txt"
+vocabulary = open(vocab_filepath, 'r', encoding='utf-8').read()
+
+# Create a new node for the custom op
+customop_node = onnx.helper.make_node(
```
wenbingl, line 34: Why do we use ONNX API, instead of pnp.export here?

wenbingl:
What's the function difference from bert_e2e.py in the same folder?

shaahji:
> What's the function difference from bert_e2e.py in the same folder?

bert_e2e is an example demonstrating use of bert with pre & post processing steps.
This example is demonstrating use of bert tokenizer as a custom op.

This example was requested by the AutoML team.

tutorials/bert_customop.py
shaahji: We are loading the model only as a base64 encoded data. We aren't really exporting the model here to disk. We are merely building it in memory.

shaahji:
Discard!

Issue #244, opened 2022-06-06
[COMPLIANCE] Create cgmanifest.json


natke:
Add all C++ dependencies

natke:
@shaahji follow up with Changming

natke:
PR here https://github.com/microsoft/onnxruntime-extensions/pull/254

natke:
Get Office team cgmanifest

shaahji:
Fixed by #254

PR #254, opened 2022-06-22
Issue #244: Create cgmanifest.json for the project

shaahji:
Issue #244: Create cgmanifest.json for the project


cgmanifest.json
```diff
@@ -0,0 +1,167 @@
...
+          "Name": "torch",
+          "Version": "1.11.0"
+        }
+      }
+    },
+    {
+      "component": {
+        "type": "git",
+        "git": {
+          "commitHash": "aaa5eab410c2e3fe41ebf7a2316be9c51572dbd2",
```
wenbingl, line 151: comes from any dependency?

wenbingl:
How do you generate this file? can we have some script to do it?

cgmanifest.json
```diff
@@ -0,0 +1,167 @@
...
+          "Name": "onnxruntime",
+          "Version": "1.11.1"
+        }
+      }
+    },
+    {
+      "component": {
+        "type": "pip",
+        "pip": {
+          "Name": "onnxconverter",
```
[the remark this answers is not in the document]
shaahji: I have had to install this to get build working.

cgmanifest.json
```diff
@@ -0,0 +1,167 @@
...
+          "repositoryUrl": "https://gitlab.kitware.com/cmake/cmake"
+        }
+      }
+    },
+    {
+      "component": {
+        "type": "pip",
+        "pip": {
+          "Name": "cmake",
+          "Version": "3.22.5"
```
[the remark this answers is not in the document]
shaahji: ONNXRuntime lists it in their version so I presume it's required.

cgmanifest.json
shaahji: It's version 3.18.2.

cgmanifest.json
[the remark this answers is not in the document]
wenbingl: I search pypi.org, and there is no onnxconverter, where do you install it from?

shaahji:
> It is not about what are necessary for building this repo from source locally, it is about what are used in ort-extension packages. Especially, what are used for building the packages that your team publishes.

Wouldn't those two be the same, at least for the most part. Everything in that file is needed to successfully build locally and should be needed in CI as well. There might be some version differences depending on which version of compiler you use or what version of python are you running. This list might still be _incomplete_ since there might be more dependencies needed for steps that we don't necessarily run locally. Call it minimal list of dependencies.

cgmanifest.json
```diff
@@ -0,0 +1,167 @@
...
+          "commitHash": "v3.7.3",
+          "repositoryUrl": "https://github.com/nlohmann/json.git"
+        }
+      }
+    },
+    {
+      "component": {
+        "type": "git",
+        "git": {
+          "commitHash": "4.5.4",
```
snnn, line 52: opencv has a lot of dependencies. For example, libpng. If it was used when you build your package, then you need to list it here as well.  Transitively. 
Without the dependencies, opencv itself doesn't do any useful thing.

shaahji: Okay! I will cover the transitive dependencies as well. How accurate should be versions be? When the requirement is a range which should be included, the lower or upper?

snnn: Should be exactly the version that you use for building your packages. Exactly. Not a range.

snnn:
For example, if your package has OCOS_ENABLE_OPENCV_CODECS=ON, which means your binary contains jpeg code. But jpeg has patents. You will need to pay someone to get the authorization first. If you don't put jpeg in the cgmanifest.json, I don't know you are using jpeg. Then I wouldn't let know you how to start the legal process. Then it's pirating. We should carefully avoid it happening. It's like, when you write a paper, you must be careful on the references. If you didn't do it right, it may ruin your academic life.

wenbingl:
> For example, if your package has OCOS_ENABLE_OPENCV_CODECS=ON, which means your binary contains jpeg code. But jpeg has patents. You will need to pay someone to get the authorization first. If you didn't put jpeg in the cgmanifest.json, then I wouldn't know you are using jpeg, then I wouldn't ask to you to start the legal process. Then it may become pirating. Therefore we should be careful on such things. It's like, when you write a paper, you must carefully list all the references. If you didn't do it right, it may ruin your academic life. So, a "minimal list“ is not enough. The list should be full and complete and accurate.

jpeg patent expires, so the CODEC should be same as another library as well.

snnn:
> jpeg patent expires, so the CODEC should be same as another library as well.

There are two things: patents, and license agreement. Even patents are expired, we still need to register all such things in cgmanifest.json then put them in ThirdPartyNotice.txt file.

snnn (approved):

snnn:
@wenbingl , please take a look.

wenbingl (approved):

PR #255, opened 2022-06-23
Issue #230: Add test case for BertTokenizer operator

shaahji:
Issue #230: Add test case for BertTokenizer operator

Test both options - using the vocab file and using pre-created tokenizer

Fixed argument handling in pnp where the arguments weren't being passed
down to the tokenizer as expected.


shaahji:
Will reintroduce the change once #268 gets merged.

Issue #243, opened 2022-06-06
Cannot add input and output names to model with BertTokenizer


natke:
When specifying input_names and output_names to pnp.export, they are not added to the graph.

Augment model code is:

```
augmented_model = pnp.export(augmented_model,
                             test_question,
                             opset_version=12,
                             input_names=['question'],
                             output_names=['start', 'end'],
                             output_path=model_name + '-aug.onnx',
                             dynamic_axes={'question': [0]})
```

Source here: https://github.com/natke/onnxruntime-inference-examples/blob/dataproc-nlp-mobile/mobile/examples/question_answering/model/augment_model.py

Graph inputs and outputs:

![MicrosoftTeams-image (1)](https://user-images.githubusercontent.com/3302433/172221809-dfdddc12-b090-4967-b4bd-b249c67191b6.png)
![MicrosoftTeams-image (2)](https://user-images.githubusercontent.com/3302433/172221833-9647913c-f1ba-4810-a3db-d86b3db88e3d.png)

shaahji:
Fixed by #263

PR #263, opened 2022-07-18
Issue#243: Cannot rename input and output names of generated model

shaahji:
Issue#243: Cannot rename input and output names of generated model

When the input is a string, the logic takes a different route where the
input model is split into two and joined again. The user provided
input/output names were not respected on this code path. Fixed the
issue by renaming the input/output post join operation.


natke (dismissed):

natke (dismissed):

natke:
Fixes #243

onnxruntime_extensions/pnp/_torchext.py
```diff
@@ -268,4 +268,35 @@ def export(self, *args, **kwargs):
...
+                    assert len(model.graph.output) == len(kwargs['output_names']), \
+                      "Expecting {} output names but got {}".format(
+                        len(model.graph.output), len(kwargs['output_names']))
+                    for n, new_name in zip(model.graph.output, kwargs['output_names']):
+                        swaps[n.name] = new_name
+                        n.name = new_name
+
+                if swaps:
+                    for n in model.graph.node:
+                        for j in range(len(n.input)):
```
wenbingl, line 296: also need to update the model.graph.initialize, which also contains the name.

onnxruntime_extensions/pnp/_torchext.py
```diff
@@ -268,4 +268,35 @@ def export(self, *args, **kwargs):
             return super().export(*args, **kwargs)
         else:
             oxml = core_m.export(*args, **kwargs)
-            return ONNXModelUtils.join_models(prefix_m, oxml)
+            model = ONNXModelUtils.join_models(prefix_m, oxml)
+
+            # Rename the input/output node names if the user has provided any substitutions!
+            # Ref: https://github.com/onnx/onnx/issues/2052
+            if (('input_names' in kwargs) or ('output_names' in kwargs)) and \
```
wenbingl, line 275: One more notes here is that the current code doesn't support updating the node with a sub-graph.

onnxruntime_extensions/pnp/_torchext.py
```diff
@@ -268,4 +268,35 @@ def export(self, *args, **kwargs):
             return super().export(*args, **kwargs)
         else:
             oxml = core_m.export(*args, **kwargs)
-            return ONNXModelUtils.join_models(prefix_m, oxml)
+            model = ONNXModelUtils.join_models(prefix_m, oxml)
```
wenbingl, line 271: ideally, the input_names should be in prefix_m only, and output_names is in oxml only. but the implementation is also since it is more straightforward.

onnxruntime_extensions/pnp/_torchext.py
shaahji: Done!

onnxruntime_extensions/pnp/_torchext.py
shaahji: Apparently, the  join operation can change the node names to prevent name collisions. It's better to do this rename after the join operation.

onnxruntime_extensions/pnp/_torchext.py
shaahji: Made a note in the comments.

wenbingl (approved):

PR #264, opened 2022-07-18
Issue #226: Prepare BertTokenizer implementation for versioning

shaahji:
Issue #226: Prepare BertTokenizer implementation for versioning

* Moved a few variables from Kernel implementation to BertTokenizer so
  each version of the Kernel doesn't have to deal with them.
* Other decorative and code standardization changes.

### DIFF
--- a/operators/tokenizer/bert_tokenizer.cc
+++ b/operators/tokenizer/bert_tokenizer.cc
@@ -2,8 +2,8 @@
 
 #include <utility>
 
-BertTokenizerVocab::BertTokenizerVocab(std::string vocab) : raw_vocab_(vocab) {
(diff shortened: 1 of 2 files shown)


wenbingl:
@shaahji  what's the feature change on this PR, or is it just some code refactoring?

shaahji:
There were members in the Kernel which should have been in the core BertTokenizer class. A follow up PR https://github.com/microsoft/onnxruntime-extensions/pull/265 adds a new Kernel & CustomOp. Implementing the same logic again in this CustomOp would just be duplication.

shaahji:
The change is merged as part of #265

PR #265, opened 2022-07-20
Issue #226: Functional e2e NLP example

shaahji:
Issue #226: Functional e2e NLP example

* Implemented a new version of Kernel and the CustomOp to support
  output that matches the HuggingFace model's input without the need
  for intermediate python logic.
* Implemented a e2e tutorial for exporting and inferencing using the
  HuggingFace's QuestionAnswering model.

Known Issue: Python side doesn't have an implementation of Bert Decoder
and so the augmented model is only half-complete. At the time of
inferencing the HuggingFace tokenizer is used to decode the result back
to string.

### DIFF
--- a/onnxruntime_extensions/pnp/_nlp.py
+++ b/onnxruntime_extensions/pnp/_nlp.py
@@ -13,9 +13,7 @@ def make_custom_op(ctx, op_type, input_names, output_names, container, operator_
                        op_version=1, name=op_name, op_domain=default_opset_domain(), **kwargs)
 
 
-@schema(inputs=((_dt.STRING, []),),
-        outputs=((_dt.INT64, []), (_dt.INT64, []), (_dt.INT64, [])))
-def bert_tokenize(ctx, input_names, output_names, container, operator_name=None, **kwargs):
+def create_bert_tokenizer(ctx, name, input_names, output_names, container, operator_name=None, **kwargs):
     if 'hf_tok' in kwargs:
         hf_bert_tokenizer = kwargs['hf_tok']
         ordered_vocab = OrderedDict(sorted(hf_bert_tokenizer.vocab.items(), key=lambda item: int(item[1])))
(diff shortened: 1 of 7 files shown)


onnxruntime_extensions/pnp/_nlp.py
```diff
@@ -35,10 +33,25 @@ def bert_tokenize(ctx, input_names, output_names, container, operator_name=None,
...
+        outputs=((_dt.INT64, []), (_dt.INT64, []), (_dt.INT64, [])))
+def bert_tokenizer(ctx, input_names, output_names, container, operator_name=None, **kwargs):
+    return create_bert_tokenizer(ctx, 'BertTokenizer', input_names, output_names,
+                                 container, operator_name=operator_name, **kwargs)
+
+
+@schema(inputs=((_dt.STRING, []),),
+        outputs=((_dt.INT64, []), (_dt.INT64, []), (_dt.INT64, [])))
+def bert_tokenizer_v2(ctx, input_names, output_names, container, operator_name=None, **kwargs):
+    return create_bert_tokenizer(ctx, 'HfBertTokenizer',
```
wenbingl, line 50: Can we unify the name, either it starts with hf, or ends with v2?

wenbingl (commented):
Besides the e2e script, can you also add a unit test for HfBertTokenizer?

operators/tokenizer/bert_tokenizer.cc
```diff
@@ -2,8 +2,8 @@
 
 #include <utility>
 
-BertTokenizerVocab::BertTokenizerVocab(std::string vocab) : raw_vocab_(vocab) {
-  auto tokens = SplitString(raw_vocab_, "\n", true);
+BertTokenizerVocab::BertTokenizerVocab(std::string_view vocab) : raw_vocab_(vocab) {
```
wenbingl, line 5: is string_view introduced in C++ 17? IIRC, the largest C++ standard used in onnxruntime is c++14 for some reasons. And the onnxruntime-extensions may need to be consistent with it.

onnxruntime_extensions/pnp/_nlp.py
shaahji: Done!

natke (approved):
Approving the e2e example :shipit: 

Someone else should review the other changes.

shaahji:
> Besides the e2e script, can you also add a unit test for HfBertTokenizer?

Yes, I have the test locally. Trying to get the CI build issues resolved before I introduce the tests. PR coming soon to update the workflow, followed with tests.

shaahji:
> Besides the e2e script, can you also add a unit test for HfBertTokenizer?

#269

wenbingl (approved):

operators/tokenizer/bert_tokenizer.cc
wenbingl: @snnn , is C++ 17 already supported in onnxruntime now?

PR #267, opened 2022-07-21
Leqiao/code sign

leqiao-1:
add code sign task for libraries.

### DIFF
--- a/.pipelines/OneBranch.Official.yml
+++ b/.pipelines/OneBranch.Official.yml
@@ -86,4 +86,21 @@ extends:
             python -m pip install cibuildwheel
             python -m cibuildwheel --platform windows --output-dir $(REPOROOT)\out
           displayName: Build wheels
+        - script: |
+            7z x $(REPOROOT)\out\*.whl -o$(REPOROOT)\out\*
+          displayName: unzip the package
+        - task: onebranch.pipeline.signing@1
+          inputs:
+            command: 'sign'
+            signing_profile: 'external_distribution'
+            files_to_sign: '**/**/*.pyd;**/**/*.dll;**/**/*.so'
+            search_root: '$(REPOROOT)\out'
+          displayName: 'Sign runtime DLLs'
+        - script: |
+            for /D %%i in ("out\onnxruntime_extensions*-win32") do (
+              7z u %%i.whl .\%%i\*
+              rmdir /s /q %%i
+              )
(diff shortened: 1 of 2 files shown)


.pipelines/OneBranch.PullRequest.yml
```diff
@@ -86,4 +86,20 @@ extends:
...
-        
\ No newline at end of file
+        - script: |
+            7z x $(REPOROOT)\out\*.whl -o$(REPOROOT)\out\*
+          displayName: unzip the package
+        - task: onebranch.pipeline.signing@1
+          inputs:
+            command: 'sign'
+            signing_profile: 'external_distribution'
+            files_to_sign: '**/**/*.pyd;**/**/*.dll;**/**/*.so'
```
snnn, line 96: *.so is not signable. Please do not include *.so files.

leqiao-1: *.so files removed

snnn (approved):

wenbingl:
@snnn , @leqiao-1 , why do we need sign both official and pull-request?

.pipelines/OneBranch.Official.yml
```diff
@@ -86,4 +86,21 @@ extends:
...
           displayName: Build wheels
+        - script: |
+            7z x $(REPOROOT)\out\*.whl -o$(REPOROOT)\out\*
+          displayName: unzip the package
+        - task: onebranch.pipeline.signing@1
+          inputs:
+            command: 'sign'
+            signing_profile: 'external_distribution'
+            files_to_sign: '**/**/*.pyd;**/**/*.dll'
+            search_root: '$(REPOROOT)\out'
```
wenbingl, line 97: 'out' happens to be the default native build target folder, can we pick another name to avoid the potential conflict in the future?

snnn:
I would suggest keeping the two pipeline as similar as possible.

.pipelines/OneBranch.Official.yml
snnn: You can change the line 62:
```
          ob_outputDirectory: '$(REPOROOT)\out'
```

leqiao-1: I did not quite unterstand. Should I modify the output folder of wheel build task ? ```python -m cibuildwheel --platform windows --output-dir $(REPOROOT)\out```

wenbingl (approved):

PR #268, opened 2022-07-21
Update workflow matrix

shaahji:
Update workflow matrix

Update workflow matrix so it's consistent across the platforms and using
newer versions of the dependencies. Current supported matrix -

Python | 3.7 | 3.8 | 3.9 | 3.10
---|---|---|---|---
Onnxruntime | 1.9.0 (Sept 22, 2021) | 1.10.0 (Dec 7, 2021) | 1.11.0 (Mar 26, 2022) | 1.12.1 (Aug 4, 2022)
Torch | 1.9.1 (Sept 22, 2021) | 1.10.0 (Oct 21, 2021) | 1.11.0 (Mar 10, 2022) | 1.12.1 (Aug 5, 2022)
TorchVision | 0.10.1 (Jun 15, 2021) | 0.11.1 (Oct 21, 2021) | 0.12.0 (Mar 10, 2022) | 0.13.1 (Aug 5, 2022)
TorchAudio | 0.9.0 (Jun 15, 2021) | 0.10.0 (Oct 21, 2021) | 0.11.0 (Mar 10, 2022) | 0.12.1 (Aug 5, 2022)

Release versions strictly follow the convention of onnxruntime being one
release ahead of all its dependencies.

### DIFF
--- a/.az/mshost.yaml
+++ b/.az/mshost.yaml
@@ -1,3 +1,15 @@
+# IMPORTANT: READ ME BEFORE MAKING ANY CHANGES
+#
+# Supported matrix should follow these conventions strictly and
+# for any exceptions to the rules, please document them in comments.
+#
+#  * When adding new configurations to the matrix, update all platforms simultaneously.
+#  * Onnxruntime version should be one release ahead of torch, torchvision, and torchaudio.
+#
+# The rationale is that the onnxruntime release process would have verified the
+# last available version of its dependencies (including the three listed above)
+# and so it's'likely to be be _better_ supported relative to any older versions.
+
 jobs:
 
   #######
@@ -10,25 +22,30 @@ jobs:
 
     strategy:
       matrix:
-        py39-170:
+        py310-1121:
+          python.version: '3.10'
+          torch.version: 'torch==1.12.1+cpu torchvision==0.13.1+cpu torchaudio==0.12.1+cpu -f https://download.pytorch.org/whl/cpu/torch_stable.html'
+          ort.version: '1.12.0'
+        py39-1110:
           python.version: '3.9'
+          torch.version: 'torch==1.11.0+cpu torchvision==0.12.0+cpu torchaudio==0.11.0+cpu -f https://download.pytorch.org/whl/cpu/torch_stable.html'
+          ort.version: '1.11.0'
+        py38-1100:
+          python.version: '3.8'
           torch.version: 'torch==1.10.0+cpu torchvision==0.11.1+cpu torchaudio==0.10.0+cpu -f https://download.pytorch.org/whl/cpu/torch_stable.html'
-          ort.version: '1.7.0'
-          ortlib.version: '38443267'
-        py37-160:
+          ort.version: '1.10.0'
+        py37-190:
           python.version: '3.7'
-          torch.version: 'torch==1.7.1+cpu torchvision==0.8.2+cpu torchaudio==0.7.2 -f https://download.pytorch.org/whl/torch_stable.html'
-          ort.version: '1.6.0'
-          ortlib.version: '34858191'
-      maxParallel: 2
+          torch.version: 'torch==1.9.0+cpu torchvision==0.10.0+cpu torchaudio==0.9.0 -f https://download.pytorch.org/whl/cpu/torch_stable.html'
+          ort.version: '1.9.0'
 
     steps:
       - task: DownloadGitHubRelease@0
         inputs:
           connection: 'microsoft'
           userRepository: 'microsoft/onnxruntime'
-          defaultVersionType: 'specificVersion'
-          version: '$(ortlib.version)'
+          defaultVersionType: 'specificTag'
+          version: 'v$(ort.version)'
           itemPattern: '*-linux-x64-$(ort.version)*'
           downloadPath: '$(Build.sourcesdirectory)'
         displayName: Download the ONNXRuntime prebuilt package.
@@ -79,19 +96,24 @@ jobs:
   # macOS C++
   ###########
 
-  - job: MacOSC
+  - job: MacOSX
     pool:
       vmImage: 'macOS-latest'
 
     strategy:
       matrix:
-        ort-170:
-          ort.version: '1.7.0'
-          ortlib.version: '38443267'
-        ort-160:
-          ort.version: '1.6.0'
-          ortlib.version: '34858191'
-      maxParallel: 2
+        ort-1121:
+          ort.version: '1.12.1'
+          ort.dirname: 'onnxruntime-osx-x86_64-$(ort.version)'
+        ort-1110:
+          ort.version: '1.11.0'
+          ort.dirname: 'onnxruntime-osx-x86_64-$(ort.version)'
+        ort-1100:
+          ort.version: '1.10.0'
+          ort.dirname: 'onnxruntime-osx-x86_64-$(ort.version)'
+        ort-190:
+          ort.version: '1.9.0'
+          ort.dirname: 'onnxruntime-osx-x64-$(ort.version)'
 
     steps:
       # needed for onnxruntime
@@ -102,9 +124,9 @@ jobs:
         inputs:
           connection: 'microsoft'
           userRepository: 'microsoft/onnxruntime'
-          defaultVersionType: 'specificVersion'
-          version: '$(ortlib.version)'
-          itemPattern: '*-osx-x64-$(ort.version)*'
+          defaultVersionType: 'specificTag'
+          version: 'v$(ort.version)'
+          itemPattern: '$(ort.dirname)*'
           downloadPath: '$(Build.sourcesdirectory)'
         displayName: Download the ONNXRuntime prebuilt package.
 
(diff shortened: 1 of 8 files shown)


wenbingl (commented):
The header files are copied from onnxruntime of 1.6 or 1.7, it would be better to upgrade the files as well?

I didn't see the benefits of this kind of upgrading,  unless onnxruntime-extensions needs to use the feature from the newer version of onnxruntime.

shaahji:
> The header files are copied from onnxruntime of 1.6 or 1.7, it would be better to upgrade the files as well?

Do you mean that we have copies of onnxruntime headers in this repository? Any reason why not just pull in a release (just like all other dependencies) instead of copying the headers?

> I didn't see the benefits of this kind of upgrading, unless onnxruntime-extensions needs to use the feature from the newer version of onnxruntime.

Consistency across platforms is one obvious benefit. Plus, having a document that explicitly lists the dependencies and respective versions making it easier for clients.

natke (dismissed):

natke:
Can we add this information to the README please?

wenbingl:
> Can we add this information to the README please?



> Can we add this information to the README please?



> > The header files are copied from onnxruntime of 1.6 or 1.7, it would be better to upgrade the files as well?
> 
> Do you mean that we have copies of onnxruntime headers in this repository? Any reason why not just pull in a release (just like all other dependencies) instead of copying the headers?
> 
> > I didn't see the benefits of this kind of upgrading, unless onnxruntime-extensions needs to use the feature from the newer version of onnxruntime.
> 
> Consistency across platforms is one obvious benefit. Plus, having a document that explicitly lists the dependencies and respective versions making it easier for clients.

the current pipeline ensure the backward compatible with ORT versions. please don't remove the test on the old version so rush.

shaahji:
> the current pipeline ensure the backward compatible with ORT versions. please don't remove the test on the old version so rush.

What versions are you trying to be compatible with? We should extend the pipeline to include exactly what we are supporting and testing.

This is a new library that we plan to release. Working with a very wide matrix makes things harder to develop and test. There is always room to widen the matrix once a smaller subset is stabilized. If absolutely necessary, a client's requirements can always be accommodated.

wenbingl:
> Can we add this information to the README please?

in requirements.txt,
onnx>=1.9.0
onnxruntime>=1.6.0

which means it should be able to work with any versions of onnxruntime above 1.6.0. If we want to reduce the version range, I need to understand the reason before we make the change.

wenbingl:
> 

@shaahji , check this reply. I am ok with torch upgrading since it will be removed soon. But we need a reason to drop the supports on ort 1.6 - 1.9. At least, ort 1.9 and the lower version was still widely used in internal partner teams.

shaahji:
> > 
> 
> @shaahji , check this reply. I am ok with torch upgrading since it will be removed soon. But we need a reason to drop the supports on ort 1.6 - 1.9. At least, ort 1.9 and the lower version was still widely used in internal partner teams.

We are not dropping support for any version. The CI builds dictate what we are developing against and actively testing against. If we include 1.6 - 1.9 then we should include all those configurations in our matrix as well and verify every change against all those configurations.

wenbingl:
> > > 
> > 
> > 
> > @shaahji , check this reply. I am ok with torch upgrading since it will be removed soon. But we need a reason to drop the supports on ort 1.6 - 1.9. At least, ort 1.9 and the lower version was still widely used in internal partner teams.
> 
> We are not dropping support for any version. The CI builds dictate what we are developing against and actively testing against. If we include 1.6 - 1.9 then we should include all those configurations in our matrix as well and verify every change against all those configurations.

Yes, all configurations in extreme case.  But in reality, ORT ABI is backward compatible, so we may skip some configurations but the upbound and low bound configurations needs to be there always.
Without this CI pipeline checking, the supports on the older version can be easily broken.

shaahji:
> Yes, all configurations in extreme case. But in reality, ORT ABI is backward compatible, so we may skip some configurations but the upbound and low bound configurations needs to be there always. Without this CI pipeline checking, the supports on the older version can be easily broken.

As it stands today, even a difference in opset_version of 11 vs. 12 causes half of the CI builds to fail. So, is this the right time to be spending cycles on trying to drill down and find the problem? May be or may be not. That's not my call to make. However, having a client with a justified need to support a specific version raises the stakes. I can follow up with another PR to find the lower bound for all the dependencies and make that part of the process. That's an iterative process and likely going to take some time.

wenbingl:
> > Yes, all configurations in extreme case. But in reality, ORT ABI is backward compatible, so we may skip some configurations but the upbound and low bound configurations needs to be there always. Without this CI pipeline checking, the supports on the older version can be easily broken.
> 
> As it stands today, even a difference in opset_version of 11 vs. 12 causes half of the CI builds to fail. So, is this the right time to be spending cycles on trying to drill down and find the problem? May be or may be not. That's not my call to make. However, having a client with a justified need to support a specific version raises the stakes. I can follow up with another PR to find the lower bound for all the dependencies and make that part of the process. That's an iterative process and likely going to take some time.

I am not sure why you end-to-end qna example has to be linked to this test configuration. Can you be more specific what the issue is if we don't upgrade the pipeline.

Normally, if your example needs to work with, let's say, ort 1.11 and the above only. you can add the configuration if it doesn't exist, and the unit test skips the old version configuration. This is a pretty general in Python development practice.

In the future, ort-extension may be 1:1 mapping with ort version, which may help reduce some cost, but that's another story.

shaahji:
> I am not sure why you end-to-end qna example has to be linked to this test configuration. 

Created a new draft PR #270 which will fail on current CI. Feel free to dig in. 

> Can you be more specific what the issue is if we don't upgrade the pipeline.

This has nothing to do with qna example. The failures are to do with using a standalone BertTokenizer as a customop. Plus, with AutoML asking what version of dependencies are supported, we didn't have a good answer to give them. This change is to counter that.

> Normally, if your example needs to work with, let's say, ort 1.11 and the above only. you can add the configuration if it doesn't exist, and the unit test skips the old version configuration. This is a pretty general in Python development practice.

Ideally this example should work with any version of ORT but some combination of ORT, torch, and the exporter fails on specific platforms.

shaahji:
@wenbingl Here's the link to the issue - https://aiinfra.visualstudio.com/Lotus/_build/results?buildId=219940&view=logs&j=de5a03c9-78f6-5547-81a7-ff6fd764fa78&t=4251e826-658c-5967-f761-6c64b7841cbb&l=99

This problem happens only on specific configurations.

wenbingl:
> @wenbingl Here's the link to the issue - https://aiinfra.visualstudio.com/Lotus/_build/results?buildId=219940&view=logs&j=de5a03c9-78f6-5547-81a7-ff6fd764fa78&t=4251e826-658c-5967-f761-6c64b7841cbb&l=99
> 
> This problem happens only on specific configurations.

According to this line:
  if (!model_proto.has_ir_version() || model_proto.ir_version() > ONNX_NAMESPACE::Version::IR_VERSION) {
    ORT_THROW("Unknown model file format version.");
  }

The model IR_version is higher than onnxruntime 1.6 and 1.7 supports

wenbingl:
> > @wenbingl Here's the link to the issue - https://aiinfra.visualstudio.com/Lotus/_build/results?buildId=219940&view=logs&j=de5a03c9-78f6-5547-81a7-ff6fd764fa78&t=4251e826-658c-5967-f761-6c64b7841cbb&l=99
> > This problem happens only on specific configurations.
> 
> According to this line: if (!model_proto.has_ir_version() || model_proto.ir_version() > ONNX_NAMESPACE::Version::IR_VERSION) { ORT_THROW("Unknown model file format version."); }
> 
> The model IR_version is higher than onnxruntime 1.6 and 1.7 supports

I just viewed your model, it's opset 11, then ir version should be 6, which should be ok with 1.6 or 1.7.
There are some implications in ONNX api, and make it be error-prone.  How do you get this model?

https://github.com/shaahji/onnxruntime-extensions/blob/3b2409d88030d5e989668f81bf623877e81838e9/onnxruntime_extensions/pnp/_onnx_ops.py#L13

shaahji:
Moving this conversation to https://github.com/microsoft/onnxruntime-extensions/pull/270

wenbingl:
@shaahji , can you re-check the meeting note and convert it to a MD doc in this PR as well?
I remember the note should have more configurations than the current change.

shaahji:
> can you re-check the meeting note and convert it to a MD doc in this PR as well?

MD doc?

> I remember the note should have more configurations than the current change.

Yes, that's what I am working on. The current push was a rebase to bring the branch up-to-date. I will mark this PR as a draft until it's ready for review.

wenbingl:
> @shaahji , can you re-check the meeting note and convert it to a MD doc in this PR as well? I remember the note should have more configurations than the current change.



> > can you re-check the meeting note and convert it to a MD doc in this PR as well?
> 
> MD doc?
> 
> > I remember the note should have more configurations than the current change.
> 
> Yes, that's what I am working on. The current push was a rebase to bring the branch up-to-date. I will mark this PR as a draft until it's ready for review.

MD doc = markdown doc

natke (approved):

wenbingl (approved):
It looks great, Thanks!

PR #270, opened 2022-07-22
Issue #230, #226: Add tests for BertTokenizer & HfBertTokenizer customop

shaahji:
Issue #230, #226: Add tests for BertTokenizer & HfBertTokenizer customop

Includes a new onnx model for testing the HfBertTokenizer that accepts
two strings as input and returns output compatible with inputs for
pretrained HuggingFace models.


test/test_bert_tokenizer_op.py
```diff
@@ -0,0 +1,96 @@
...
+    np.testing.assert_array_equal(actual_result[0], expect_result["input_ids"])
+    np.testing.assert_array_equal(actual_result[1], expect_result["token_type_ids"])
+    np.testing.assert_array_equal(actual_result[2], expect_result["attention_mask"])
+
+
+class TestBertTokenizerOp(unittest.TestCase):
+    def test_text_to_case1_with_vocab_file(self):
+        ort_tok = pnp.PreHuggingFaceBert(vocab_file=_vocab_file, do_lower_case=0, strip_accents=1)
+        model = pnp.export(pnp.SequentialProcessingModule(ort_tok), ["whatever"], opset_version=12)
+        t2stc = OrtPyFunction.from_model(model)
```
shaahji, line 26: > How do you get this model?

The model is created right here. There is no external model loaded. The tokenizer comes from transformers (Line 10-11 above).

shaahji: @wenbingl CC

test/test_bert_tokenizer_op.py
```diff
@@ -0,0 +1,96 @@
...
+        _run_basic_case(
+            t2stc,
+            input="cat isnot playing toyssss"
+        )
+
+
+    def test_text_to_case1_with_hf_tok(self):
+        ort_tok = pnp.PreHuggingFaceBert(hf_tok=_bert_cased_tokenizer)
+        model = pnp.export(pnp.SequentialProcessingModule(ort_tok), ["whatever"], opset_version=12)
+        t2stc = OrtPyFunction.from_model(model)
```
shaahji, line 62: @wenbingl Same here. The model is created right here.

wenbingl: I think this is a bug that pnp doesn't handle opset_version very well on generating the final model. I need some time to check why it happens since there is some code logic here https://github.com/microsoft/onnxruntime-extensions/blob/476a56e6fe50647d4bbca65405ff1c2b5d083a32/onnxruntime_extensions/pnp/_onnx_ops.py#L13 to do the conversion.

that's a good example that why old version of ort CI pipeline needs to be kept.

shaahji:
Abandoning this in favor of #268

Issue #279, opened 2022-08-20
TryGetAttribute<std::string> fails for valid input


adrianlizarraga:
Hi,

The `TryGetAttribute<std::string>()` function in [BaseKernel](https://github.com/microsoft/onnxruntime-extensions/blob/main/operators/ocos.cc#L56) does not correctly handle the return value of `KernelInfoGetAttribute_string()` when querying for the attribute's size (i.e., the 3rd argument is a nullptr). As a result, this function always returns `false` for any valid input.

Refer to the [implementation of KernelInfoGetAttribute_string](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/custom_ops.cc#L59) : 
```cpp
ORT_API_STATUS_IMPL(OrtApis::KernelInfoGetAttribute_string, _In_ const OrtKernelInfo* info, _In_ const char* name, _Out_ char* out, _Inout_ size_t* size) {
  std::string value;
  auto status = reinterpret_cast<const onnxruntime::OpKernelInfo*>(info)->GetAttr<std::string>(name, &value);
  if (status.IsOK()) {
    if (out == nullptr) {  // User is querying the true size of the attribute
      *size = value.size() + 1;
      return nullptr;
    } 
    else ...
```

The fix should be something like:
```cpp
template <>
bool BaseKernel::TryToGetAttribute(const char* name, std::string& value) {
  if (info_ == nullptr) {
    ORT_CXX_API_THROW("Kernel was incorrectly initialized, pointer info_ cannot be null.", ORT_INVALID_ARGUMENT);
  }

  size_t size = 0;
  OrtStatus* status = api_.KernelInfoGetAttribute_string(info_, name, nullptr, &size);

  // The status should be a nullptr when querying for the size.
  if (status != nullptr) {
    api_.ReleaseStatus(status);
    return false;
  }

  ...
```

I can go ahead and create a PR with the fix.

PR #287, opened 2022-09-08
Issue #279: Fix bug that causes TryGetAttribute<std::string> to fail for valid inputs

adrianlizarraga:
Fixes https://github.com/microsoft/onnxruntime-extensions/issues/279

The `TryGetAttribute<std::string>()` function in [BaseKernel](https://github.com/microsoft/onnxruntime-extensions/blob/main/operators/ocos.cc#L56) does not correctly handle the return value of KernelInfoGetAttribute_string() when querying for the attribute's size (i.e., the 3rd argument is a nullptr). As a result, this function always returns false for any valid input. Note that this function is not currently used by any custom op.

I added a test that uses the TryGetAttribute function to get a string attribute from a node. Let me know if the test should be placed elsewhere, or if a simpler test can be used. The added ONNX models have the following structure (one has a string attribute and the other does not):
ort_ext_str_attr_test_model


wenbingl (approved):
Thanks for the fixing!

snnn (approved):

PR #290, opened 2022-09-12
Issue #288: Prevent copying of OrtApi struct

### LINKED ISSUE #288: OrtApi is currently passed by value
Hi,

Due to the following [merged ORT PR](https://github.com/microsoft/onnxruntime/pull/12297), passing `OrtApi` struct arguments by value will result in compilation errors _if/when_ the new C API header file is copied to this repository, which admittedly could be a long time from now. 

It might be a good idea to eventually update the various Kernel creation functions to use `const OrtApi& api` instead of `OrtApi api`.

If this is something you're interested in "fixing" now, I would be happy to create a PR that updates function signatures. Otherwise, please feel free to close this or keep it open for tracking purposes.

Thanks

adrianlizarraga:
Fixes #288 

Copying of `OrtApi` is now disabled in ORT. See https://github.com/microsoft/onnxruntime/pull/12297, which deletes the OrtApi copy constructor.

This PR replaces `OrtApi api` with `const OrtApi& api` in all custom ops and tests. 

**Question:**
I temporarily deleted the OrtApi copy constructor while testing this PR to ensure that all instances were handled. However, I ultimately reverted this change to the OrtApi struct because it seemed wrong to create a new version of the C API header that has never existed in any ORT release. Is this the correct choice, or should we just delete the copy constructor?


pyop/pyfunc.cc
```diff
@@ -377,7 +377,7 @@ void PyCustomOpDef::AddOp(const PyCustomOpDef* cod) {
   // No need to protect against concurrent access, GIL is doing that.
   auto val = std::make_pair(op_domain, std::vector<PyCustomOpFactory>());
   const auto [it_domain_op, success]  = PyOp_container().insert(val);
-  assert(success || it_domain_op->second.length() > 0);
+  assert(success || it_domain_op->first.length() > 0);
```
adrianlizarraga, line 380: This statement fails to compile in debug builds because `it_domain_op->second` is a `std::vector`, which does not have a `length()` method. From the surrounding context, I'm assuming that the assert is meant to check the domain string's length.

shaahji: Use `std::string::empty` to check for emptiness not `std::string::legnt() > 0`. The former is more readable than the latter.

shaahji:
> I temporarily deleted the OrtApi copy constructor while testing this PR to ensure that all instances were handled. However, I ultimately reverted this change to the OrtApi struct because it seemed wrong to create a new version of the C API header that has never existed in any ORT release. Is this the correct choice, or should we just delete the copy constructor?

The files under the include/onnxruntime folder are copies of the headers from onnxruntime release. Modifying them isn't recommended i.e. unless you are upgrading to a newer version of the onnxruntime. Since the OrtApi parameter in `CustomOpBase` is actually a const pointer which is being dereferenced for a call to OrtCustomOp::CreateKernel keeping it a const reference, as you did, is safe. The C/CPP api headers will get upgraded when they do and if the onnxruntime team fixes the issue, it will come along with that upgrade.

pyop/pyfunc.cc
wenbingl: Good catch.
The assertion checks either the op_domain already exists(the vector is not empty) , or it was newly created. So the 'length' should be replaced with 'size' of std::vector.

wenbingl:
> > I temporarily deleted the OrtApi copy constructor while testing this PR to ensure that all instances were handled. However, I ultimately reverted this change to the OrtApi struct because it seemed wrong to create a new version of the C API header that has never existed in any ORT release. Is this the correct choice, or should we just delete the copy constructor?
> 
> The files under the include/onnxruntime folder are copies of the headers from onnxruntime release. Modifying them isn't recommended i.e. unless you are upgrading to a newer version of the onnxruntime. Since the OrtApi parameter in `CustomOpBase` is actually a const pointer which is being dereferenced for a call to OrtCustomOp::CreateKernel keeping it a const reference, as you did, is safe. The C/CPP api headers will get upgraded when they do and if the onnxruntime team fixes the issue, it will come along with that upgrade.

Unfortunately, the C++ header and inline functions are keeping updating and not committed to be compatible among the versions. Only C header file are committed to be compatible. So the conclusion is the C header file is immutable, but the C++ header file should be taken as source code and extensions team has the liability to make it be better. for example, the prefast warnings in this file should be fixed.

adrianlizarraga:
Thanks @shaahji and @wenbingl.

The assert has been modified to check that the std::vector is not empty (in the event the item already exists in the map).

wenbingl (approved):
thanks!

snnn:
> Only C header file are committed to be compatible

It is compatible if you don't use what get changed.

adrianlizarraga:
> > Only C header file are committed to be compatible
> 
> It is compatible if you don't use what get changed.

Hi @snnn, I don't think I understand what you mean. Can you please clarify if you're suggesting an update? I'll hold off on merging. Thanks.

snnn:
Please merge it. My last comment wasn't quite relevant.

PR #292, opened 2022-09-15
Initial Java API for the JAR package.

wenbingl:


shaahji:
@wenbingl If extensions is meant to be built with onnxruntime, any reason why we need to support the dynamic library build? Shouldn't a static library build suffice? Any reason why we need to build a custom deployment package when onnxruntime already has a complete deployment support?

wenbingl:
> @wenbingl If extensions is meant to be built with onnxruntime, any reason why we need to support the dynamic library build? Shouldn't a static library build suffice? Any reason why we need to build a custom deployment package when onnxruntime already has a complete deployment support?

resolved in the internal discussions.

PR #295, opened 2022-09-23
OpenCV Image Decoder & SuperResolution CustomOps

shaahji:
OpenCV Image Decoder & SuperResolution CustomOps


cmake/externals/opencv.cmake
```diff
@@ -1,108 +1,124 @@
...
-set(BUILD_opencv_imgcodecs OFF CACHE INTERNAL "")
-set(BUILD_opencv_videoio OFF CACHE INTERNAL "")
-set(BUILD_opencv_calib3d OFF CACHE INTERNAL "")
-set(BUILD_opencv_highgui OFF CACHE INTERNAL "")
-set(BUILD_opencv_flann OFF CACHE INTERNAL "")
-set(BUILD_opencv_objdetect OFF CACHE INTERNAL "")
-set(BUILD_opencv_stitching OFF CACHE INTERNAL "")
-set(BUILD_opencv_ml OFF CACHE INTERNAL "")
-set(BUILD_opencv_world OFF CACHE INTERNAL "")
+set(BUILD_ANDROID_EXAMPLES      OFF CACHE INTERNAL "")
```
wenbingl, line 1: Did you format it manually, or by some tool?

tutorials/supres_e2e.py
```diff
@@ -0,0 +1,174 @@
...
+import torch
+import torch.nn as nn
+import torch.utils.model_zoo as model_zoo
+
+# ref: https://pytorch.org/tutorials/advanced/super_resolution_with_onnxruntime.html
+
+
+_this_dirpath = os.path.dirname(os.path.abspath(__file__))
+_data_dirpath = os.path.join(_this_dirpath, 'data')
+_onnx_model_filepath = os.path.join(_data_dirpath, 'supres.onnx')
```
wenbingl, line 15: It looks like there are two 'supres.onnx' are same? Any idea of deduplication?

operators/cv2/imdecode.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+struct KernelImageDecoder : BaseKernel {
+  KernelImageDecoder(const OrtApi& api) : BaseKernel(api) {}
+
+  void Compute(OrtKernelContext* context) {
+    // Setup inputs
+    const OrtValue* const inputs = ort_.KernelContext_GetInput(context, 0ULL);
+    OrtTensorDimensions dimensions(ort_, inputs);
+    if (dimensions.size() != 1ULL) {
+      throw std::runtime_error("Only raw image formats are supported.");
```
wenbingl, line 23: ORT_CXX_API_THROW is preferred for support NO-Exception mobile build.

operators/cv2/imdecode.hpp
```diff
@@ -0,0 +1,85 @@
...
+// Licensed under the MIT License.
+
+#pragma once
+
+#include <opencv2/core.hpp>
+#include <opencv2/imgproc.hpp>
+#include <opencv2/imgcodecs.hpp>
+
+#include "ocos.h"
+#include "string_utils.h"
```
wenbingl, line 11: is this file including needing?

wenbingl (dismissed):
Thanks for the work, Approved with some comments.

operators/cv2/super_resolution_preprocess.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+#include "ocos.h"
+
+struct KernelSuperResolutionPreProcess : BaseKernel {
+  KernelSuperResolutionPreProcess(const OrtApi& api);
+  void Compute(OrtKernelContext* context);
+};
+
+struct CustomOpSuperResolutionPreProcess : Ort::CustomOpBase<CustomOpSuperResolutionPreProcess, KernelSuperResolutionPreProcess> {
+  void* CreateKernel(const OrtApi& api, const OrtKernelInfo* info) const;
```
skottmckay, line 14: Is there a plan to make the pre-processing steps more general purpose? It's not going to scale to be adding model specific pre/post processing operators.

shaahji: I would let @wenbingl answer this one. As far as I know, the current design calls for one pre and one post custom operator only.

cmake/externals/opencv.cmake
shaahji: Sorted by tool and aligned manually.

tutorials/supres_e2e.py
shaahji: Yes. It's an oversight. I will remove the one in tutorial sub folder.

operators/cv2/super_resolution_preprocess.hpp
wenbingl: @joburkho @skottmckay Overall, this single op node of processing is recommended for the external user for their customization. we decided to drop the dependency on TorchScript in the last review meeting, which means there is no very convenient way to convert Python code into ONNX Script. e.g. offline processing model could be very difficult for the developer.  Technically, since OpenCV is already used, and we didn't think it will impact the binary size and the efficiency will be much better.

Internally, we want to support more models by the minimum number of operators. And it requires to check more model types to let us find the common parts, add some attributes and the script to compose the ONNX nodes. This PR is just a first step, stay tuned.

joburkho: @wenbingl : When we create a minimized build that should only contain the operators used by a set of models, this operator will not be included in that build if it is not used ... correct?  For instance, if a minimized build for Android uses some other opencv-related operators, but not this one, this operator will not exist in the final build ... right?

wenbingl: @joburkho , in the minimized build, all operators are disabled by default and would be enabled by adding item in this table.
https://github.com/microsoft/onnxruntime-extensions/blob/main/tools/gen_selectedops.py.  So it should be fine.
How to organize these vision ops is still under discussion so it would be too early to add this op in the table now.

tutorials/supres_e2e.py
shaahji: Done.

operators/cv2/imdecode.hpp
shaahji: Done.

operators/cv2/imdecode.hpp
shaahji: Yes. For MakeString.

operators/cv2/super_resolution_preprocess.hpp
skottmckay: As long as this 'first step' is considered a temporary operator I'm fine with it.

I don't understand why the user cares how many operators the pre/post processing tools insert into the model. I'm assuming we provide tooling to do the insertion. Having a single op per model won't scale. Having a single generic op for image preprocessing that has a large number of attributes is probably fine (e.g. input format (jpg, heic, bitmap), layout, channel ordering, crop, normalization values, target height and width, etc). Having multiple ops should also be fine given we could use ONNX for crop, center, normalization, transpose but would want a custom op to say convert from jpg.

wenbingl (approved):

operators/cv2/super_resolution_preprocess.hpp
wenbingl: > As long as this 'first step' is considered a temporary operator I'm fine with it.
> 
> I don't understand why the user cares how many operators the pre/post processing tools insert into the model. I'm assuming we provide tooling to do the insertion. Having a single op per model won't scale. Having a single generic op for image preprocessing that has a large number of attributes is probably fine (e.g. input format (jpg, heic, bitmap), layout, channel ordering, crop, normalization values, target height and width, etc). Having multiple ops should also be fine given we could use ONNX for crop, center, normalization, transpose but would want a custom op to say convert from jpg.

I think there is some misunderstanding here. For the external user who want to leverage onnxruntimae-extensions to customize their own pre-processing, we only recommend the single op solution for them, since ONNX convert tool is unavailable and the direct editing is error-prone.

For all pre- and post processing solution provided by us, it could be multiple ops with some helper script to edit ONNX-graph so that the user directly manipulate ONNX model. is this that you are talking about?

skottmckay: Yes. For pre/post processing we should provide tools that can add multiple ops to the model so it's not error prone, and any custom ops we add to the repo should be done at the most re-usable granularity. The user shouldn't necessarily need to know the individual ops to be used and whether they come from this repo or use ONNX operators. 

And if we have a tool that can add multiple ops for pre/post processing I would hope it is implemented in a general purpose way so it can be used to add any other individual custom op in the repo to the model.

PR #296, opened 2022-09-28
fix the C++ dangling pointer from the security check

wenbingl:


snnn (approved):
Thanks for the hard work!

PR #300, opened 2022-10-05
Partner team's code security warning fixings

wenbingl:


includes/ocos.h
```diff
@@ -70,19 +70,25 @@ struct OrtTensorDimensions : std::vector<int64_t> {
 template <typename... Args>
 class CuopContainer {
  public:
+#if defined(_MSC_VER) && !defined(__clang__)
```
joburkho, line 73: Shouldn't this just be around the `delete ocos_list_[i];` inside the for loop?

joburkho (approved):

includes/ocos.h
wenbingl: the 'new' statement below also needs to be suppressed.

joburkho: Generally, warning suppression in code should be scoped to the absolute minimum ... so we don't accidentally suppress a warning ... otherwise, why would you be using the push/pop.  The lambda containing the `new` would be reformatted to something like the following:
```
  CuopContainer() : ocos_list_({[]() {
#if defined(_MSC_VER) && !defined(__clang__)
#pragma warning(push)
#pragma warning(disable : 26409)
#endif
      return new Args;
#if defined(_MSC_VER) && !defined(__clang__)
#pragma warning(pop)
#endif
    }()...}) {
    ocos_list_.push_back(nullptr);
  }
```
... and the delete would be:
```
  CuopContainer() : ocos_list_({[]() {
#if defined(_MSC_VER) && !defined(__clang__)
#pragma warning(push)
#pragma warning(disable : 26409)
#endif
      return new Args;
#if defined(_MSC_VER) && !defined(__clang__)
#pragma warning(pop)
#endif
    }()...}) {
    ocos_list_.push_back(nullptr);
  }

    if (0 < ocos_list_.size()) {
      for (size_t i = 0; i < ocos_list_.size() - 1; i++) {
#if defined(_MSC_VER) && !defined(__clang__)
#pragma warning(push)
#pragma warning(disable : 26409)
#endif
          delete ocos_list_[i];
#if defined(_MSC_VER) && !defined(__clang__)
#pragma warning(pop)
#endif
      }
    }
```

wenbingl: @joburkho , your point makes sense. but this kind of suppression actually ruins the code readability.  So, the suppression makes the tool happy, and less of suppression makes the people (code readers) happy. It's kind of trade-off. :)

I will do this change in the coming PR since this one already passed in CI.

PR #312, opened 2022-11-08
Starter changes for supporting pre/post processing for vision models.

skottmckay:
- Script to update model with pre/post processing
- custom ops for decode/encode
  - user just has to provide jpg or png bytes
  - superresolution can return the updated image in jpg or png
- models for testing
  - not sure where the best place to add tests are - will add once that's established. have tested locally and end-to-end running updated model for both mobilenet and superresolution works.
  - also not sure what the expected test setup is, as building tests for the C++ custom ops seems to be disabled by default 

Updated cmake setup to enable building of the vision pre/post processing ops
  - opencv2 is treated as an internal dependency rather than the mechansim for selecting which operators to include.
  - clang-format has changed a bunch of spacing so easiest to ignore whitespace diffs for the changes


wenbingl:
> not sure where the best place to add tests are - will add once that's established. have tested locally and end-to-end running updated model for both mobilenet and superresolution works.
the code can be checked in ./onnxruntime-extensions and test in ./test folder.
https://github.com/microsoft/onnxruntime-extensions/blob/main/test/test_onnxprocess.py can be an example of the organization, though it is deprecated.



> also not sure what the expected test setup is, as building tests for the C++ custom ops seems to be disabled by default
Updated cmake setup to enable building of the vision pre/post processing ops

Enabling CTest needs ONNXRUNTIME_LIB_DIR, which is the folder libonnxruntime.so, it is already setup in CI pipeline. On local development, you need specify it on build.

> TODO: Can we use a domain like 'com.microsoft.extensions' for these new ops instead of 'ai.onnx.contrib'? that would be more consistent with ORT contrib ops (which use com.microsoft). currently official onnx ops use the 'ai.onnx' prefix. I can address supporting multiple domains in a separate PR if there's no objection to using this domain name.

The new name looks better, but the existing domain may needs to be kept to avoid breaking the Office models.
We may support both domains? It needs the changes in many places like build, tool, API and test...

wenbingl:
BTW, extensions supports editable pip install, i.e. pip install -e .. The setup will compile the C++ source and install so and python module in-place, so all consequential changes on python files will take effort immediately, without any re-installation.

edgchen1 (commented):
some initial comments, will finish reviewing tools/pre_post_processing later

includes/ocos.h
```diff
@@ -57,55 +59,42 @@ struct OrtTensorDimensions : std::vector<int64_t> {
...
-  const OrtCustomOp** GetList() {
-    return &const_cast<const OrtCustomOp*&>(ocos_list_.front());
+  const std::vector<const OrtCustomOp*>& GetCustomOps() const {
+    return ocos_list_;
   }
 
  private:
-  std::vector<OrtCustomOp*> ocos_list_;
+  std::vector<const OrtCustomOp*> ocos_list_;
+  std::vector<std::shared_ptr<OrtCustomOp>> op_instances_;  // use shared_ptr to capture type specific deleter
```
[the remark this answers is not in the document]
skottmckay: This can hold any class derived from OrtCustomOp which is a C struct with no virtual destructor. I believe using shared_ptr will capture the deleter for the 'real' class and might be the only option in std for ensuring correct cleanup whilst storing the pointer in a generic vector containing OrtCustomOp instances.

includes/ocos.h
```diff
@@ -57,55 +59,42 @@ struct OrtTensorDimensions : std::vector<int64_t> {
...
 };
 
-struct CustomOpClassBegin{
+struct CustomOpClassBegin {
 };
 
-typedef std::function<const OrtCustomOp**()> FxLoadCustomOpFactory;
+using FxLoadCustomOpFactory = std::function<const std::vector<const OrtCustomOp*>&()>;
 
 template <typename _Begin_place_holder, typename... Args>
```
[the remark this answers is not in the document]
skottmckay: I believe it handles the 'no classes in list' scenario. maybe the name is slightly misleading and it's more a 'dummy void type'

operators/vision/decode_image.cc
```diff
@@ -0,0 +1,41 @@
...
+  };
+
+  // Setup output & copy to destination
+  const cv::Size decoded_image_size = decoded_image.size();
+  const int64_t colors = 3;
+
+  const std::vector<int64_t> output_dimensions{decoded_image_size.height, decoded_image_size.width, colors};
+  OrtValue* output_value = ort_.KernelContext_GetOutput(context, 0, output_dimensions.data(), output_dimensions.size());
+  uint8_t* decoded_image_data = ort_.GetTensorMutableData<uint8_t>(output_value);
+  memcpy(decoded_image_data, decoded_image.data, decoded_image.total() * decoded_image.elemSize());
```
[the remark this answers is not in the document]
skottmckay: Yes. Would height * width * colors be clearer? I believe 'total()' returns 'height * width'

operators/vision/decode_image.hpp
```diff
@@ -0,0 +1,57 @@
...
+  size_t GetInputTypeCount() const {
+    return 1;
+  }
+
+  ONNXTensorElementDataType GetInputType(size_t index) const {
+    switch (index) {
+      case 0:
+        return ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT8;
+      default:
+        ORT_CXX_API_THROW(MakeString("Invalid input index ", index), ORT_INVALID_ARGUMENT);
```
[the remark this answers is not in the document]
skottmckay: I agree something more like the ORT_THROW behavior would be nice. Will leave for separate PR as I assume that would touch a lot of places.

operators/vision/encode_image.cc
```diff
@@ -0,0 +1,43 @@
...
+
+void KernelEncodeImage ::Compute(OrtKernelContext* context) {
+  // Setup inputs
+  const OrtValue* input_bgr = ort_.KernelContext_GetInput(context, 0ULL);
+  const OrtTensorDimensions dimensions_bgr(ort_, input_bgr);
+
+  if (dimensions_bgr.size() != 3 || dimensions_bgr[2] != 3) {
+    // expect {H, W, C} as that's the inverse of what decode_image produces.
+    // we have no way to check if it's BGR or RGB though
+    throw std::runtime_error("[EncodeImage] requires rank 3 BGR input in channels last format.");
```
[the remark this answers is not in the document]
skottmckay: I believe so. IIRC there's 1P usage in a minimal build with no exceptions. 

Will update to use the throw macro (copied this from the superres custom op and forgot to update this line).

test/shared_test/test_ortops_vision.cc
```diff
@@ -0,0 +1,79 @@
...
+    if (std::filesystem::exists(data_dir) && std::filesystem::is_directory(data_dir)) {
+      break;
+    }
+
+    cur = cur.parent_path();
+    ASSERT_NE(cur, cur.root_path()) << "Reached root directory without finding 'data' directory.";
+  } while (true);
+
+  // set current path as the extensions library is also loaded from that directory by TestInference
+  std::filesystem::current_path(cur);
```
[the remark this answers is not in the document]
skottmckay: Probably - but it's not used anywhere else currently and only changes the directory if running via Google Test Adapter in Visual Studio.

Was going to find a more central place to do it if other PPP custom ops require this sort of handling to be able to be debugged that way.

test/shared_test/test_ortops_vision.cc
```diff
@@ -0,0 +1,79 @@
...
+// Test DecodeImage and EncodeImage by providing a jpg image. Model will decode to BGR, encode to PNG and decode
+// again to BGR. We validate that the BGR output from that matches the original image.
+TEST(VisionOps, image_decode_encode) {
+  FixCurrentDir();
+
+  auto data_dir = std::filesystem::current_path() / "data";
+  auto model_path = data_dir / "ppp_vision" / "decode_encode_decode_test.onnx";
+  auto image_path = data_dir / "test_colors.jpg";
+
+  auto ort_env = std::make_unique<Ort::Env>(ORT_LOGGING_LEVEL_WARNING, "Default");
```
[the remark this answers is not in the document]
skottmckay: Copied pattern from other unit tests here. I don't think so - if anything there should be one central Ort::Env in the overall test infrastructure like we do with the ORT tests. Can do that in a separate PR.

tools/pre_post_processing/Example usage of the PrePostProcessor.md
```diff
@@ -0,0 +1,205 @@
...
+
+
+## Initial imports
+
+Import the PrePostProcessor, the steps and a utility to simplify creating new model inputs.
+
+```py
+import onnx
+from pre_post_processing import PrePostProcessor
+from pre_post_processing.Steps import *
```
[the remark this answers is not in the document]
skottmckay: directory name has capital S. can make lower case.

shared/lib/ortcustomops.cc
```diff
@@ -97,28 +98,31 @@ extern "C" ORTX_EXPORT OrtStatus* ORT_API_CALL RegisterCustomOps(OrtSessionOptio
   static std::vector<FxLoadCustomOpFactory> c_factories = {
     LoadCustomOpClasses<CustomOpClassBegin>
 #if defined(ENABLE_TF_STRING)
-    , LoadCustomOpClasses_Text
-#endif // ENABLE_TF_STRING
+    ,
```
wenbingl, line 101: Any reason that the comma and the class name are separated into two lines? The existing one looks like more compact.

shared/lib/ortcustomops.cc
```diff
@@ -6,6 +6,7 @@
 
 #include "onnxruntime_extensions.h"
 #include "ocos.h"
+#include "vision/vision.hpp"
```
wenbingl, line 9: Did you check whether if the entire source code related to PPP_vision will be excluded if PPP_VISION was disabled by CMAKE argument?

skottmckay: If it's disabled the factory will have no classes and the custom ops will have no usage. When that happens the linker should be able to throw them away. At least that's how it works in ORT where we just remove the registration of kernels when doing a reduced ops build (we don't need to remove the code for the kernels).

If you think something more is required I can add it.

shared/lib/ortcustomops.cc
skottmckay: clang-format did that. Are we not using the same coding conventions and standards as for ORT?

CMakeLists.txt
```diff
@@ -45,6 +44,11 @@ option(OCOS_BUILD_PYTHON "Enable building the Python package" OFF)
 option(OCOS_BUILD_JAVA "Enable building the Java package" OFF)
 option(OCOS_BUILD_ANDROID "Enable building the Android package" OFF)
 
+# Pre/post processing operator collections.
+option(OCOS_ENABLE_PPP_VISION "Enable operators for adding pre/post processing to vision models" ON)
```
wenbingl, line 48: The existing options are created for selected operator build since the 3rd party libraries binary take most of proportion of footprint. It meets the footprint requirements while avoiding #ifdef flood in the source code.

We can group these options to create the domain package, instead of adding of these new options which is not same granularity.

skottmckay: I also don't want excessive `#ifdef`s in the code. 

If I look at the existing tokenizer options for example, OCOS_ENABLE_PPP_TEXT could simply turn on the required tokenizer options (e.g. turn on OCOS_ENABLE_BERT_TOKENIZER but turn off the GPT2 tokenizer as nobody is going to be running GPT2 on mobile) without adding any new ifdefs. i.e. all the handling for it is at the cmake level

For vision I saw there's an existing OPENCV option, but 'opencv' seems like an internal dependency/implementation detail. Can a user instead pick functionality (e.g. 'vision') and based on that we enable the required dependencies like opencv? 

Do the existing operators in the operators\cv2 have any production usage currently? I don't need any of them for the scenarios we're targeting.

shared/lib/ortcustomops.cc
wenbingl: The problem is the 3rd party library will not be fetched when some operators are disabled. So, the file including statement will cause some compiling issue. This is different from ORT since it uses submodule.

BTW, I am not sure whether there will be some binary increase if we just include a 3rd party library and do not reference any functions from it. Is C++ compiler smart enough to handle the global variable issue in that case?

includes/ocos.h
```diff
@@ -3,19 +3,21 @@
...
 #undef ORT_API_MANUAL_INIT
 
-
 // A helper API to support test kernels.
 // Must be invoked before RegisterCustomOps.
 extern "C" bool ORT_API_CALL AddExternalCustomOp(const OrtCustomOp* c_op);
 
-const char c_OpDomain[] = "ai.onnx.contrib";
+constexpr const char* c_OpDomain = "ai.onnx.contrib";
+constexpr const char* c_ComMsExtOpDomain = "com.microsoft.extensions";
```
wenbingl, line 20: This domain name looks better but I don't know the way we can migrate all the user's existing model to this new name?  Do you have a solution for this? Or we move this change to another PR to address the issue completely?

skottmckay: I'm only using the new domain name for the new ops currently. For additional things we want to include in the packages (e.g. BERT tokenizer) we can register in both domains so existing models still work. 

Does that approach work for you?

shared/lib/ortcustomops.cc
skottmckay: I can wrap the include of the vision.hpp header in an ifdef if you want. Shouldn't matter though. There are no external includes in the vision custom op headers. I assumed there wouldn't be any compile issues from including ocos.h or string_utils.h. Is that the case? 

As for the binary size it might depend on the 3rd party library implementation as to whether the linker can throw it all away. Easily tested once we create a package with the desired operators. Ideally the cmake setup isn't adding unused 3rd party libraries to the link list. If it is it would be better to fix that.

operators/vision/decode_image.cc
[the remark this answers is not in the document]
edgchen1: got it. I think that would be clearer, or setting output_dimensions to `[height, width, elemSize]`

edgchen1 (commented):
some more comments, will finish up python review in the next round

test/shared_test/test_ortops_vision.cc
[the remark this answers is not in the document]
edgchen1: sure, ok to change it later

tools/pre_post_processing/docs/pre_post_processing/Steps/general.md
```diff
@@ -0,0 +1,83 @@
...
+
+    * pre_post_processing.step.Step
+
+`Softmax(name: str = None)`
+:   ONNX Softmax
+    
+    Initialize the step.
+    
+    Args:
+        inputs: List of default input names.
```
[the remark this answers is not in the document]
skottmckay: Looks like it's inheriting from the base class when not explicitly documented which is a little confusing. Will add explicit doco.

tools/pre_post_processing/docs/pre_post_processing/Steps/vision.md
```diff
@@ -0,0 +1,155 @@
...
+    
+    NOTE: Input image format is inferred and does not need to be specified.
+
+    ### Ancestors (in MRO)
+
+    * pre_post_processing.step.Step
+
+`FloatToImageBytes(multiplier: float = 255.0, name: str = None)`
+:   Converting floating point values to uint8 values in range 0..255.
+    Typically this reverses ImageBytesToFloat by converting input data in the range 0..1, but an optional multiplier
```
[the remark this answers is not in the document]
skottmckay: Currently it's used to not change the range of the Cb/Cr data in super resolution as that never got moved from 0..255 to 0..1 (just gets converted to float to resize). Primarily I want the round + clip + cast functionality in that case to make sure it's a valid and accurate RGB value. 

Not sure if there will be any other use cases than a multiplier of 255 or 1 to warrant complicating things more.  Could change it from 'multiplier' to something less flexible though.

tools/add_pre_post_processing_to_model.py
```diff
@@ -0,0 +1,218 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import argparse
+import enum
+import os
+
+#
+# from dataclasses import dataclass
```
edgchen1, line 9: remove comment?

tools/add_pre_post_processing_to_model.py
```diff
@@ -0,0 +1,218 @@
...
+            FloatToImageBytes(multiplier=1.0, name="Resized_Cr"),
+            # as we're selecting outputs from multiple previous steps we need to map them to the inputs using step names
+            (
+                YCbCrToPixels(layout="BGR"),
+                [
+                    IoMapEntry("Yout_to_bytes", 0, 0),  # uint8 Y' with shape {h, w}
+                    IoMapEntry("Resized_Cb", 0, 1),  # uint8 Cb'
+                    IoMapEntry("Resized_Cr", 0, 2),
+                ],
+            ),  # uint8 Cr'
```
edgchen1, line 154: looks like this comment is on the wrong line

tools/add_pre_post_processing_to_model.py
```diff
@@ -0,0 +1,218 @@
...
+
+    parser.add_argument("model", type=Path, help="Provide path to ONNX model to update.")
+
+    args = parser.parse_args()
+
+    model_path = args.model.resolve(strict=True)
+    new_model_path = model_path.with_suffix(".with_pre_post_processing.onnx")
+
+    if args.model_type == "mobilenet":
+        source = ModelSource.PYTORCH if args.model_source == "pytorch" else "tensorflow"
```
edgchen1, line 211:
```suggestion
        source = ModelSource.PYTORCH if args.model_source == "pytorch" else ModelSource.TENSORFLOW
```

edgchen1 (commented):
done with first pass, generally looks good to me

tools/add_pre_post_processing_to_model.py
skottmckay: The joys of 'black'. Most of its formatting is good. Some is dubious.

tools/pre_post_processing/pre_post_processor.py
```diff
@@ -0,0 +1,322 @@
...
+
+        for item in items:
+            step = None
+            explicit_io_map_entries = None
+
+            if isinstance(item, Step):
+                step = item
+            elif isinstance(item, tuple):
+                step_or_str, explicit_io_map_entries = item
+                step = self.__producer_from_step_or_str(step_or_str)  # throws if not found
```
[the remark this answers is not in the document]
skottmckay: I wanted to be explicit that this was the producer of the value and as such must exist in a previous step.

tools/pre_post_processing/pre_post_processor.py
```diff
@@ -0,0 +1,322 @@
...
+        """
+        Add the pre/post processing steps and join with existing steps.
+
+        Args:
+            processors: List of processors to add items to.
+            processor_connections: Manual connections to create between the pre/post processing graph and the model.
+            items: Items to add to processors.
+                   Can be:
+                     A Step instance. This will be implicitly joined to the immediately previous Step if one exists.
+                     A tuple of (Step or step name, list of IoMapEntry)
```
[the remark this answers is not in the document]
skottmckay: Will clarify. The lookup by name is for values in the IoMapEntry.

CMakeLists.txt
wenbingl: Adding a new CMake option criteria is whether the new one or some operators increase the footprint significantly. 
The new options you are adding here seem to be a higher-level conception, like option group, which is what I mean 'not same granularity'. 
We can create the group option in CMakeLists.txt technically, but it would be better to create some separated files to specify which options should be turned on for the certain domains. So, we still can keep a simple and straightforward cmake option design pattern.

operators/vision/vision.hpp
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+
+#include "ocos.h"
+#include "decode_image.hpp"
+#include "encode_image.hpp"
+
+FxLoadCustomOpFactory LoadCustomOpClasses_PPP_Vision =
```
wenbingl, line 10: With these new operators, can we remove the previous operators checked in from super-resolution model?

wenbingl (approved):

tools/pre_post_processing/docs/pre_post_processing/Steps/general.md
```diff
@@ -0,0 +1,70 @@
+Module pre_post_processing.steps.general
```
edgchen1, line 1: nit: can we also rename this directory from "Steps" to "steps"?

tools/pre_post_processing/steps/general.py
```diff
@@ -0,0 +1,207 @@
...
+
+        return reverse_graph
+
+
+class Squeeze(Step):
+    """
+    ONNX Squeeze
+    """
+
+    def __init__(self, axes: List[int] = None, name: Optional[str] = None):
```
edgchen1, line 66:
```suggestion
    def __init__(self, axes: Optional[List[int]] = None, name: Optional[str] = None):
```

tools/pre_post_processing/pre_post_processor.py
```diff
@@ -0,0 +1,334 @@
...
+)
+from .step import Step
+
+
+class PrePostProcessor:
+    """
+    Class to handle running all the pre/post processing steps and updating the model.
+    """
+
+    def __init__(self, inputs: List[onnx.ValueInfoProto] = None, outputs: List[onnx.ValueInfoProto] = None):
```
edgchen1, line 24:
```suggestion
    def __init__(self, inputs: Optional[List[onnx.ValueInfoProto]] = None, outputs: Optional[List[onnx.ValueInfoProto]] = None):
```

shared/lib/ortcustomops.cc
edgchen1: could we instead do something like:
```cpp
  static std::vector<FxLoadCustomOpFactory> c_factories = {
    LoadCustomOpClasses<CustomOpClassBegin>,
#if defined(ENABLE_TF_STRING)
    LoadCustomOpClasses_Text,
#endif  // ENABLE_TF_STRING
    ...,
  };
```

wenbingl: I remember if all of operators are disabled, the last comma after <CustomOpClassBegin> fails the build.  Maybe we can add <CustomOpClassEnd> to solve that.

tools/pre_post_processing/Example usage of the PrePostProcessor.md
natke, line 16: Need to update these markdown instruction with steps instead of Steps

shared/lib/ortcustomops.cc
skottmckay: One alternative would be to change the factory setup so the factory has a method to return the shared_ptr instances for its enabled ops so we're not using a template unpack to expand the complete set of class names here. 

```c++
std::vector<std::shared_ptr<OrtCustomOp>> GetCustomOps() {
  std::vector<std::shared_ptr<OrtCustomOp>> ops;

#ifdef ENABLE_1
  ops.push_back(std::make_shared<CustomOp1>());
#endif

#ifdef ENABLE_2
  ops.push_back(std::make_shared<CustomOp2>());
#endif

  return ops;
}
```

operators/vision/vision.hpp
skottmckay: Yes. Rachel has been updating an example app that currently uses them. Once we switch that over I will remove them.

includes/ocos.h
skottmckay: Is there documentation somewhere on current usage of ops by partner to have an idea of who's using what? 

A script to update the domain in an existing model is trivial so easily added.

CMakeLists.txt
skottmckay: I don't care if the grouping is done in this file or some external file that specifies a bunch of cmake options. Assuming we do the latter, I can remove the OCOS_ENABLE_PPP_* values.

Not clear how I should structure it to provide a way to select the operators for vision pre/post processing though, which is just the 2 new ones. I don't want the existing ones in the cv2 folder that are currently enabled by OCOS_ENABLE_OPENCV. Can you please provide a suggestion of what you have in mind?

wenbingl: Adding a new flag of OCOS_ENABLE_OPENCV_IMG or some other name? A OPENCV in the middle of the name means it depends on OPENCV library, actually there are already OCOS_ENABLE_OPEN_CODEC, we may reuse that one and make any change you want since they are not using in any product yet.

includes/ocos.h
```diff
@@ -119,12 +108,12 @@ extern FxLoadCustomOpFactory LoadCustomOpClasses_Math;
...
-#endif // ENABLE_TOKENIZER
+#endif  // ENABLE_TOKENIZER
 
 #ifdef ENABLE_TF_STRING
 extern FxLoadCustomOpFactory LoadCustomOpClasses_Text;
 #endif  // ENABLE_TF_STRING
 
-#ifdef ENABLE_OPENCV
-extern FxLoadCustomOpFactory LoadCustomOpClasses_OpenCV;
+#ifdef ENABLE_CV2
```
wenbingl, line 117: Why don't we follow this if/def patterns to manage all enable_op flags in one unified header?

Defining an extern variable here is to avoid including each operator file in ort_customops.cc.

shared/lib/ortcustomops.cc
wenbingl: In the old ort C++ header files, some methods of CustomOp are non-const, so the class has to be created dynamically. But now, with some fixing, the custom op object can be a constant object.  We may solve this problem in by the c++ initializer list, and we may do it later.

includes/ocos.h
skottmckay: Are we using the extern because we're concerned about the impact on build time by including a header for the factory that includes the declaration of the individual operators? Or are there other negative consequences of the operator declarations begin included in ort_customops.cc?

Maybe @edgchen1 can comment on best practices for usage of `extern` in C++ code. Historically I've avoided it unless necessary as `extern` statements can get out of sync, but maybe I'm overlooking some benefits of it.

CMakeLists.txt
```diff
@@ -394,7 +414,7 @@ foreach(nf ${NO_USE_FILES})
 endforeach()
 
 # test section
-if (OCOS_ENABLE_CTEST)
+if(OCOS_ENABLE_CTEST AND NOT OCOS_ENABLE_SELECTED_OPLIST)
```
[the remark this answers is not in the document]
skottmckay: will add

CMakeLists.txt
```diff
@@ -174,11 +175,23 @@ if (OCOS_ENABLE_MATH)
   list(APPEND TARGET_SRC ${TARGET_SRC_MATH} ${TARGET_SRC_DLIB} ${TARGET_SRC_INVERSE})
 endif()
 
-if (OCOS_ENABLE_OPENCV)
+# enable the opencv dependency if we have ops that require it
+if (OCOS_ENABLE_CV2 OR OCOS_ENABLE_VISION)
+  set(_ENABLE_OPENCV ON)
+  set(_ENABLE_OPENCV_CODECS ON)  # TODO: Do we have any real use cases where we need to disable codecs?
```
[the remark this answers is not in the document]
skottmckay: The only op that is enabled without it is gaussian blur. What is the use case for that in inferencing? Not clear why you'd want to blur something in that scenario.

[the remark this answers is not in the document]
wenbingl: Couples of issues related to the codecs library, potential license issue, and seems not works well on ARM64EC, may be WebAssembly either. So, we need to a way to disable it while they can still use openCV.

includes/ocos.h
wenbingl: Because we don't have a set test to verify enable/disable CMake options,  I avoid including all header files in one file(ort-customops.cc) to detected unintentionally macro/function reference among these operator implementations, build speed is a bonus.

The prototype sync is a protentional issue in some cases, which in this header, we can only use a name, and out-of-sync will cause link error, so it should be good.

CMakeLists.txt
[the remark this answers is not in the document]
skottmckay: Added back for potential future usage of opencv APIs that do not have dependencies on the codecs.

includes/ocos.h
edgchen1: will note that the Google C++ style guideline prefers including headers to forward declarations:
https://google.github.io/styleguide/cppguide.html#Forward_Declarations

we (aim to) follow these guidelines in onnxruntime, not sure what the coding standards for onnxruntime-extensions are

wenbingl (approved):
Thanks for the contribution!

edgchen1 (approved):

PR #313, opened 2022-11-17
fix wasm ort builds

guschmue:


snnn (approved):

CMakeLists.txt
```diff
@@ -323,7 +323,7 @@ target_compile_definitions(ocos_operators PRIVATE ${OCOS_COMPILE_DEFINITIONS})
 target_link_libraries(ocos_operators PRIVATE ${ocos_libraries})
 
 file(GLOB shared_TARGET_LIB_SRC "shared/lib/*.cc" "shared/lib/*.h")
-if(CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
+if(NOT OCOS_ENABLE_STATIC_LIB AND CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
```
wenbingl, line 326: IIRC, there is only one static library target for Emscripten, so, those build flags are bad now?

wenbingl: So, the following target is only for WASM build pass verification only.

natke:
Will this PR fix the CI failure in #312?

CMakeLists.txt
edgchen1, line 326: curious, where is OCOS_ENABLE_STATIC_LIB actually used?

I only see one text match where the option is defined:
```
onnxruntime-extensions % git grep -n OCOS_ENABLE_STATIC_LIB
CMakeLists.txt:42:option(OCOS_ENABLE_STATIC_LIB "Enable generating static library" OFF)
```

wenbingl: It's kind of obsoleted now, since the static-library always be generated. In some version, dynamic-library or static-library are exclusive option that user only choose one. @edgchen1

edgchen1: I see. if we keep using it, can we come up with a clearer name than `OCOS_ENABLE_STATIC_LIB`?
for example, in the case where `CMAKE_SYSTEM_NAME != "Emscripten"`, a static library is generated even if `OCOS_ENABLE_STATIC_LIB == OFF`, which seems unexpected.

PR #315, opened 2022-11-18
Update OneBranch.Official.yml for Azure Pipelines

wenbingl:


snnn:
Please replicate the change to the pull request pipeline yaml file: https://github.com/microsoft/onnxruntime-extensions/blob/main/.pipelines/OneBranch.PullRequest.yml , so that it can be run before this PR is merged.

snnn:
Your pull request yaml file has grammar error.

.pipelines/OneBranch.Official.yml
```diff
@@ -54,8 +54,7 @@ extends:
     - stage: linux_stage
       jobs:
       - job: linux_job
-        pool:
-          type: linux
+        pool: {vmImage: 'ubuntu-latest', name: 'Linux-CPU'}
```
snnn, line 57: No... you need to use their machine pool.

snnn: Ah... probably it is because the job itself is already in docker.

wenbingl: can we ask onebranch team for a solution, this is a very common problem.

snnn: I will give you their support mail list.

wenbingl:
temporarily close this PR before we can find a solution for docker for Linux.

PR #316, opened 2022-11-24
Minor opencv tweaks

skottmckay:
* Separate ops in operators/cv2 that do and do not require codecs so they're easier to include/exclude from a build.

* Don't comment out the opencv options that we enable when building with codecs
  - commenting them out results in them not being unset as expected if you change build settings

* Remove jpeg2000 from opencv file formats. 
  - It costs 1MB and is (afaict) not a common format.

* Add ability to enable/disable cv2 ops to gen_selectedops.py.


operators/cv2/opencv.cc
```diff
@@ -1,10 +1,10 @@
 #include "ocos.h"
-#include "gaussian_blur.hpp"
+#include "no_codecs/gaussian_blur.hpp"
```
wenbingl, line 2: let's use opencv module name - 'imgproc' here. 'no_codecs' sounds a little weird.

operators/cv2/opencv.cc
```diff
@@ -1,10 +1,10 @@
...
 #ifdef ENABLE_OPENCV_CODECS
-#include "imread.hpp"
-#include "imdecode.hpp"
-#include "super_resolution_preprocess.hpp"
-#include "super_resolution_postprocess.hpp"
+#include "codecs/imread.hpp"
+#include "codecs/imdecode.hpp"
+#include "codecs/super_resolution_preprocess.hpp"
```
wenbingl, line 6: Do we still need them, or we can get rid of them now? There is no production usage right now and was only developed for Mobile.

skottmckay: I'll remove the super_resolution_* ones as they aren't needed. Do you want imread/imdecode left or deleted as well?

wenbingl: probably we can undo the previous PR to clean up test and tutorial files as well in another PR?

there are some nice python tests based on imread/imdecodec that help evaluating the openCV library. Let's keep it for a while.

wenbingl (approved):

edgchen1 (approved):

PR #318, opened 2022-11-28
Add build.py to make it easier for developers to build different variants

skottmckay:
Add python based build infrastructure to simplify developer builds for various platforms. Majority was copied from the ORT build script so usage is consistent with that.

Left the existing build.bat/build.sh but ideally the CI can be updated to use the new infrastructure so things are more consistent.

Updated gradle to 7.5.1 and Android gradle tools to 7.3.0.

Validated Windows and cross-compiling Android on Windows including builds with explicitly selected ops. WASM and iOS builds aren't tested yet and might need minor tweaks.


wenbingl:
> 

ort-extensions avoids any possible build dependency to make different platform developer be easier to use it (only cmake and native script command used). Python is great but not all developers are familiar with it, so we may put it to be optional for some complicated CI jobs. And still need to keep build scripts w/o Python be available for the developers.

tools/utils/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+from .logger import get_logger
```
wenbingl, line 4: are all of these scripts same with ORT? can we download it from ORT repo.
There always are some accusations of using the same code from ORT.

java/build-android.gradle
```diff
@@ -27,13 +27,14 @@ project.group = "com.microsoft.onnxruntime"
...
 def defaultDescription = 'ONNXRuntime-Extensions is an extension of onnxruntime to support pre- and post-processing.'
 
+
 buildscript {
 	repositories {
 		google()
 		mavenCentral()
 	}
 	dependencies {
-		classpath 'com.android.tools.build:gradle:4.0.1'
```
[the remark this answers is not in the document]
skottmckay: Most likely it just hasn't been updated for a long time. I did start looking at doing that but haven't had time to finish doing so. https://github.com/microsoft/onnxruntime/pull/13093/ had an initial set of changes. Need to use an older JDK to make Android happy though.

tools/utils/__init__.py
skottmckay: Largely the same, but downloading ORT in order to use a couple of helper scripts feels like a really heavy dependency. If it was production code it would be worth sharing the code though.

skottmckay:
> ort-extensions avoids any possible build dependency to make different platform developer be easier to use it (only cmake and native script command used). Python is great but not all developers are familiar with it, so we may put it to be optional for some complicated CI jobs. And still need to keep build scripts w/o Python be available for the developers.

No problem. Can leave existing scripts as-is and provide this as an additional option.

In the last couple of days I've built different configs (Debug, Release, RelWithDebInfo and MinSizeRel for various reasons) both for Windows, Linux, and cross-compiling Android on Windows with different NDKs and ABIs, with different output directories to enable diffing the output and the python script made it a lot quicker and simpler to do that.

tools/build.py
```diff
@@ -0,0 +1,687 @@
...
+class UsageError(Exception):
+    def __init__(self, message: str):
+        super().__init__(message)
+
+
+def _check_python_version():
+    # Require python 3.6+
+    if sys.version_info[0] != 3:
+        raise UsageError("Bad python major version: expecting python 3, found version " "'{}'".format(sys.version))
+    if sys.version_info[1] < 6:
```
edgchen1, line 33: 3.6 is no longer supported, we could update this to (at least) 3.7

CMakeLists.txt
```diff
@@ -412,11 +415,11 @@ endif()
...
 # clean up the requirements.txt files from 3rd party project folder to suppress the code security false alarms
 file(GLOB_RECURSE NO_USE_FILES ${CMAKE_BINARY_DIR}/_deps/*requirements.txt)
-message("Found the follow requirements.txt: ${NO_USE_FILES}")
+message("Found the following requirements.txt: ${NO_USE_FILES}")
 foreach(nf ${NO_USE_FILES})
     file(TO_NATIVE_PATH ${nf} nf_native)
-    if (CMAKE_SYSTEM_NAME MATCHES "Windows")
-        execute_process(COMMAND cmd /c "del ${nf_native}")
+    if (CMAKE_HOST_SYSTEM_NAME MATCHES "Windows")
+        execute_process(COMMAND cmd /c del ${nf_native})
```
edgchen1, line 422: could use "cmake -E remove" to avoid platform specific commands

tools/build.py
```diff
@@ -0,0 +1,687 @@
...
+                        choices=["Debug", "MinSizeRel", "Release", "RelWithDebInfo"],
+                        help="Configuration(s) to build.")
+    parser.add_argument("--update", action="store_true", help="Update submodules and makefiles.")
+    parser.add_argument("--build", action="store_true", help="Build.")
+    parser.add_argument("--test", action="store_true", help="Run unit tests.")
+
+    parser.add_argument("--clean", action="store_true",
+                        help="Run 'cmake --build --target clean' for the selected config/s.")
+
+    parser.add_argument("--skip_submodule_sync", action="store_true",
```
edgchen1, line 83: probably don't need this if there are no submodules

tools/build.py
```diff
@@ -0,0 +1,681 @@
+#!/usr/bin/env python3
```
edgchen1, line 1: would be good if we can prune this further to what we want to support in ORT extensions. ideally we can inherit only as much complexity as we need from ORT's build.py

tools/build.py
```diff
@@ -0,0 +1,681 @@
...
+            if args.use_xcode:
+                cmake_extra_args += ["-G", "Xcode"]
+            if not args.ios and not args.android and args.osx_arch == "arm64" and platform.machine() == "x86_64":
+                if args.test:
+                    log.warning("Cannot test ARM64 build on X86_64. Will skip test running after build.")
+                    args.test = False
+
+        if args.build_wasm:
+            emsdk_version = args.emsdk_version
+            emsdk_dir = REPO_DIR / "cmake" / "external" / "emsdk"
```
edgchen1, line 640: this path doesn't exist. do we need `--build_wasm`?

tools/build.py
```diff
@@ -0,0 +1,681 @@
...
+
+    parser.add_argument("--enable_wasm_debug_info", action="store_true",
+                        help="Build WebAssembly with DWARF format debug info")
+
+    parser.add_argument("--emsdk_version", default="3.1.19", help="Specify version of emsdk")
+    parser.add_argument("--emscripten_settings", nargs="+", action="append",
+                        help="Extra emscripten settings to pass to emcc using '-s <key>=<value>' during build.")
+
+    # x86 args
+    # TODO: Are these needed?
```
edgchen1, line 173: if not, let's remove it

tools/build.py
```diff
@@ -0,0 +1,681 @@
...
+    log.info("Generating CMake build tree")
+
+    cmake_args = [
+        str(cmake_path),
+        str(source_dir),
+        # There are two ways of locating python C API header file. "find_package(PythonLibs 3.5 REQUIRED)"
+        # and "find_package(Python 3.5 COMPONENTS Development.Module)". The first one is deprecated and it
+        # depends on the "PYTHON_EXECUTABLE" variable. The second needs "Python_EXECUTABLE". Here we set both
+        # of them to get the best compatibility.
+        "-DPython_EXECUTABLE=" + sys.executable,
```
edgchen1, line 312: do we need to define: Python_EXECUTABLE, PYTHON_EXECUTABLE, OCOS_CROSS_COMPILING?

tools/build.py
```diff
@@ -0,0 +1,681 @@
...
+    configs = set(args.config)
+
+    # setup paths and directories
+    # cmake_path can be None. For example, if a person only wants to run the tests, they don't need cmake.
+    cmake_path = _resolve_executable_path(args.cmake_path)
+    build_dir = args.build_dir
+
+    if args.update or args.build:
+        for config in configs:
+            os.makedirs(_get_build_config_dir(build_dir, config), exist_ok=True)
```
edgchen1, line 581: also done in _generate_build_tree, probably not needed here

tools/build.py
```diff
@@ -0,0 +1,681 @@
...
+    parser.add_argument("--ios", action="store_true", help="build for ios")
+    parser.add_argument("--ios_sysroot", default="",
+                        help="Specify the name of the platform SDK to be used. e.g. iphoneos, iphonesimulator")
+    parser.add_argument("--ios_toolchain_file", default=f"{REPO_DIR}/cmake/ortext_ios.toolchain.cmake", type=Path,
+                        help="Path to ios toolchain file. Default is <repo>/cmake/ortext_ios.toolchain.cmake")
+    parser.add_argument("--xcode_code_signing_team_id", default="",
+                        help="The development team ID used for code signing in Xcode")
+    parser.add_argument("--xcode_code_signing_identity", default="",
+                        help="The development identity used for code signing in Xcode")
+    parser.add_argument("--use_xcode", action="store_true",
```
edgchen1, line 142: could we just use --cmake_generator instead of having an additional option for xcode?

tools/build.py
```diff
@@ -0,0 +1,681 @@
...
+            continue
+        elif args.ios:
+            _run_ios_tests(args, config, cwd)
+            continue
+
+        if ctest_path:
+            ctest_cmd = [str(ctest_path), "--build-config", config, "--verbose", "--timeout", "10800"]
+            _run_subprocess(ctest_cmd, cwd=cwd)
+
+        else:
```
edgchen1, line 490: why do we have both a ctest and non-ctest way to run tests?

tools/build.py
skottmckay: @guschmue are you folks likely to want to use build.py in a development enivronment to build ort-extensions? If not I will remove the wasm related stuff from here.

tools/build.py
skottmckay: this one runs first so removed the one in _generate_build_tree

tools/build.py
skottmckay: I'll leave Python_EXECUTABLE as we do have a `find_package(python3 ...)` usage. that means the same python binary used to run this script is guaranteed to be used for the build.

tools/build.py
skottmckay: To get code coverage from the tests we'd need to run manually. 

Maybe we should default to ctest (assume it's available in PATH if not explicitly provided) unless the user requests code coverage to be generated (windows only, execute test exes directly with vstest)

build.android
```diff
@@ -39,7 +39,7 @@ pushd ${target_dir}
 #       the great change of file system permission in Android 7
 cmake "$@" \
     -DCMAKE_TOOLCHAIN_FILE=${NDK_ROOT}/build/cmake/android.toolchain.cmake  \
-    -DANDROID_NDK=${NDK_ROOT}                                               \
+    -DANDROID_NDK_HOME=${NDK_ROOT}                                               \
```
wenbingl, line 42: I was confused by this doc, https://github.com/android/ndk-samples/wiki/Configure-NDK-Path#android--ndk--home. It says the ANDROID_NDK_HOME is deprecated while it also says it's useful in CI pipeline.

CMakeLists.txt
```diff
@@ -459,16 +462,11 @@ endif()
...
 
 foreach(nf ${NO_USE_FILES})
-  file(TO_NATIVE_PATH ${nf} nf_native)
-
-  if(CMAKE_SYSTEM_NAME MATCHES "Windows")
-    execute_process(COMMAND cmd /c "del ${nf_native}")
-  else()
-    execute_process(COMMAND bash -c "rm ${nf_native}")
-  endif()
+    file(TO_NATIVE_PATH ${nf} nf_native)
```
wenbingl, line 468: Looks better, is to_native_path still needed then?

cmake/ext_ortlib.cmake
```diff
@@ -2,7 +2,9 @@ if(_ONNXRUNTIME_EMBEDDED)
   set(ONNXRUNTIME_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/../include/onnxruntime/core/session)
   set(ONNXRUNTIME_LIB_DIR "")
 else()
-  set(ONNXRUNTIME_VER "1.10.0" CACHE STRING "ONNX Runtime version")
+  if (NOT ONNXRUNTIME_VER)
```
wenbingl, line 5: Is that what a cache variable doing. If it is not set, it will create one with the specified value?

build.android
skottmckay: I don't think we need this at all. android.toolchain.cmake will set it correctly, and I didn't find anywhere in our files that it's used.

cmake/ext_ortlib.cmake
skottmckay: Ah cool. Didn't realize it did that. Will remove this change.

tools/build.py
edgchen1: I see. maybe a separate option for code coverage would be good. the default is to use "ctest" now. one has to specify (oddly) `--ctest_path=""` to not use ctest.

tools/build.py
```diff
@@ -0,0 +1,645 @@
...
+
+    if args.ios:
+        required_args = [
+            args.ios_sysroot,
+            args.apple_deploy_target,
+        ]
+
+        arg_names = [
+            "--ios_sysroot          " + "<the location or name of the macOS platform SDK>",
+            "--apple_deploy_target  " + "<the minimum version of the target platform>",
```
edgchen1, line 354: I think we'd also need the target arch too. can add it later when enabling iOS builds here

tools/build.py
skottmckay: Updated to use ctest unless code coverage (windows only) is requested. as the user has cmake it seems reasonable that ctest is available, and is preferred as it avoids manually specifying tests to run.

CMakeLists.txt
skottmckay: Without TO_NATIVE_PATH the path always has `/` in it. That works as Windows allows both `\\` and `/` as a path separator, so strictly speaking it isn't necessary.

tools/build.py
skottmckay: Simplified the wasm setup to be the minimal required for now. emsdk needs to be manually cloned.

tools/build.py
skottmckay: Pruned more stuff.

tools/build.py
```diff
@@ -0,0 +1,653 @@
...
+    parser.add_argument("--emsdk_version", default="3.1.26", help="Specify version of emsdk")
+
+    # x86 args
+    parser.add_argument("--x86", action="store_true",
+                        help="[cross-compiling] Create Windows x86 makefiles. Requires --update and no existing cache "
+                             "CMake setup. Delete CMakeCache.txt if needed")
+
+    # Arguments needed by CI
+    parser.add_argument("--cmake_path", default="cmake", type=Path, help="Path to the CMake program.")
+    parser.add_argument("--ctest_path", default="ctest",
```
edgchen1, line 164: nit: maybe just make it consistent?
`parser.add_argument("--ctest_path", default="ctest", type=Path, help="Path to the CTest program.")`

edgchen1 (approved):

edgchen1 (approved):

PR #319, opened 2022-11-28
using the cmake bundled in visual studio

wenbingl:


setup.py
```diff
@@ -22,9 +22,10 @@
 
 def load_msvcvar():
     if os.environ.get('vcvars'):
+        os.environ['__vcvars'] = '"{}"'.format(os.environ['vcvars'])
         stdout, _ = subprocess.Popen([
-            'cmd', '/q', '/c', '(%vcvars% & set)'],
-            stdout=subprocess.PIPE, shell=True, universal_newlines=True).communicate()
+            'cmd', '/q', '/c', 'call %__vcvars% && set'],
+            stdout=subprocess.PIPE, shell=False, universal_newlines=True).communicate()
```
skottmckay, line 28: Does it need to create a new environment variable, or would just setting `shell=False` be enough?

.az/mshost.yaml
```diff
@@ -317,6 +317,7 @@ jobs:
 
       - script: |
           call activate pyenv
+          $env:vcvars="$env:VCINSTALLDIR\Auxiliary\Build\vcvars64.bat"
```
skottmckay, line 320: Might be good to add a comment saying we do this so cmake comes from VS

skottmckay (approved):

setup.py
wenbingl: I think shell=True is to create another cmd to execute the command line 'cmd /q /c ...', which looks redundant.

.az/mshost.yaml
wenbingl: yes. I will put some comments here after the tryout of getting where VS is installed

.az/mshost.yaml
```diff
@@ -317,7 +317,16 @@ jobs:
 
       - script: |
           call activate pyenv
-          set vcvars=%VCINSTALLDIR%Auxiliary\Build\vcvars64.bat
+          @echo find where Visual Studio was installed and use the bundled cmake
```
snnn, line 320: You can try this code : https://github.com/microsoft/onnxruntime-inference-examples/blob/main/.pipelines/OneBranch.Official.yml#L114

wenbingl:
@skottmckay , @snnn who can help with an approval?

xadupre (approved):

wenbingl:
@xadupre , Thanks!

PR #320, opened 2022-11-30
Add infrastructure to build iOS static framework and xcframework

edgchen1:
Initial changes for adding an iOS package.

Add infrastructure to build iOS static framework and xcframework.
Add xcframework build to CI pipeline.


CMakeLists.txt
```diff
@@ -45,6 +45,7 @@ option(OCOS_ENABLE_SELECTED_OPLIST "Enable including the selected_ops tool file"
 option(OCOS_BUILD_PYTHON "Enable building the Python package" OFF)
 option(OCOS_BUILD_JAVA "Enable building the Java package" OFF)
 option(OCOS_BUILD_ANDROID "Enable building the Android package" OFF)
+option(OCOS_BUILD_APPLE_FRAMEWORK "Enable building of the MacOS/iOS framework" OFF)
```
wenbingl, line 48: Is it 'Apple framework' or 'iOS framework'?

cmake/externals/opencv/ortextensions_readme.md
```diff
@@ -0,0 +1,8 @@
+This directory contains a subset of the files in the OpenCV repo at the 4.5.4 tag. In particular:
+- https://github.com/opencv/opencv/tree/4.5.4/platforms/ios/cmake/Toolchains
+- https://github.com/opencv/opencv/tree/4.5.4/platforms/ios/cmake/Modules/Platform
+- https://github.com/opencv/opencv/blob/4.5.4/COPYRIGHT
+- https://github.com/opencv/opencv/blob/4.5.4/LICENSE
+
+We duplicate them in this repo in order to use the OpenCV-specific CMake toolchain files for iOS builds.
```
wenbingl, line 7: Are there any changes on these files?

tools/ios/build_xcframework.py
```diff
@@ -0,0 +1,185 @@
...
+        f"-DCMAKE_XCODE_ATTRIBUTE_CODE_SIGNING_ALLOWED=NO",
+        "-DOCOS_BUILD_APPLE_FRAMEWORK=ON",
+        # our version of sentencepiece doesn't support iOS build
+        #   CMake Error at out/ios/RelWithDebInfo/_deps/spm-src/src/CMakeLists.txt:288 (install):
+        #     install TARGETS given no BUNDLE DESTINATION for MACOSX_BUNDLE executable
+        #     target "spm_encode".
+        "-DOCOS_ENABLE_SPM_TOKENIZER=OFF",
+        # OpenCV CMake config
+        # tell OpenCV to build zlib so we can link to the static library
+        "-DBUILD_ZLIB=ON",
```
wenbingl, line 70: should we turn on this flag in externals/opencv.cmake?

tools/ios/build_xcframework.py
```diff
@@ -0,0 +1,185 @@
...
+        # use OpenCV's CMake toolchain files
+        f"-DCMAKE_TOOLCHAIN_FILE={_get_opencv_toolchain_file(platform, opencv_dir)}",
+        "-DCPU_BASELINE=DETECT",
+        f"-DIOS_ARCH={arch}",
```
wenbingl, line 74: does DETECT mode work on cross-compiling iOS native framework, since I thought DETECT mode is to detect the CPU on the compiling machine?

cmake/externals/opencv/ortextensions_readme.md
edgchen1: nope, I copied them from `<binary dir>/_deps/opencv-src`

tools/ios/build_xcframework.py
edgchen1: not sure if this is something we generally want to enable. it's just for the iOS static framework build now.

tools/ios/build_xcframework.py
edgchen1: the opencv build script uses `CPU_BASELINE=DETECT` when cross-compiling:
https://github.com/opencv/opencv/blob/4223495e6cd67011f86b8ecd9be1fa105018f3b1/platforms/ios/build_framework.py#L253

as to how/whether it works, I don't understand their build system well enough to say...

CMakeLists.txt
edgchen1: I think at this level in CMakeLists.txt it is flexible enough to be either for iOS or MacOS. so went with the more generic "APPLE"

tools/ios/build_xcframework.py
wenbingl: I mean something like 
if (iOS)
set(BUILD_ZLIB ON CACHE INTERNAL "")
endif()
and CPU_BASELINE and IOS_ARCH problem should be same.

Right now, none of internal flag was exposed for build-script and all of them are managed by CMakefile. it would be better to keep the same pattern.

CMakeLists.txt
wenbingl: I don't know too much macOS framework, was it very similar to iOS's one on building script? If not, the iOS framework seems more specific.

edgchen1: `tools/ios/build_xcframework.py` is iOS specific. but here it should work for MacOS as well, at the moment.

tools/ios/build_xcframework.py
edgchen1: moved some of them into opencv.cmake. but others are required by the toolchain file so setting it in opencv.cmake would be too late.

.az/mshost.yaml
```diff
@@ -98,7 +98,7 @@ jobs:
 
   - job: MacOSX
     pool:
-      vmImage: 'macOS-latest'
+      vmImage: 'macOS-12'
```
wenbingl, line 101: did this framework have to be built under macOS-12?

wenbingl (approved):

.az/mshost.yaml
edgchen1: I think the framework itself doesn't, but the OrtExtensionsUsage project didn't build successfully with an older version of Xcode. Maybe it's because I created it with Xcode 14.1. So I just updated the CI build to use 14.1 too (in xcode-build-and-test-steps.yml).

PR #323, opened 2022-12-06
Initial CLIP tokenizer implementation

sayanshaw24:
### For reference: [HuggingFace Implementation](https://github.com/huggingface/transformers/blob/v4.24.0/src/transformers/models/clip/tokenization_clip.py)

Validation
- [x] Added sample test cases under `test_cliptok.py` and cross-validated with HuggingFace pretrained CLIPTokenizer.
- [x] Extensions building correctly using `python setup.py develop`

Note: This is not a final implementation. There are still edge cases that the tokenizer may not be able to handle at the moment.


operators/tokenizer/clip_tokenizer.cc
```diff
@@ -0,0 +1,627 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
wenbingl, line 1: create a bpetokenizer.hpp to put all the common function and classes into it, then remove the code from gpt2_* and clip_* files.

operators/tokenizer/clip_tokenizer.hpp
```diff
@@ -0,0 +1,27 @@
...
+struct KernelClipBpeTokenizer : BaseKernel {
+  KernelClipBpeTokenizer(const OrtApi& api, const OrtKernelInfo* info);
+  void Compute(OrtKernelContext* context);
+
+ private:
+  std::vector<int64_t> Tokenize(ustring& input, int64_t max_length);
+
+  int64_t padding_length_;
+  std::list<int> byte_list_;
+  std::shared_ptr<VocabData> bbpe_tokenizer_;
```
wenbingl, line 17: why does it have to be shared_prt, instead of unique_ptr

operators/tokenizer/clip_tokenizer.cc
```diff
@@ -0,0 +1,226 @@
...
+}
+
+std::vector<int64_t> KernelClipBpeTokenizer::Tokenize(ustring& input, int64_t max_length) {
+  std::vector<int64_t> res;
+
+  if (IsEmptyUstring(input)) {
+    return res;
+  }
+  // Add <|startoftext|> token to result
+  res.push_back(49406);
```
wenbingl, line 100: Can this number be queried from Vocab object?

operators/tokenizer/clip_tokenizer.cc
```diff
@@ -0,0 +1,226 @@
...
+        if (static_cast<int64_t>(res.size()) >= max_length) {
+          break;
+        }
+
+        res.push_back(p);
+      }
+    }
+  }
+  // Add <|endoftext|> token to result
+  res.push_back(49407);
```
wenbingl, line 156: same as above

operators/tokenizer/clip_tokenizer.cc
```diff
@@ -0,0 +1,226 @@
...
+  std::vector<int64_t> res;
+
+  if (IsEmptyUstring(input)) {
+    return res;
+  }
+  // Add <|startoftext|> token to result
+  res.push_back(49406);
+
+  // Convert to lowercase
+  std::transform(input.begin(), input.end(), input.begin(), [](char c) { return static_cast<char>(ToLower(c)); });
```
wenbingl, line 103: the character type of ustring should be char32_t, instead of 'char'. The 'static_cast' will introduce some Unicode issues here.

operators/tokenizer/clip_tokenizer.cc
sayanshaw24: Yes, changed it.

operators/tokenizer/clip_tokenizer.hpp
sayanshaw24: I believe it is because we are using the same implementation of VocabData for gpt2 and clip tokenizers, and the former requires multiple pointers to it. I tried it still however and changed our implementation accordingly but ran into memory errors while building.

wenbingl (approved):
Thanks!

PR #324, opened 2022-12-06
Fix ci pipelines failure caused by the latest transformer package

wenbingl:
[transformer 4.25 broken on pytorch 1.9/1.10


sayanshaw24 (approved):

edgchen1 (approved):
thanks!

PR #327, opened 2022-12-12
Add iOS packaging pipeline.

edgchen1:
Add new build definition.
Add wrapper script for tools/ios/build_xcframework.py.

iOS packaging pipeline: https://aiinfra.visualstudio.com/Lotus/_build?definitionId=1206


.az/ios_packaging.yml
```diff
@@ -0,0 +1,107 @@
...
+
+      set_var() {
+        local VAR_NAME=${1:?}
+        local VAR_VALUE=${2:?}
+        echo "##vso[task.setvariable variable=${VAR_NAME}]${VAR_VALUE}"
+        echo "${VAR_NAME}: ${VAR_VALUE}"
+      }
+
+      set_var "ORT_EXTENSIONS_POD_VERSION" "${VERSION}"
+    displayName: "Set ORT_EXTENSIONS_POD_VERSION variable"
```
YUNQIUGUO, line 58: not a comment, question:

noticed this is added as an extra step compared to the `CI pipline - IosPackage job`, do we have to do similar set environment variable thing for Android packaging pipline?

YUNQIUGUO: ah sorry, I think I see. In the CI pipline, it is using some dev version pods, here we just formalize it.

edgchen1: yeah, this is just adjusting the version based on the IsReleaseBuild parameter. maybe I can put it into a template so we can reuse it.
would be good to have something similar for the Android pipeline.

.az/ios_packaging.yml
```diff
@@ -0,0 +1,91 @@
+# packaging pipeline for iOS CocoaPods package
```
wenbingl, line 1: let's move this file to .pipeline folder. sooner or later, all pipelines in .az folder will be merged into .pipeline folder which has a better name.

tools/gen_selectedops.py
```diff
@@ -1,80 +1,99 @@
...
-    'StringMapping': 'OCOS_ENABLE_TF_STRING',
-    'VectorToString': 'OCOS_ENABLE_TF_STRING',
-    'MaskedFill': 'OCOS_ENABLE_TF_STRING',
-    'BertTokenizer': 'OCOS_ENABLE_BERT_TOKENIZER',
-    'BasicTokenizer': 'OCOS_ENABLE_BERT_TOKENIZER',
-    'BertTokenizerDecoder': 'OCOS_ENABLE_BERT_TOKENIZER',
-    'SentencepieceTokenizer': 'OCOS_ENABLE_SPM_TOKENIZER',
-    'ImageDecode': 'OCOS_ENABLE_VISION',
-    'ImageEncode': 'OCOS_ENABLE_VISION',
+    "BlingFireSentenceBreaker": "OCOS_ENABLE_BLINGFIRE",
```
edgchen1, line 7: note: ran black for formatting and it made quite a few cosmetic changes here

.pipelines/ios_packaging.yml
```diff
@@ -0,0 +1,92 @@
...
+        pod install
+    displayName: "Install pods for OrtExtensionsUsage"
+    workingDirectory: $(Build.SourcesDirectory)/test/ios/OrtExtensionsUsage
+
+  - template: templates/xcode-build-and-test-step.yml
+    parameters:
+      xcWorkspacePath: '$(Build.SourcesDirectory)/test/ios/OrtExtensionsUsage/OrtExtensionsUsage.xcworkspace'
+      scheme: 'OrtExtensionsUsage'
+
+  # TODO run test on app center
```
edgchen1, line 70: note: UI test works locally but running into an app center error. might defer this TODO to another PR

.pipelines/templates/with-ios-simulator-steps.yml
```diff
@@ -16,7 +16,7 @@ steps:
 - bash: |
     set -e -x
 
-    if [[ -v ORT_EXTENSIONS_BUILD_SIMULATOR_ID ]]; then
+    if [[ -n "${ORT_EXTENSIONS_BUILD_SIMULATOR_ID}" ]]; then
```
wenbingl, line 19: try [[ -n ${ORT_EXTENSIONS_BUILD_SIMULATOR_ID-} ]] 
if you want to test if it existing and not empty, then do something.

edgchen1: thanks for the tip.
`[[ -n "${ORT_EXTENSIONS_BUILD_SIMULATOR_ID}" ]]` does work in the CI build. I'm not very familiar with bash though, is there a reason to prefer `[[ -n ${ORT_EXTENSIONS_BUILD_SIMULATOR_ID-} ]]`?

wenbingl: it is more robust since it can also run with 'bash -u'.

test/ios/OrtExtensionsUsage/OrtExtensionsUsage.xcodeproj/project.pbxproj
```diff
@@ -15,9 +16,18 @@
 		2ECE960B293A77FD00039409 /* OrtClient.mm in Sources */ = {isa = PBXBuildFile; fileRef = 2ECE9608293A77FD00039409 /* OrtClient.mm */; };
 		2ECE960C293A77FD00039409 /* decode_image.onnx in Resources */ = {isa = PBXBuildFile; fileRef = 2ECE9609293A77FD00039409 /* decode_image.onnx */; };
 		2ECE960E293A784E00039409 /* r32_g64_b128_32x32.png in Resources */ = {isa = PBXBuildFile; fileRef = 2ECE960D293A784E00039409 /* r32_g64_b128_32x32.png */; };
+		70D1E184C8218474862CBDDE /* Pods_OrtExtensionsUsageTests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = C8950873A72FE487AF82F47D /* Pods_OrtExtensionsUsageTests.framework */; };
```
edgchen1, line 19: revert this

tools/gen_selectedops.py
```diff
@@ -1,80 +1,99 @@
...
+    "BertTokenizer": "OCOS_ENABLE_BERT_TOKENIZER",
+    "BasicTokenizer": "OCOS_ENABLE_BERT_TOKENIZER",
+    "BertTokenizerDecoder": "OCOS_ENABLE_BERT_TOKENIZER",
+    "SentencepieceTokenizer": "OCOS_ENABLE_SPM_TOKENIZER",
+    "ImageDecode": "OCOS_ENABLE_VISION",
+    "ImageEncode": "OCOS_ENABLE_VISION",
 }
 
 
-def gen_cmake_oplist(opconfig_file, oplist_cmake_file='_selectedoplist.cmake'):
```
wenbingl, line 35: This file name was fixed to be consistent with cmake option OCOS_ENABLE_SELECTED_OPLIST which will include the file name. It's good to use this option to build the package, but I didn't find how you specify the file name on CMake configuration.

tools/gen_selectedops.py
```diff
@@ -1,80 +1,99 @@
...
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Generate the cmake/_selectedoplist.cmake file from the selected op config file.",
+    )
 
-    print('[onnxruntime-extensions] The cmake tool file has been generated successfully.')
+    parser.add_argument(
+        "selected_op_config_file",
```
wenbingl, line 79: It's kind of being intentional to always to have the same file to cmake. Given there is opconfig file, we don't add one more file to confuse the user.

.pipelines/templates/with-ios-simulator-steps.yml
snnn: Add a letter, for example "x", in front of the string. Then test if the result equals to "x".

tools/gen_selectedops.py
edgchen1: the path actually passed to this is still the same. the default argument value was always being overridden before so I just removed it
https://github.com/microsoft/onnxruntime-extensions/blob/56ef48073c4e58c44c69f27e9b5fb7c8cf29b7f4/tools/gen_selectedops.py#L93-L99

tools/gen_selectedops.py
edgchen1: the script still takes one op config file argument, as it did before. now there is more usage information

.pipelines/templates/with-ios-simulator-steps.yml
edgchen1: > Add a letter, for example "x", in front of the string. Then test if the result equals to "x".

I see that that should work too, but why is it better?

skottmckay:
> 

DecodeImage,EncodeImage

---
Refers to: tools/ios/package_ops.config:2 in aaabd23. [](commit_id = aaabd237eb1701c015aaf8d4ceb9e077b915855f, deletion_comment = False)

tools/gen_selectedops.py
```diff
@@ -1,80 +1,99 @@
...
+    "SegmentExtraction": "OCOS_ENABLE_MATH",
+    "StringMapping": "OCOS_ENABLE_TF_STRING",
+    "VectorToString": "OCOS_ENABLE_TF_STRING",
+    "MaskedFill": "OCOS_ENABLE_TF_STRING",
+    "BertTokenizer": "OCOS_ENABLE_BERT_TOKENIZER",
+    "BasicTokenizer": "OCOS_ENABLE_BERT_TOKENIZER",
+    "BertTokenizerDecoder": "OCOS_ENABLE_BERT_TOKENIZER",
+    "SentencepieceTokenizer": "OCOS_ENABLE_SPM_TOKENIZER",
+    "ImageDecode": "OCOS_ENABLE_VISION",
+    "ImageEncode": "OCOS_ENABLE_VISION",
```
skottmckay, line 33: Sorry - I created the mismatch with names at some point. Should be DecodeImage and EncodeImage to match the actual operator names. I can fix in a separate PR if you like.

tools/gen_selectedops.py
skottmckay: Given the path is hardcoded in CMakeLists.txt is there any reason it should even be configurable?

tools/ios/build_xcframework.py
```diff
@@ -191,6 +204,15 @@ def parse_args():
...
     )
+    parser.add_argument(
+        "--cmake-extra-define",
+        "-D",
+        action="append",
+        default=[],
+        metavar="CMAKE_EXTRA_DEFINE",
+        dest="cmake_extra_defines",
+        help="Extra definition to pass to CMake (with the CMake -D option) during build system generation.",
+    )
```
skottmckay, line 215: nit: Can this be the same as the ORT build.py for consistency? That allows multiple to be specified at once (e.g. `--cmake_extra_defines A=0 B=1` as well, and is called `cmake_extra_defines`. 

```
    parser.add_argument("--cmake_extra_defines", nargs="+", action="append", help="...",
```

tools/gen_selectedops.py
edgchen1: got it. np, will fix it here

tools/gen_selectedops.py
edgchen1: good point. hardcoded it instead of using a function parameter because it's always the same now. hope that is clearer.

tools/ios/build_xcframework.py
edgchen1: sure, updated it to be more consistent. using the name "--cmake-extra-defines" to be consistent with other options here

.pipelines/templates/with-ios-simulator-steps.yml
snnn: https://tldp.org/LDP/abs/html/comparison-ops.html#AEN3669 Maybe it is not a concern anymore.

skottmckay (dismissed):
:shipit:

tools/gen_selectedops.py
```diff
@@ -1,80 +1,103 @@
...
+    "StringUpper": "OCOS_ENABLE_TF_STRING",
+    "SegmentExtraction": "OCOS_ENABLE_MATH",
+    "StringMapping": "OCOS_ENABLE_TF_STRING",
+    "VectorToString": "OCOS_ENABLE_TF_STRING",
+    "MaskedFill": "OCOS_ENABLE_TF_STRING",
+    "BertTokenizer": "OCOS_ENABLE_BERT_TOKENIZER",
+    "BasicTokenizer": "OCOS_ENABLE_BERT_TOKENIZER",
+    "BertTokenizerDecoder": "OCOS_ENABLE_BERT_TOKENIZER",
+    "SentencepieceTokenizer": "OCOS_ENABLE_SPM_TOKENIZER",
+    "DecodeImage": "OCOS_ENABLE_VISION",
```
wenbingl, line 33: it may not directly relate to this PR, but I want to have discussion about the Op name. In most case, there *Tokenizer, or *decoder. so it would be consistent to name them as ImageDecoder and ImageEncoder?

tools/gen_selectedops.py
wenbingl: black is pretty controversial for me since it sacrifices some flexibility, and its goal is also questionable, check this article
https://luminousmen.com/post/my-unpopular-opinion-about-black-code-formatter

So, currently it wasn't enabled for this project by default, and was optional.

tools/gen_selectedops.py
```diff
@@ -1,80 +1,103 @@
...
+    "EncodeImage": "OCOS_ENABLE_VISION",
 }
 
+SCRIPT_DIR = pathlib.Path(__file__).parent.resolve()
+GENERATED_CMAKE_CONFIG_FILE = SCRIPT_DIR.parent / "cmake/_selectedoplist.cmake"
```
wenbingl, line 38: nit: "cmake" / "..."?

tools/gen_selectedops.py
wenbingl: it looks like we are all on the same page that there is only one file of cmake op option with a fixed name.

tools/gen_selectedops.py
edgchen1: sure, I'm happy to go with whatever formatting convention we want to use for this project. but I didn't know of any so just copied ORT's. let me know if there is another one to follow.

skottmckay: Personally I was happy with flake8 enforcing most of PEP8 in ORT. There's a config for that in the ORT repo.

ORT started using black a few months ago, but I wasn't involved in the discussion (assuming there was one) about adopting that as the standard. I also don't like some of the black formatting as I think it diminishes readability in many cases (although #fmt off/#fmt on can turn it off for sections).

tools/gen_selectedops.py
skottmckay: I named them this way to avoid a clash with the opencv ops as there's an `ImageDecoder` there. 

I would like to separate the operators that are 1:1 with opencv functions and behavior from operators that may or may not use opencv as a dependency. e.g. we may directly use libpng and libjpeg-turbo for the pre/post processing to reduce binary size, but currently are using opencv due to the timeline. 

The vision operators are aimed at providing the required decode/encode for mobile scenarios rather than being a 1:1 with opencv. e.g. I need to support RGB input for web (basically a no-op decode), and also want to make the output layout configurable so it could be RGB or BGR instead of always being the (slightly odd) opencv BGR layout. 

If we rename the opencv operators I'm fine with changing general purpose image operators to use 'Image' as the prefix. Not sure what a good naming scheme would be though. Use a different domain for the opencv ops and the function name (but function name is all lowercase so that's slightly inconsistent with operator names in general)?

.pipelines/ios_packaging.yml
```diff
@@ -0,0 +1,128 @@
...
+      signingOption: 'manual'
+      signingIdentity: '$(APPLE_CERTIFICATE_SIGNING_IDENTITY)'
+      provisioningProfileName: 'iOS Team Provisioning Profile'
+      args: '-derivedDataPath $(Build.BinariesDirectory)/appcenter_test_derived_data'
+    displayName: "Build appcenter test"
+
+  - script: |
+      appcenter test run xcuitest \
+        --app "AI-Frameworks/ORT-Ext-Mobile-iOS-Testapp" \
+        --devices "AI-Frameworks/apple-test-device-set" \
```
YUNQIUGUO, line 101: looks good to me, nit:  `--devices "AI-Frameworks/apple-test-device-set"` shall we also create a variable for apple test sets and use $(app_center_apple_test_devices) here?

YUNQIUGUO (approved):

skottmckay (approved):
:shipit:

.pipelines/ios_packaging.yml
edgchen1: ok, we can refine it. I think either a device set or a variable specifying device(s) would work

PR #328, opened 2022-12-13
Include requirements.txt from requirements-dev.txt.

edgchen1:
Include requirements.txt from requirements-dev.txt so that pip has context to avoid installing incompatible dependencies.

Fix the CI build.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,3 +1,4 @@
+# include requirements.txt so pip has context to avoid installing incompatible dependencies
 -r requirements.txt
 pytest
 transformers     >= 4.9.2,<=4.24.0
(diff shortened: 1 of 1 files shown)


requirements-dev.txt
```diff
@@ -1,4 +1,5 @@
+# include requirements.txt so pip has context to avoid installing incompatible dependencies
+-r requirements.txt
 pytest
 transformers     >= 4.9.2,<=4.24.0
 tensorflow_text  >=2.5.0
-protobuf         >=3.12.2,<=3.20.1
```
wenbingl, line 4: this restriction is introduced by some version of ONNX, but I am not sure if it is fixed in ONNX now.

wenbingl (approved):
Since all CI pipelines succeeded, let's remove this line.

requirements-dev.txt
edgchen1: I think later versions of ONNX specify the protobuf dependency more correctly

PR #331, opened 2022-12-15
Updates to PPP model update scripts

skottmckay:
- Fix weights for YCbCr to BGR
  - this was causing the slight color diff in the test image I've been using
- Fix Resize when input is HWC or CHW by converting to NHWC/NCHW so bilinear interpolation is inferred. 
  - trilinear is inferred otherwise.
    - Not aware of a use case where we'd want to use trilinear - if one exists the temporary addition of the batch dim would need to be made explicit.
- Allow Resize to use antialias if ONNX opset is 18 or higher
- Skip adding unnecessary Mul if FloatToImageBytes multiplier is 1.0.
- Make Debug step available at the top level

Validated opset 18 Resize vs PT for super resolution and the results are now a really good match when antialiasing is used.


tools/pre_post_processing/steps/vision.py
```diff
@@ -118,6 +118,11 @@ def _create_graph_for_step(self, graph: onnx.GraphProto):
                                 [0.500, -0.587 / 1.402, -0.114 / 1.402]],
                                dtype=np.float32)  # fmt: skip
 
+        # https://github.com/python-pillow/Pillow/blob/63f64de8b5cc6d065f19cc73c865633ab4940a52/src/libImaging/ConvertYCbCr.c#L17
+        # rgb_weights = np.array([[0.299, 0.587, 0.114],
```
edgchen1, line 122: the commented out values seem to be close to the values above. e.g., `-0.299 / 1.772` vs `-0.16874`. why are they listed here? maybe add some more context in the comment?

tools/pre_post_processing/steps/vision.py
```diff
@@ -324,25 +339,46 @@ def _create_graph_for_step(self, graph: onnx.GraphProto):
...
 
-        # TODO: Make this configurable. Matching PIL resize for now
+        # TODO: Make this configurable. Matching PIL resize for now.
         resize_attributes = 'mode = "linear", nearest_mode = "floor"'
+        if PRE_POST_PROCESSING_ONNX_OPSET >= 18:
+            # Resize matches PIL better if antialiasing is used, but that isn't available until ONNX opset 18
+            # Allow this to be used with older opsets as well.
+            resize_attributes += ', antialias = 1'
+
+        # resize_attributes = 'mode = "linear", nearest_mode = "floor"'
```
edgchen1, line 349: remove commented code?

tools/pre_post_processing/steps/vision.py
```diff
@@ -324,25 +339,46 @@ def _create_graph_for_step(self, graph: onnx.GraphProto):
...
+
+        # if input is HWC or CHW we need to add a 'N' dim so Resize infers works correctly. 3D input is inferred to
```
edgchen1, line 351: "so Resize infers correctly"?

tools/add_pre_post_processing_to_model.py
```diff
@@ -130,27 +130,29 @@ def superresolution(model_file: Path, output_file: Path):
...
-                    IoMapEntry("Yout_to_bytes", 0, 0),  # uint8 Y' with shape {h, w}
-                    IoMapEntry("Resized_Cb", 0, 1),  # uint8 Cb'
-                    IoMapEntry("Resized_Cr", 0, 2),  # uint8 Cr'
+                    IoMapEntry("Y1_uint8", 0, 0),  # uint8 Y' with shape {h, w}
+                    IoMapEntry("Cb1_uint8", 0, 1),
+                    IoMapEntry("Cr1_uint8", 0, 2),
                 ],
             ),
-            ConvertBGRToImage(image_format="jpg"),  # jpg or png are supported
+            ConvertBGRToImage(image_format="png"),  # jpg or png are supported
```
edgchen1, line 155: do we need to update the test to handle a png instead?

tools/pre_post_processing/steps/vision.py
skottmckay: Was temporary whilst trying to get to the bottom of the diffs. Removed.

tools/pre_post_processing/steps/vision.py
skottmckay: Reworded to clarify.

tools/add_pre_post_processing_to_model.py
skottmckay: Yes.

edgchen1 (dismissed):

edgchen1 (approved):

PR #332, opened 2022-12-15
Add initial android packaging pipeline

YUNQIUGUO:
Add initial android packaging pipeline

TODO: 

1. Set version number
2. Move app center test to this pipeline
3. Ort android aar package published to maven also contains pom, etc. file as well, here we only have aar/javadoc/sources for the produced artifacts.


.pipelines/android_packaging.yml
```diff
@@ -0,0 +1,78 @@
...
+          set -e -x
+
+          AAR_STAGING_DIR="$(Build.sourcesdirectory)/out/android/Release/java/build/android/outputs/aar"
+          LIBS_STAGING_DIR="$(Build.sourcesdirectory)/out/android/Release/java/build/android/libs"
+          ARTIFACTS_STAGING_DIR="$(Build.ArtifactStagingDirectory)"
+
+          PACKAGE_NAME="onnxruntime-extensions"
+          cp ${LIBS_STAGING_DIR}/${PACKAGE_NAME}-0.5.0-javadoc.jar  ${ARTIFACTS_STAGING_DIR}
+          cp ${LIBS_STAGING_DIR}/${PACKAGE_NAME}-0.5.0-sources.jar  ${ARTIFACTS_STAGING_DIR}
+          cp ${AAR_STAGING_DIR}/${PACKAGE_NAME}-release.aar         ${ARTIFACTS_STAGING_DIR}
```
YUNQIUGUO, line 72: not sure how we can modify the aar package name here, wanted to match "onnxruntime-extensions-android" or "onnxruntime-android-ext".

seems like it's named here:
https://github.com/microsoft/onnxruntime-extensions/blob/main/java/build-android.gradle#L123

.pipelines/android_packaging.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - script: |
+          set -e -x
+
+          AAR_STAGING_DIR="$(Build.sourcesdirectory)/out/android/Release/java/build/android/outputs/aar"
+          LIBS_STAGING_DIR="$(Build.sourcesdirectory)/out/android/Release/java/build/android/libs"
+          ARTIFACTS_STAGING_DIR="$(Build.ArtifactStagingDirectory)"
+
+          PACKAGE_NAME="onnxruntime-extensions"
+          cp ${LIBS_STAGING_DIR}/${PACKAGE_NAME}-0.5.0-javadoc.jar  ${ARTIFACTS_STAGING_DIR}
```
YUNQIUGUO, line 70: hardcoded the version here now.

.pipelines/android_packaging.yml
```diff
@@ -0,0 +1,78 @@
...
+        workingDirectory: $(OnnxruntimeDirectory)
+
+      - task: JavaToolInstaller@0
+        displayName: Use jdk 11
+        inputs:
+          versionSpec: "11"
+          jdkArchitectureOption: "x64"
+          jdkSourceOption: "PreInstalled"
+
+      # TODO: set package version number here
```
edgchen1, line 47: currently we need to do a build first so CMake generates `.version.txt`. some ideas to handle that:
- check in version.txt in the repo and derive the cmake project version variables from it
- allow a version suffix to be specified for a package

.pipelines/android_packaging.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - script: |
+          bash ./tools/android.package.sh
```
edgchen1, line 50: should we specify the ops for the package?
e.g.:
https://github.com/microsoft/onnxruntime-extensions/blob/56ef48073c4e58c44c69f27e9b5fb7c8cf29b7f4/.pipelines/ios_packaging.yml#L30-L32
and then set OCOS_ENABLE_SELECTED_OPLIST=ON

however, I think the android test uses ops which aren't currently in the ios package ops list

YUNQIUGUO: Yes, thanks for the info. I can add an android/package_ops.config file. The android test currently uses BertTokenizer op.

Do you suggest me to modify the android test or what's the requirement for the package ops list? is it fine to just include BertTokenizer?

edgchen1: I think we should include the EncodeImage and DecodeImage ops in the config. it's probably fine to also include the BertTokenizer op for now if that is easier.

.pipelines/android_packaging.yml
```diff
@@ -0,0 +1,113 @@
...
+
+          PROJECT_NAME="onnxruntime-extensions"
+          PACKAGE_NAME="onnxruntime-mobile-extensions"
+          VERSION_FILE="$(Build.SourcesDirectory)/version.txt"
+          PROJECT_VERSION="$(cat "${VERSION_FILE}")"
+          cp ${LIBS_STAGING_DIR}/${PROJECT_NAME}-${PROJECT_VERSION}-javadoc.jar  ${ARTIFACTS_STAGING_DIR}
+          cp ${LIBS_STAGING_DIR}/${PROJECT_NAME}-${PROJECT_VERSION}-sources.jar  ${ARTIFACTS_STAGING_DIR}
+          cp ${AAR_STAGING_DIR}/${PROJECT_NAME}-release.aar                      ${ARTIFACTS_STAGING_DIR}
+          cp ${POM_FILE_STAGING_DIR}/pom-default.xml                             ${ARTIFACTS_STAGING_DIR}
+          mv ${ARTIFACTS_STAGING_DIR}/${PROJECT_NAME}-${PROJECT_VERSION}-javadoc.jar  ${ARTIFACTS_STAGING_DIR}/${PACKAGE_NAME}-${ORT_EXTENSIONS_AAR_VERSION}-javadoc.jar
```
edgchen1, line 104: is it sufficient to rename the files with a new version number? do we need to update build-android.gradle's project.version too?
https://github.com/microsoft/onnxruntime-extensions/blob/81e7799c69044c745239202085eb0a98f102937b/java/build-android.gradle#L24

.pipelines/android_packaging.yml
```diff
@@ -0,0 +1,113 @@
...
+          mv ${ARTIFACTS_STAGING_DIR}/${PROJECT_NAME}-${PROJECT_VERSION}-sources.jar  ${ARTIFACTS_STAGING_DIR}/${PACKAGE_NAME}-${ORT_EXTENSIONS_AAR_VERSION}-sources.jar
```
edgchen1, line 105: nit: instead of `cp <dir1>/<file> <dir2>; mv <dir2>/<file> <dir2>/<renamed file>` could just do `cp <dir1>/<file> <dir2>/<renamed file>`

build.android
```diff
@@ -42,7 +42,8 @@ cmake "$@" \
     -DANDROID_NDK=${NDK_ROOT}                                               \
     -DANDROID_ABI=${abi_name}                                               \
     -DANDROID_PLATFORM=android-24                                           \
-    -DOCOS_BUILD_ANDROID=ON                                                 \
     -DANDROID_NDK_VERSION=${CURRENT_NDK_VERSION}                            \
-    ../../.. && cmake --build . --config Release --parallel
+    -DOCOS_BUILD_ANDROID=ON                                                 \
+    -DOCOS_ENABLE_SELECTED_OPLIST=ON                                        \
```
edgchen1, line 47: maybe we can keep this argument outside of the script so the script is more general.
instead, we can call `tools/android.package.sh -DOCOS_ENABLE_SELECTED_OPLIST=ON`. it should pass the arguments through to this cmake command.

tools/gen_selectedops.py
```diff
@@ -1,80 +1,136 @@
-import os
+#!/usr/bin/env python3
```
edgchen1, line 1: FYI, probably don't need to include the gen_selectedops.py changes in your branch. it should functionally be the same as before.

tools/android/package_ops.config
```diff
@@ -0,0 +1 @@
+com.microsoft.extensions;1;DecodeImage,EncodeImage,BertTokenizer
```
edgchen1, line 1: nit: newline at end

.pipelines/android_packaging.yml
```diff
@@ -0,0 +1,113 @@
...
+          appcenter test run espresso \
+          --app "AI-Frameworks/Ort-Ext-Mobile-Android-Testapp" \
+          --devices $(app_center_android_test_devices) \
+          --app-path ./app/build/outputs/apk/debug/app-debug.apk \
+          --test-series "master" \
+          --locale "en_US" \
+          --build-dir ./app/build/outputs/apk/androidTest/debug \
+          --token $(app_center_api_token)
+        displayName: Run extensions Android unit test using App Center
+        workingDirectory: $(Build.sourcesdirectory)
```
edgchen1, line 86: can just set workingDirectory directly to `$(Build.SourcesDirectory)/out/android/Release/java/androidtest/android` and skip the `cd`
also, note that the default value for the script task's workingDirectory is $(Build.SourcesDirectory)

.pipelines/android_packaging.yml
YUNQIUGUO: yeah it would be better if we can directly generate the aar file with dev/release version number, but I am not sure how to replace the ../version.txt here with the package version we generated in the pipeline (base and dev version + hash suffix)

edgchen1: one idea is to read a value from an environment variable if it is defined and otherwise use the contents of version.txt. something like
`project.version = System.getenv('ORT_EXTENSIONS_AAR_VERSION') ?: rootProject.file('../version.txt').text.trim()`

edgchen1 (approved):

PR #333, opened 2022-12-16
Add version.txt and derive CMake version variables from it.

edgchen1:
Add version.txt and derive CMake version variables from it. Previously .version.txt was generated by CMake.
This way, we don't need to run CMake before we can get the version from this file.


YUNQIUGUO (approved):
thanks for the change!

wenbingl (dismissed):
Rename it to version.txt (without dot) since it was checked in, instead of an intermediate file?

wenbingl (approved):

PR #334, opened 2022-12-16
fix the ORT API version in header files.

wenbingl:


snnn (approved):

includes/onnxruntime_customop.hpp
```diff
@@ -99,7 +99,7 @@ struct CustomOpApi {
 template <typename TOp, typename TKernel>
 struct CustomOpBase : OrtCustomOp {
   CustomOpBase() {
-    OrtCustomOp::version = ORT_API_VERSION;
+    OrtCustomOp::version = 10;  // The minimum ORT version supported
```
edgchen1, line 102: is this comment in this link inaccurate?
https://github.com/microsoft/onnxruntime/blob/main/include/onnxruntime/core/session/onnxruntime_c_api.h#L3665

PR #335, opened 2022-12-17
Patch sentencepiece to enable iOS build.

edgchen1:


cmake/externals/sentencepieceproject.cmake
```diff
@@ -8,14 +8,16 @@ FetchContent_GetProperties(spm)
...
-  if(onnxruntime_BUILD_WEBASSEMBLY)
-    set(SPM_PATCH "${PROJECT_SOURCE_DIR}/cmake/externals/sentencepieceproject.patch")
+  foreach(SPM_PATCH IN ITEMS
+          # use protobuf provided by ORT
+          "${PROJECT_SOURCE_DIR}/cmake/externals/sentencepieceproject.protobuf.patch"
+          # fix for iOS build
+          "${PROJECT_SOURCE_DIR}/cmake/externals/sentencepieceproject.ios.patch"
+          )
     message("-- sentencepiece: patching with ${SPM_PATCH}")
     execute_process(COMMAND git apply --ignore-space-change --ignore-whitespace ${SPM_PATCH} WORKING_DIRECTORY ${spm_SOURCE_DIR})
```
snnn, line 19: Please note that "git apply" won't work if neither the working directory nor its ancestor-directories contain ".git". For example, when doing a out of source build.

edgchen1: Noted. With the current setup it should work since we specify a git repo in FetchContent_Declare().

cmake/externals/sentencepieceproject.cmake
```diff
@@ -1,21 +1,22 @@
+# spm is abbreviation of sentencepiece to meet the path length limits on Windows
+
+set(spm_patches
+    # use protobuf provided by ORT
+    "${PROJECT_SOURCE_DIR}/cmake/externals/sentencepieceproject.protobuf.patch"
```
wenbingl, line 5: How do you create another patch package based on an existing one? To be simple, should we just create one since the changes for iOS can be enclosed by if (iOS)?

cmake/externals/sentencepieceproject.cmake
wenbingl: Yes, git apply is common on FetchContent_declare. And not sure if there is a case that git cannot be used with FetchContent_

cmake/externals/sentencepieceproject.ios.patch
```diff
@@ -0,0 +1,28 @@
...
+ list(APPEND SPM_INSTALLTARGETS
+   spm_encode spm_decode spm_normalize spm_train spm_export_vocab)
+ 
+-install(TARGETS ${SPM_INSTALLTARGETS}
+-  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+-  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+-  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
++if (CMAKE_SYSTEM_NAME STREQUAL "iOS")
++  install(TARGETS ${SPM_INSTALLTARGETS}
++    BUNDLE DESTINATION ${CMAKE_INSTALL_BINDIR}
```
wenbingl, line 15: is it also needed on ort-extensions + onnxruntime combined build? if not, _ONNXRUNTIME_EMBEDDED flag may also should be tested.

cmake/externals/sentencepieceproject.cmake
```diff
@@ -1,27 +1,33 @@
+# spm is abbreviation of sentencepiece to meet the path length limits on Windows
+
+set(spm_patches)
+
+if(_ONNXRUNTIME_EMBEDDED)
+  # use protobuf provided by ORT
+  list(APPEND spm_patches "${PROJECT_SOURCE_DIR}/cmake/externals/sentencepieceproject.protobuf.patch")
```
edgchen1, line 7: the original patch does assume the presence of ORT variables.
updating the condition to apply it from `onnxruntime_BUILD_WEBASSEMBLY` to  `_ONNXRUNTIME_EMBEDDED`

cmake/externals/sentencepieceproject.ios.patch
edgchen1: I think it would be needed for iOS builds in general. this applies to the sentencepiece project targets and, unless we  further specify otherwise, those should be present in either case

cmake/externals/sentencepieceproject.cmake
edgchen1: the patch is actually based on the original repo. but both patches can apply cleanly together.

actually, realized that the protobuf-related patch only applies for the embedded in ORT build, so keeping that separate

wenbingl: It's kind of special that you can apply two patches together. 
If I also want to do a patch for spm for Windows build, do I need to do patches locally first or directly make the change? If latter, how can I avoid the conflict. Overall, these multiple patches method add unnecessary complex for the process. And I don't see any benefits from the multiple patches. So, let's keep the process simple and straightforward.

wenbingl: It's ok to keep them be separated in this PR, I will have a PR to make the combined build and standalone build with same configuration. Then, they can be merged.

wenbingl (approved):

cmake/externals/sentencepieceproject.cmake
edgchen1: ok, we can merge them later. it doesn't hurt to always include this ios fix patch.

PR #337, opened 2022-12-19
Use `object` instead of `np.object` and `str` instead of `np.str`.

edgchen1:
Fix CI build.

### DIFF
--- a/test/test_pyops.py
+++ b/test/test_pyops.py
@@ -147,7 +147,7 @@ class TestPythonOp(unittest.TestCase):
             res = []
             for x in xs:
                 res.append(sep.join(x))
-            return np.array(res, dtype=np.object_)
+            return np.array(res, dtype=object)
 
     def test_python_operator(self):
         so = _ort.SessionOptions()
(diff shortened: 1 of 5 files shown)


test/test_pyops.py
```diff
@@ -147,7 +147,7 @@ def join(xs, **kwargs):
             res = []
             for x in xs:
                 res.append(sep.join(x))
-            return np.array(res, dtype=np.object)
```
wenbingl, line 150: according to the warning, 
DeprecationWarning: `np.object` is a deprecated alias for the builtin `object`. To silence this warning, use `object` by itself

should we use `object' or np.object_ here?

wenbingl (commented):
Nice fix

test/test_pyops.py
edgchen1: thanks for the suggestion. updated np.object_ to object

skottmckay (approved):

wenbingl (approved):

PR #343, opened 2023-01-11
Added initial CustomOp template generator

sayanshaw24:
Creates `custom_op_template.hpp` file in `tools` directory when run with ONNX model.

Usage: `python tools\gen_customop_template.py onnx_model_filepath`


tools/gen_customop_template.py
```diff
@@ -0,0 +1,67 @@
...
+
+# Load and check ONNX model
+model = onnx.load(sys.argv[1])
+onnx.checker.check_model(model)
+
+# Get input and output type count from the ONNX model using protobuf information
+input_count = 0
+output_count = 0
+
+# Note: if there is no custom op node in the graph, GetInputTypeCount() and
```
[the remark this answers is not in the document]
sayanshaw24: Yep, thanks! Added that.

tools/gen_customop_template.py
```diff
@@ -0,0 +1,67 @@
...
+Generates the C++ template code for a new custom operator.
+
+Usage: $ python tools\gen_customop_template.py onnx_model_filepath
+'''
+
+import onnx
+import sys
+
+# Load and check ONNX model
+model = onnx.load(sys.argv[1])
```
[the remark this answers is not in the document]
sayanshaw24: Sounds good!

tools/gen_customop_template.py
```diff
@@ -0,0 +1,67 @@
+'''
+Generates the C++ template code for a new custom operator.
```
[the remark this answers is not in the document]
sayanshaw24: Will do.

tools/gen_customop_template.py
```diff
@@ -0,0 +1,98 @@
...
+    custom_op_node_exists = False
+
+    for node in model.graph.node:
+        # Find CustomOp node using domain
+        if node.domain == "ai.onnx.contrib" or node.domain == "com.microsoft.extensions":
+            custom_op_node_exists = True
+            input_count = len(node.input)
+            output_count = len(node.output)
+
+    if not custom_op_node_exists:
```
edgchen1, line 31: what if there is more than one?

sayanshaw24: From my conversation with @wenbingl I believe we are just handling one custom op node case as of now - this is just an initial wip, there is also more research to be done on shape inference to support the template more extensively.

tools/gen_customop_template.py
```diff
@@ -0,0 +1,98 @@
...
+    // mapping of TensorProto.DataType value
+    ONNXTensorElementDataType {custom_op}::GetOutputType(size_t /*index*/) const {{
+      return ONNX_TENSOR_ELEMENT_DATA_TYPE_STRING;
+    }};
+    '''
+
+    hpp_file = customop_template.format(custom_op = op, op_name = op_name, input_type_count = input_type_count, output_type_count = output_type_count)
+
+    # Write code to C++ temmplate file
+    new_op = "tools/custom_op_template.hpp"
```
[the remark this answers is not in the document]
sayanshaw24: yep! i'll change it

tools/gen_customop_template.py
wenbingl: Generally, either you need to handle the case, or print a message to let the user know.

tools/gen_customop_template.py
```diff
@@ -0,0 +1,98 @@
...
+
+import argparse
+import onnx
+import pathlib
+import sys
+
+# Load and check ONNX model
+def load_onnx_model(model_path):
+    model = onnx.load(model_path)
+    onnx.checker.check_model(model)
```
wenbingl, line 15: Did you verify that the model check passes if there is a custom operator in the model?

wenbingl (commented):
Can you also add some examples in tutorial to demonstrate how it works in end to end?

tools/gen_customop_template.py
sayanshaw24: Yep! Will add a test for the same.

tools/gen_customop_template.py
sayanshaw24: Sounds good!

tools/gen_customop_template.py
```diff
@@ -0,0 +1,109 @@
...
+    ONNXTensorElementDataType {custom_op}::GetOutputType(size_t /*index*/) const {{
+      return ONNX_TENSOR_ELEMENT_DATA_TYPE_STRING;
+    }};
+    '''
+
+    hpp_file = customop_template.format(custom_op = op, op_name = op_name, input_type_count = input_type_count, output_type_count = output_type_count)
+
+    # Write code to C++ template file
+    with open(customop_template_filepath,'wt') as file:
+        print("Added C++ CustomOp template code to output filepath: " + str(customop_template_filepath) + "\n")
```
edgchen1, line 77: print() will add a newline at the end by default.
f-strings are a convenient shorthand (https://docs.python.org/3/reference/lexical_analysis.html#f-strings):
```suggestion
        print(f"Added C++ CustomOp template code to output filepath: {customop_template_filepath}")
```

tools/gen_customop_template.py
```diff
@@ -0,0 +1,109 @@
...
+def get_io_count(model):
+    input_count = 0
+    output_count = 0
+    custom_op_node_exists = False
+
+    print("Note: This C++ CustomOp template generator currently only supports models with one custom op node.\n________\n")
+    for node in model.graph.node:
+        # Find CustomOp node using domain
+        if node.domain == "ai.onnx.contrib" or node.domain == "com.microsoft.extensions":
+            custom_op_node_exists = True
```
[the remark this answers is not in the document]
sayanshaw24: this assert statement could also be triggered if there were no custom op nodes either right? but sounds good, will change it in similar fashion and add a note mentioning future support for multiple custom op nodes.

test/test_tools_customop_template.py
```diff
@@ -0,0 +1,34 @@
...
+
+    # check input and output type count of models extracted by template generator
+    def check_io_count(self, model_name, output_path, expected_input_count, expected_output_count):
+        model_path = os.path.join(test_data_dir, model_name)
+        input_count, output_count = gen_customop_template.main([model_path, output_path])
+        self.assertEqual(input_count, expected_input_count)
+        self.assertEqual(output_count, expected_output_count)
+
+    def test_template(self):
+        gpt2tok_template_output_path = os.path.join(tools_dir, "custom_op_template_gpt2tok.hpp")
```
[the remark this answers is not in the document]
sayanshaw24: right, will look into it.

tools/gen_customop_template.py
sayanshaw24: right, missed replacing that with a f-string, thanks!

wenbingl: given ORT stops support of Python < 3.6, we can switch to f-string now

tools/gen_customop_template.py
```diff
@@ -0,0 +1,110 @@
...
+
+def main(args):
+    args = parse_args(args)
+    model = load_onnx_model(args.onnx_model_filepath)
+    input_count, output_count = get_io_count(model)
+    create_hpp(customop_template_filepath = args.customop_template_filepath, op = "CustomOp", op_name = "CustomOpName", input_type_count = input_count, output_type_count = output_count)
+    return input_count, output_count
+
+if __name__ == "__main__":
+    import sys
```
wenbingl, line 109: Did you use any Python lint tool to check your source code, like the flake8? https://flake8.pycqa.org/en/latest/.

The tool should suggest not import any module in the middle of source file.

test/test_tools_customop_template.py
```diff
@@ -0,0 +1,37 @@
...
+        input_count, output_count = gen_customop_template.main([model_path, output_path])
+        self.assertEqual(input_count, expected_input_count)
+        self.assertEqual(output_count, expected_output_count)
+
+    def test_template(self):
+        template_output_path = os.path.join(test_data_dir, "generated")
+        os.mkdir(template_output_path)
+
+        gpt2tok_template_output_path = os.path.join(template_output_path, "custom_op_template_gpt2tok.hpp")
+        self.check_io_count(model_name = "custom_op_gpt2tok.onnx", output_path = gpt2tok_template_output_path, expected_input_count = 1, expected_output_count = 2)
```
wenbingl, line 31: Where did you get these two ONNX files? If they are generated by some Python script, we may just add the generation script in this test file here.

tools/gen_customop_template.py
sayanshaw24: Got it, will change that.

test/test_tools_customop_template.py
sayanshaw24: Sounds good.

wenbingl (approved):

PR #344, opened 2023-01-11
Android package build updates

edgchen1:
- Generate AAR with associated artifacts (javadoc.jar, sources.jar, pom)
- Fix JAR/AAR package descriptions
- Remove `gradle -b build-android.gradle` invocation from CMake. This creates the AAR and it is not necessary to run it for each Android ABI build.
- Support building on Windows

TODO:
- Support snapshot (i.e., pre-release) versions. This may come in a later PR.


wenbingl:
Given there is already gradle + cmake to be the build scripting language, I am not favor of adding new Python dependencies on building package, which adds additional maintenance skill requirements.

It's OK to use Python in the ci pipeline for generating the official package. But the local development and testing, I think gradle + cmake is already enough.

edgchen1:
> Given there is already gradle + cmake to be the build scripting language, I am not favor of adding new Python dependencies on building package, which adds additional maintenance skill requirements.
> 
> It's OK to use Python in the ci pipeline for generating the official package. But the local development and testing, I think gradle + cmake is already enough.

It is possible to call CMake and gradle directly to produce the AAR. build_aar.py provides an easier way to do so.

Regarding Python dependencies, note that we already have some before this change (for example, tools/gen_selectedops.py).

One significant benefit is that a Python script can be cross platform. For scripts of non-trivial complexity, this makes a difference. Consider a single Python script vs. a bash script and a batch file/Powershell script which do the same thing. There can be significant maintenance overhead with the latter.

wenbingl:
gen_selectedops is only used for ORT reduced build here. [> gen_selectedops](https://github.com/microsoft/onnxruntime/blob/83ad562826dc4abf79be2863a64ca401af427d3e/tools/ci_build/build.py#L1246)
And only used for ORT+ORT-Extensions scenario.

The 'benefits' of Python dependency you mentioned is actually something I'd like to avoid since we want to move the build workflow into cmake script and let it to be centralized, and all other scripts are just some thin wrappers to save the length of command line.

Another reason is that you can find some useful cmake/gradle script for building from third party code base that can be some 'standard' way to do that. While implementing all of them into Python which is unique and add our efforts for maintenance.

wenbingl:
And the criteria could be whether it can be done by one or multiple cmake command lines. And whether bash/psh or Python script will only be responsible for composing the cmake command line.

cmake/ext_java.cmake
```diff
@@ -168,36 +161,12 @@ add_custom_command(TARGET onnxruntime_extensions4j_jni POST_BUILD COMMAND ${GRAD
 if (CMAKE_SYSTEM_NAME STREQUAL "Android")
   set(ANDROID_PACKAGE_JNILIBS_DIR ${JAVA_OUTPUT_DIR}/android)
   set(ANDROID_PACKAGE_ABI_DIR ${ANDROID_PACKAGE_JNILIBS_DIR}/${ANDROID_ABI})
-  file(MAKE_DIRECTORY ${ANDROID_PACKAGE_JNILIBS_DIR})
-  file(MAKE_DIRECTORY ${ANDROID_PACKAGE_ABI_DIR})
 
   # Copy onnxruntime_extensions4j_jni.so for building Android AAR package
   add_custom_command(TARGET onnxruntime_extensions4j_jni
-    POST_BUILD COMMAND ${CMAKE_COMMAND}
```
wenbingl, line 176: Where does these code logic go?

edgchen1: A few pieces here:
- calling gradlew with build-android.gradle to build the AAR - this is moved to build_aar.py and done once instead of per ABI
- copying the android test app files, copying the AAR to it, running `gradlew assembleDebug assembleDebugAndroidTest` - this is moved to the CI build. previously we had a separate `gradlew connectedDebugAndroidTest` in the CI build so just moved the rest of the test app setup there

build.android_aar
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
edgchen1, line 1: remove this

.pipelines/ci.yml
```diff
@@ -386,6 +386,13 @@ jobs:
     variables:
       OnnxruntimeDirectory: cmake/external/onnxruntime
     steps:
+      - task: UsePythonVersion@0
```
edgchen1, line 389: remove this

wenbingl (approved):

PR #346, opened 2023-01-14
Copy run_android_emulator.py from ORT

edgchen1:
Add copy of run_android_emulator.py to ORT Extensions and use it directly.


.pipelines/android_packaging.yml
```diff
@@ -19,57 +18,46 @@ jobs:
...
-          jdkArchitectureOption: "x64"
-          jdkSourceOption: "PreInstalled"
-
       - script: |
           python ./tools/gen_selectedops.py ./tools/android/package_ops.config
         displayName: "Generate selected ops CMake file"
 
-      - bash: |
-          set -e -x
+      - template: templates/with-running-android-emulator-steps.yml
```
edgchen1, line 25: suggest hiding whitespace when viewing diff

YUNQIUGUO (dismissed):

.pipelines/android_packaging.yml
```diff
@@ -19,57 +18,46 @@ jobs:
...
+        parameters:
+          steps:
 
-          python ./tools/android/build_aar.py \
-            --output-dir $(Build.BinariesDirectory)/android_aar \
-            --config $(buildConfig) \
-            --cmake-extra-defines OCOS_ENABLE_SELECTED_OPLIST=ON
+          # Start switching to jdk 11 after the Android Emulator is started
+          # since Android SDK manager requires java 8
```
skottmckay, line 30: Does the Android SDK manager still require an ancient version of Java?

https://stackoverflow.com/a/65782803 suggests maybe not.

.pipelines/ci.yml
```diff
@@ -383,56 +383,50 @@ jobs:
...
         displayName: Install coreutils and ninja
 
-      - script: |
-          mkdir cmake/external
-          cd cmake/external
-          git clone https://github.com/microsoft/onnxruntime.git
-          cd onnxruntime
-          git checkout v1.12.1
-        displayName: Download Onnxruntime
+      - template: templates/with-running-android-emulator-steps.yml
```
skottmckay, line 397: nit: 'with-running' in the name sounds like the emulator needs to be started previously. would 'run-with-android-emulator-steps' be slightly clearer that it will take care of emulator startup?

tools/android/run_android_emulator.py
```diff
@@ -0,0 +1,99 @@
...
+
+    elif args.stop:
+        with open(args.emulator_pid_file, mode="r") as emulator_pid_file:
+            emulator_pid = int(emulator_pid_file.readline().strip())
+
+        android.stop_emulator(emulator_pid)
+
+
+if __name__ == "__main__":
+    sys.exit(main())
```
skottmckay, line 99: nit: main() doesn't return a value so the sys.exit isn't doing anything meaningful is it?

skottmckay (dismissed):

.pipelines/android_packaging.yml
```diff
@@ -19,57 +18,46 @@ jobs:
...
-
-      - task: JavaToolInstaller@0
-        displayName: Use jdk 11
-        inputs:
-          versionSpec: "11"
-          jdkArchitectureOption: "x64"
-          jdkSourceOption: "PreInstalled"
-
       - script: |
           python ./tools/gen_selectedops.py ./tools/android/package_ops.config
```
wenbingl, line 22: an alternative is to save a converted _selectedops.cmake in some folder and copy it to the /cmake folder, which can skip this conversions step.

tools/android/run_android_emulator.py
wenbingl: I also saw it in other PR as well and never know why sys.exit is needed?

wenbingl (dismissed):

.pipelines/ci.yml
edgchen1: yeah I think that's a better name

.pipelines/android_packaging.yml
edgchen1: that would work too. I'll leave it like this now, would prefer to generate the file instead of checking in a generated file where possible.

tools/android/run_android_emulator.py
edgchen1: I'd read about it somewhere before and forgot until looking again:

https://docs.python.org/3/library/__main__.html#packaging-considerations

I think we don't need it here

.pipelines/android_packaging.yml
edgchen1: it works with a newer version. thanks for the suggestion

skottmckay (approved):

PR #347, opened 2023-01-18
Add BertTokenizer to iOS package ops config.

edgchen1:


YUNQIUGUO (dismissed):

tools/ios/package_ops.config
```diff
@@ -1 +1 @@
-com.microsoft.extensions;1;DecodeImage,EncodeImage
+com.microsoft.extensions;1;DecodeImage,EncodeImage,BertTokenizer
```
wenbingl, line 1: is it in "ai.onnx.contrib" domain?

edgchen1: hm, it is. the op config parsing doesn't seem to care about that now though.

wenbingl (approved):

tools/ios/package_ops.config
skottmckay: the op config parsing probably doesn't matter unless we had different implementations in the old and new domains (which I hope is unlikely). 

the model would care though, and ideally we use the new domain when updating models, so we'll need to also register the tokenizer in the new domain. 

maybe we could add something to only register ops in the new domain when building the mobile packages, but I don't think that would make much of a difference to binary size (would be slightly cleaner though).

edgchen1: updated to also register the tokenizer ops in the new domain. so now the config should be correct. perhaps we can look into stricter validation in the op config parsing later.

wenbingl (approved):

PR #349, opened 2023-01-23
Move the pre/post processing scripts into the python module.

skottmckay:
Move the scripts into the directory the python module is created from.
Update usage/examples.


onnxruntime_extensions/tools/pre_post_processing/pre_post_processor.py
```diff
@@ -21,7 +21,22 @@ class PrePostProcessor:
...
+            inputs: The inputs the model will use if pre-processing is added.
+            onnx_opset:  The ONNX opset to use.
+                         Minimum is 16. 18 or higher is strongly preferred if image resizing is involved due to its
+                         anti-aliasing ability.
+        """
+
+        if onnx_opset < 16:
+            raise ValueError("ONNX opset must be 16 or later.")
+
+        Settings.pre_post_processing_onnx_opset = onnx_opset
```
edgchen1, line 38: thanks for adding the opset parameter. perhaps keep the pre/post processing onnx opset value as a member of this class instead of setting a global value? could also create the custom op checker context instance in this class and share it among Steps (e.g., passed to Step.apply()).

onnxruntime_extensions/tools/pre_post_processing/docs/pre_post_processing/utils.md
```diff
@@ -57,4 +57,12 @@ Classes
...
 
     `producer_idx: int`
+    :
+
+`Settings()`
+:   
+
+    ### Class variables
+
+    `pre_post_processing_onnx_opset`
```
edgchen1, line 67: with the PrePostProcessor ctor taking an opset parameter now, maybe we don't need to expose this variable to users. and if my other comment sounds ok, perhaps we don't need it at all.

edgchen1 (dismissed):

onnxruntime_extensions/tools/pre_post_processing/pre_post_processor.py
skottmckay: That works for most parts but still leaves a gap with getting the opset info to the step implementation of _create_graph_for_step for things like Resize where the graph created is dependent on the opset. Could set an onnx opset value in the base Step class to minimize the changes or could update _create_graph_for_step to add an arg for the onnx opset value.

edgchen1 (approved):
looks good!

PR #350, opened 2023-01-23
Fix PreFast warning

skottmckay:
Convert opencv values to int64_t to avoid prefast warning.

Couldn't repro the warning locally but this will make all types in calculating the memcpy size the same.

### DIFF
--- a/operators/vision/decode_image.cc
+++ b/operators/vision/decode_image.cc
@@ -31,11 +31,13 @@ void KernelDecodeImage::Compute(OrtKernelContext* context) {
 
   // Setup output & copy to destination
   const cv::Size decoded_image_size = decoded_image.size();
+  const int64_t height = decoded_image_size.height;
+  const int64_t width = decoded_image_size.width;
   const int64_t colors = decoded_image.elemSize();  //  == 3 as it's BGR
 
-  const std::vector<int64_t> output_dims{decoded_image_size.height, decoded_image_size.width, colors};
+  const std::vector<int64_t> output_dims{height, width, colors};
   OrtValue* output_value = ort_.KernelContext_GetOutput(context, 0, output_dims.data(), output_dims.size());
   uint8_t* decoded_image_data = ort_.GetTensorMutableData<uint8_t>(output_value);
-  memcpy(decoded_image_data, decoded_image.data, decoded_image_size.height * decoded_image_size.width * colors);
+  memcpy(decoded_image_data, decoded_image.data, height * width * colors);
 }
 }  // namespace ort_extensions


xadupre (dismissed):

joburkho:
`memcpy(decoded_image_data, decoded_image.data, height * width * colors);`

The third param to `memcpy` has type `size_t` ( which is unsigned and is only a 4 byte integer in a 32 bit build ( such as in arm, x86, and armeabi-v7a builds ) ).  Since `height * width * colors` is an `int64_t` ( a signed 8 byte integer ) , then we'll get a signed to unsigned integer warning in all builds and we'll get a 8 byte to 4 byte potential loss of data warning in 32 bit builds.

wenbingl:
> `memcpy(decoded_image_data, decoded_image.data, height * width * colors);`
> 
> The third param to `memcpy` has type `size_t` ( which is unsigned and is only a 4 byte integer in a 32 bit build ( such as in arm, x86, and armeabi-v7a builds ) ). Since `height * width * colors` is an `int64_t` ( a signed 8 byte integer ) , then we'll get a signed to unsigned integer warning in all builds and we'll get a 8 byte to 4 byte potential loss of data warning in 32 bit builds.

So static_cast is also required for memcpy 3rd parameter?

joburkho:
Yes, static_cast is required to be explicit in the conversion from signed to unsigned integer and ( in 32 bit builds ) from 8 byte to 4 byte integer.

Keep in mind, that a static_cast without a range check before that cast is also an issue, but that is a much larger problem throughout the onnxruntime-extensions and onnxruntime code that will eventually to be addressed .... so for now, a static_cast will resolve the immediate issue.
________________________________

skottmckay:
Added gsl so we can use gsl::narrow/narrow_cast going forward.

cmake/externals/gsl.cmake
```diff
@@ -0,0 +1,18 @@
+if (${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.24.0")
+    FetchContent_Declare(
+        GSL
+        URL https://github.com/microsoft/GSL/archive/refs/tags/v4.0.0.zip
+        URL_HASH SHA1=cf368104cd22a87b4dd0c80228919bb2df3e2a14
+        FIND_PACKAGE_ARGS 4.0 NAMES Microsoft.GSL
+        DOWNLOAD_EXTRACT_TIMESTAMP TRUE
```
skottmckay, line 7: This is the same setup used by ORT. Not sure if the FIND_PACKAGE_ARGS is really necessary but figured we do that in ORT for a good reason so should replicate here when the cmake version is new enough.

wenbingl (approved):
The Onebranch CI now fails due to some security warning, but it doesn't block the code merge.

PR #352, opened 2023-01-25
Fix mobile packaging pipeline builds.

edgchen1:


cmake/noexcep_ops.cmake
```diff
@@ -10,11 +10,11 @@ if (OCOS_ENABLE_WORDPIECE_TOKENIZER)
     message(FATAL_ERROR "WORDPIECE_TOKENIZER operator needs c++ exceptions support")
 endif()
 if (OCOS_ENABLE_BLINGFIRE)
-    message(STATUS "BLINGFIRE operator needs c++ exceptions support, enable exceptions by default!")
+    message(FATAL_ERROR "BLINGFIRE operator needs c++ exceptions support")
```
edgchen1, line 13: @wenbingl this looked like a mistake to me since for all other cases it's an error message. let me know if that's not the case.

tools/gen_selectedops.py
```diff
@@ -67,6 +67,11 @@ def _gen_op_to_cmake_flag():
 
 OP_TO_CMAKE_FLAG = _gen_op_to_cmake_flag()
 
+# if key flag is set, flags in value should also be set
+CMAKE_FLAG_DEPENDENCIES = {
+    "OCOS_ENABLE_VISION": ["OCOS_ENABLE_OPENCV_CODECS"],
```
edgchen1, line 72: alternatively, we could update CMakeLists.txt to make setting OCOS_ENABLE_VISION automatically set OCOS_ENABLE_OPENCV_CODECS if not set already.

`OCOS_ENABLE_VISION AND NOT OCOS_ENABLE_OPENCV_CODECS` is an error now:
https://github.com/microsoft/onnxruntime-extensions/blob/c0555471df07a7c139050ca259d26d5a0a5d9143/CMakeLists.txt#L238-L241

what's the preferred approach?

YUNQIUGUO: yes, I encountered the same build error yesterday when build ort-extension ios pods.

cmake/noexcep_ops.cmake
wenbingl: Good  catch!

tools/gen_selectedops.py
wenbingl: try cmake_dependent_option?
https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html

edgchen1: > try cmake_dependent_option? https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html

sorry, I didn't quite follow. cmake_dependent_option() will enable an option based on other option values, right?
we would still need to set both OCOS_ENABLE_OPENCV_CODECS and OCOS_ENABLE_VISION.

wenbingl: yes, if _vision is true and there will be a _codecs option which default value can be ON. Otherwise _codecs is not an option and the default value is OFF. The user don't have to specify the value of _codecs whether turning on or off the _vision.

natke (commented):
@wenbingl Can you please follow up on and finish this PR while Ed is out?

tools/gen_selectedops.py
skottmckay: Are there going to be other opencv based custom ops added that require the codecs but aren't enabled via OCOS_ENABLE_VISION? 

Currently the vision ops have a hard dependency on opencv, so enabling the opencv codecs isn't optional. I'd favor enabling them implicitly as I don't think the user shouldn't need to care/know about the required internal dependencies of OCOS_ENABLE_VISION.

wenbingl: > Are there going to be other opencv based custom ops added that require the codecs but aren't enabled via OCOS_ENABLE_VISION?
> 
> Currently the vision ops have a hard dependency on opencv, so enabling the opencv codecs isn't optional. I'd favor enabling them implicitly as I don't think the user shouldn't need to care/know about the required internal dependencies of OCOS_ENABLE_VISION.

Yes, the dependency mechanism here is unnecessary. Somehow what we should do it is to re-visit the OCOS_ENABLE_* flags for openCV.

These OCOS_ENABLE_* flags were created to be some switches that we build ort-extensions only with the selected operators and their dependencies. Naturally, there will be one flag for each operator, but to avoid the flag flooding, we group some operators that sharing the same 3rd parties and shared one flag.

So, according to this design. related to opencv, there will be two flags, OCOS_ENABLE_OPENCV_CODECS, OCOS_ENABLE_OPENCV_IMGPROC. And other flags should be removed for cleanness.

wenbingl:
> @wenbingl Can you please follow up on and finish this PR while Ed is out?

I will follow up on this PR after the other PR is done.

natke:
> > @wenbingl Can you please follow up on and finish this PR while Ed is out?
> 
> 
> 
> I will follow up on this PR after the other PR is done.

Thank you @wenbingl

cmake/noexcep_ops.cmake
skottmckay: This was set to STATUS in #143 based on blingfire completely resetting the CXX_FLAGS, so the flags to disable exceptions were essentially ignored. 

That doesn't seem like a good setup because if you build with exceptions disabled in ort-ext and enabled in blingfire, if blingfire throws it won't be handled. 

Based on that I think setting it to FATAL_ERROR is more correct

tools/gen_selectedops.py
skottmckay: Why do we group the operators based on dependencies vs. grouping them based on functionality? 

It seems like we have 3 high level groups of operators (or plan to) that we have a OCOS_ENABLE_xxx option for.

- opencv imgproc operators
- opencv imgcodec operators
  -  tbd if we end up with any real operators here
- vision pre/post processing operators

Currently the second 2 have the same dependency - the opencv codecs. We may switch the vision pre/post processing to directly use the png and jpeg libraries to save binary size/complexity, and if we did that there would be no overlap.

Given that, can we keep the 3 OCOS_ENABLE_xxx values and treat those as switch for groups of operators, and set an internal 'enable opencv codes' value based on either of OCOS_ENABLE_OPENCV_CODECS or OCOS_ENABLE_VISION being enabled so that we're not confusing enabling the opencv codec operators in operators\cv2\imgcodecs with enabling building with the opencv2 codecs dependency?

skottmckay:
I can deal with this PR tomorrow if needed.

tools/gen_selectedops.py
wenbingl: OCOS_ENABLE_ flags are purely for the sake of the binary size, and in the most case, the third-party library will take majority of the binary size in the build. I think it would be more straightforward to use the dependency name instead of functionality name which will introduce a new map from functionality to real dependency. Moreover, these flags are controlled by op-to-flag script, so it's not something for end users which doesn't need to create the functional component name for them easily be understandable. 

Initially, I thought imgcodec might depend on the opencv_core. so I put imgcodec flag basing on opencv. If it doesn't, we can make them independent from each other. i.e. OCOS_EANABLE_IMGPROC, OCOS_ENABLE_IMGCODEC, is that good for you?

BTW, these will be something like extensions_text, extensions_vision if we want to split the extensions into the different domain. it would be the right place to use vision as its name.

wenbingl:
> I can deal with this PR tomorrow if needed.

Yes, please.
I have my hands full with something unplanned and may only be available to do it in the next week.

tools/gen_selectedops.py
skottmckay: I can see that for the existing usage the setup works well as operators are fairly 1:1 with a dependency. The pre/post processing complicates things a little.

AFAIK the binary size contribution from a dependency isn't fixed. It depends on what functions in that dependency are used, so if all ops that use opencv codecs are included when building for the vision pre/post processing, that's going to include more parts of the opencv implementation (e.g. I assume imgread will pull in a bunch of file IO code) and increase binary size.

Based on that I would have expected we need a mechanism to include just the required ops and enable the necessary dependencies for these pre-defined package builds.

skottmckay: Keeping it simple for now and using the existing dependency management in cmake.
CMakeLists.txt already handles the two OCOS_ENABLE_* values that need opencv with

```cmake
if(OCOS_ENABLE_CV2 OR OCOS_ENABLE_VISION)
  set(_ENABLE_OPENCV ON)
  message(STATUS "Fetch opencv")
  include(opencv)
endif()
```

and the vision ops are included separately to the cv2 ops so we don't bloat the mobile package due to opencv usage by the cv2 ops.

The other OCOS_ENABLE_* groups are 1:1 between ops and dependency.

If we get more overlapping things in the future we can re-consider if something more complicated is needed.

wenbingl: not sure cmake dependency can solve this specific question. The issue when all OCOS_ENABLE_ flags are turned OFF for selected op build, only OCOS_ENABLE_VISION was enabled and _IMG_CODECS are still OFF, it causes a cmake error.

Instead of adding a dependency in this script, OP_TO_CMAKE_FLAG definition can be changed as {op: [flag_1, flag_2]}. Then DecodeImage op can be put in both CMAKE_FLAG_TO_OPS[OCOS_ENABLE_VISION] and CMAKE_FLAG_TO_OPS[OCOS_ENABLE_OPENCV_CODECS ]. This seems to be an enhanced op-to-flag mapping.

skottmckay: Seems reasonable to me. Updated to use this approach. Added 'ImageReader' so there's an example usage of something in cv2/imgcodecs as well.

skottmckay:
Packaging builds pass with the changes.

wenbingl (approved):

PR #353, opened 2023-01-26
update Python pipelines for release

wenbingl:


wenbingl:
@skottmckay can you please review this PR?

tools/get_vsdevcmd.ps1
```diff
@@ -0,0 +1,32 @@
+# Find all installed Visual Studio with Cmake, and only output the vsdevcmd path of the latest one
+PARAM([int]$outputEnv = 0)
+$latest_valid_vs = ""
+function choose_latter_vs {
+    param([string]$path)
+    if ($global:latest_valid_vs -lt $path) {
+        $cmake_path = $path + "\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe"
```
skottmckay, line 7: Should we support checking for cmake.exe in the PATH by using Get-Command before requiring the VS install has one? If the user has setup their PATH to point to a specific cmake (which is what I've typically done), I would have expected we simply select the latest VS install. 

```
> get-command cmake

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Application     cmake.exe                                          3.25.2.0   C:\Program Files\CMake\bin\cmake.exe
```

.pipelines/OneBranch.PullRequest.yml
```diff
@@ -37,15 +37,15 @@ resources:
...
+  template: v2/OneBranch.NonOfficial.CrossPlat.yml@templates
   parameters:
     cloudvault:
       enabled: false
     globalSdl:
       tsa:
         enabled: false
       binskim:
-        break: true # always break the build on binskim issues in addition to TSA upload
+        break: false # always break the build on binskim issues in addition to TSA upload
```
skottmckay, line 48: Is this a temporary change just to get builds to pass?

.pipelines/ci.yml
```diff
@@ -15,29 +15,28 @@ jobs:
...
       matrix:
-        py310-1121:
+        py310-1131:
           python.version: '3.10'
+          torch.version: 'torch torchvision torchaudio'
+          ort.version: '1.13.1'
+        py39-1121:
+          python.version: '3.9'
           torch.version: 'torch==1.12.1+cpu torchvision==0.13.1+cpu torchaudio==0.12.1+cpu -f https://download.pytorch.org/whl/cpu/torch_stable.html'
           ort.version: '1.12.0'
```
skottmckay, line 31: What's the reason for the ORT version differing? Does it only affect the CI build or will it affect the ort-ext python package dependencies.

.pipelines/wheels_linux.yml
```diff
@@ -3,7 +3,7 @@ jobs:
   timeoutInMinutes: 120
   pool: {vmImage: 'ubuntu-latest', name: 'Linux-CPU'}
   variables:
-    CIBW_BUILD: "cp3{7,8,9}-*"
+    CIBW_BUILD: "cp3{7,8,9,10}-*"
```
skottmckay, line 6: What's the reason this value isn't consistent with the macos and windows configs? Guessing it's the python versions supported but naively I would have expected we support the same versions on all platforms.

tools/get_vsdevcmd.ps1
wenbingl: Now, most Python packages are tending to use cmake pip package and it was required by cibuildwheel, and it works pretty well for Linux an MacOS
However there are couple of some known cmake issues on some big projects like OpenCV for pip cmake on Windows. that's why there is some VS cmake looking up in this PR since that is most stable cmake variant for ci build.
And when you are building Python package, there already is a cmake.exe (installed by cibuildwheel), so I intentionally avoid search PATH for it. Given we only validate cmake VS in pipeline. Is it a good idea to enforce dev use the same one to avoid some cmake configuration issue?

.pipelines/wheels_linux.yml
wenbingl: And win32-wheel is test only and the official release packages are produced by Onebranch.official because of code sign.

tools/get_vsdevcmd.ps1
skottmckay: I think this is fine of a CI as we control the VS version installed.

Not sure I'd enforce it for dev usage though. Say someone only has VS 2019. I assume that means an older version of cmake. If they wanted to do some dev work with a 3rd party library that required a newer version the setup wouldn't support that by default. Maybe we could add an explicit env var for the cmake path to allow overriding.

skottmckay (approved):

tools/get_vsdevcmd.ps1
wenbingl: I just checked my local VS 2019 installation, which contains cmake 3.20. I guess VS 2019 have some service update to upgrade cmake from time to time.

PR #354, opened 2023-01-29
support mobilebert_ppp

wejoncy:
1. Add a tutorial for Bert
2. add a tool to add pre/post-processing for bert

An End2End model export example via adding pre-post processing stage.

Support
|model|task|tokenizer|source|
|-------|-------|------|---|
 "xlm-roberta-base"|sequence Classification|SencePieceTokenizer|for showing how to export sencePieceTokenizer|
|"google/mobilebert-uncased"|NextSentence|hfBertTokenizer|https://huggingface.co/docs/transformers/model_doc/mobilebert#transformers.MobileBertForNextSentencePrediction|
| "csarron/mobilebert-uncased-squad-v2"|questionAnswering|hfBertTokenizer|https://huggingface.co/docs/transformers/model_doc/mobilebert#transformers.MobileBertForQuestionAnswering|
| "lordtt13/emo-mobilebert"|sequence Classification|bertTokenizer|https://huggingface.co/docs/transformers/model_doc/mobilebert#transformers.MobileBertForSequenceClassification|


onnxruntime_extensions/tools/pre_post_processing/step.py
```diff
@@ -99,7 +99,7 @@ def __merge(self, first: onnx.GraphProto, second: onnx.GraphProto):
         outputs_to_preserve = None
 
         # special handling of Debug class.
-        if isinstance(self, Debug):
+        #if isinstance(self, Debug):
```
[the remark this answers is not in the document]
skottmckay: If we can do it inside the Debug class instead of having a special case in a base class that's way better.

onnxruntime_extensions/tools/add_pre_post_processing_to_transformers.py
```diff
@@ -0,0 +1,254 @@
...
+import transformers
+import torch
+import onnxruntime
+import numpy as np
+import onnxruntime_extensions
+
+
+# avoid loading model from huggingface multiple times, it's time consuming
+@functools.lru_cache
+def get_tokenizer_and_model_from_huggingface(model_name):
```
skottmckay, line 27: Can we separate things out so that this is purely a script to add the pre/post processing to a model (or update the existing script with that specific code?

Parts to acquire a model and validate could go in tutorials as that's similar to \tutorials\superresolution_e2e.py.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,449 @@
...
+# Licensed under the MIT License.
+
+import onnx
+from collections import OrderedDict
+
+from typing import List, Optional, Tuple, Union
+from ..step import Step
+
+
+class SentencePieceTokenizer(Step):
```
skottmckay, line 11: Need to add unit tests to \test\test_tools_add_pre_post_processing_to_model.py

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,449 @@
...
+            tokenizer: SentencePiece tokenizer from huggingface,
+                usually, we can get it by "tokenizer = transformers.AutoTokenizer.from_pretrained(model_name)"
+            domain: Optional domain of step. ["ai.onnx.contrib" or "com.microsoft.extensions"] Defaults to 'ai.onnx.contrib'
+            name: Optional name of step. Defaults to 'SentencePieceTokenizer'
+
+        """
+        super().__init__(
+            ["inputs", "nbest_size", "alpha", "add_bos", "add_eos", "reverse"], ["input_ids", "attention_mask"], name
+        )
+        self.domain = domain
```
skottmckay, line 24: nit: please follow python naming convention and use '_' prefix for internal values of a class.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,449 @@
...
+
+import onnx
+from collections import OrderedDict
+
+from typing import List, Optional, Tuple, Union
+from ..step import Step
+
+
+class SentencePieceTokenizer(Step):
+    def __init__(self, tokenizer, domain: str = "ai.onnx.contrib", name: Optional[str] = None):
```
skottmckay, line 12: Do we need to take a hard dependency at this level on a specific tokenizer or could you just pass in the required info (bos_token_id and vocab_file) so this code is more generic?

Ideally the Step implementations have minimal external dependencies.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,449 @@
...
+        )
+        input_type_str3, input_shape_str3 = (
+            self._get_input_type_and_shape_strs(graph, 3) if self.input_names[3] in graph_input_names else ("bool", "0")
+        )
+        input_type_str4, input_shape_str4 = (
+            self._get_input_type_and_shape_strs(graph, 4) if self.input_names[4] in graph_input_names else ("bool", "0")
+        )
+        input_type_str5, input_shape_str5 = (
+            self._get_input_type_and_shape_strs(graph, 5) if self.input_names[5] in graph_input_names else ("bool", "0")
+        )
```
skottmckay, line 56: Not quite following all the complexity here. Before each Step is run the PrePostProcessor connects things up with the producers of the input values which includes updating the input_names.

What's the reason for all the additional logic checking individual inputs? I don't recall having to do that for any of the other steps.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,449 @@
...
+        converter_graph.node[node_idx].attribute.extend(token_model_attr)
+
+        return converter_graph
+
+
+class BertTokenizerQATask(Step):
+    def __init__(self, name: Optional[str] = None):
+        """
+        Brief:
+            Just copy input_ids for decoder
```
skottmckay, line 267: Can this be more descriptive? This text becomes part of the documentation and should be meaningful to an external user who isn't familiar with all the internals of the class. i

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,449 @@
...
+        return converter_graph
+
+
+class SequenceClassify(Step):
+    def __init__(self, name: Optional[str] = None):
+        """
+        Brief:
+            Decode the input_ids to text
+        Args:
+            tokenizer: tokenizer from huggingface,
```
skottmckay, line 418: No tokenizer arg

onnxruntime_extensions/tools/pre_post_processing/step.py
```diff
@@ -166,27 +166,37 @@ class Debug(Step):
                 to rename so it's more consistent.
     """
 
-    def __init__(self, num_inputs: int = 1, name: Optional[str] = None):
+    def __init__(self, num_inputs: int = 4, name: Optional[str] = None, custom_func: Optional[callable] = None):
```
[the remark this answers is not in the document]
wejoncy: I think num_inputs should be always equal to the real num of inputs. I want to set it as -1 by default.
Why '4' here, usally, mode of graph wouldn't has inputs more than 4, and may we don't want to add more than 4 outputs if there is one. 
3 inputs are quite normally in NLP and 1 input in CV, '4' is always working good.

Sorry, will add some comments for `custom_func `.
This function usally works for saving the graph. Comparing with add outputs for debug in the whole graph, a small "debug.onnx" might be better and easier to debug.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
wejoncy: Sure. Done.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
wejoncy: Make sence. Will pass a Dict which contains all infos.

onnxruntime_extensions/tools/pre_post_processing/step.py
```diff
@@ -166,27 +166,37 @@ class Debug(Step):
...
         """
         Initialize Debug step
         Args:
             num_inputs: Number of inputs from previous Step to make graph outputs.
```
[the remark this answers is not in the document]
wejoncy: Yes. It's correct. 
Hoping no one will insert multiple Debug steps consecutively, it will yield zeros outputs.

onnxruntime_extensions/tools/pre_post_processing/utils.py
```diff
@@ -39,7 +39,7 @@ def create_custom_op_checker_context(onnx_opset: int):
     """
     context = onnx.checker.C.CheckerContext()
     context.ir_version = onnx.checker.DEFAULT_CONTEXT.ir_version
-    context.opset_imports = {"": onnx_opset, "com.microsoft.extensions": 1}
+    context.opset_imports = {"": onnx_opset, "com.microsoft.extensions": 1, "ai.onnx.contrib": 1}
```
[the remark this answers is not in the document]
wejoncy: One question here. In this way, how can we make tokenizer working on both domain if we didn't register it?
`onnx.checker.check_graph(graph_for_step, checker_context)` will throw a exception.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,449 @@
...
+from ..step import Step
+
+
+class SentencePieceTokenizer(Step):
+    def __init__(self, tokenizer, domain: str = "ai.onnx.contrib", name: Optional[str] = None):
+        """
+        Args:
+            tokenizer: SentencePiece tokenizer from huggingface,
+                usually, we can get it by "tokenizer = transformers.AutoTokenizer.from_pretrained(model_name)"
+            domain: Optional domain of step. ["ai.onnx.contrib" or "com.microsoft.extensions"] Defaults to 'ai.onnx.contrib'
```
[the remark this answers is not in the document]
wejoncy: I though we need to provide an option where users can choose both domain "ai.onnx.contrib" or "com.microsoft.extensions"

onnxruntime_extensions/tools/add_pre_post_processing_to_transformers.py
```diff
@@ -0,0 +1,254 @@
...
+import shutil
+import re
+import tempfile
+import functools
+
+from pathlib import Path
+
+from pre_post_processing import PrePostProcessor, Debug
+from pre_post_processing.steps import *
+from pre_post_processing.utils import create_named_value
```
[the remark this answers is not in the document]
wejoncy: Done.
python -m onnxruntime_extensions.tools.add_pre_post_processing_to_transformers -t "csarron/mobilebert-uncased-squad-v2" "$model_path"

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
wejoncy: Done

onnxruntime_extensions/tools/pre_post_processing/steps/__init__.py
```diff
@@ -3,3 +3,4 @@
 
 from .general import *
 from .vision import *
+from .nlp import *
```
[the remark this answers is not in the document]
wejoncy: Sure. Done

onnxruntime_extensions/tools/add_pre_post_processing_to_transformers.py
wejoncy: I  believe seperating this out is a good idea.  But I got a quetion, How could users provide tokenizers and pass to the tool if there is only pure pre-post function?

onnxruntime_extensions/tools/pre_post_processing/step.py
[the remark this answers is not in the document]
skottmckay: '4' seems arbitrary and maybe it's cleaner/more consistent to simply require the user of the Debug class to specify the number of inputs if it's more than 1, instead of attempting to over-allocate and reducing the number of inputs later. The over-allocation also potentially breaks logic in Step.connect where it validates an IoMapEntry, as input_names won't necessarily match the real number of inputs.

Regarding outputs, I would have expected we want to provide debug outputs for all the inputs as we don't know which are important. If the user finds that noisy they can use an IoMapEntry to reduce the inputs to the Debug step.

Regarding custom_func, I'm wary of polluting the Step API for a temporary feature that might only be useful when someone is developing the Step. They could hardcode writing out the onnx model if that's the main usage, or create a pipeline with the Step last and run it against an empty model. 

That seems different to the Debug step adding outputs which is more about seeing the actual values when running the full model than checking the structure of the graph for the step.

onnxruntime_extensions/tools/pre_post_processing/step.py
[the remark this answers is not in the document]
skottmckay: Why will it yield empty output? Shouldn't it always be the case that it passes through all the inputs and it doesn't matter how many in a row there are? You might get names line x_next_next etc. though.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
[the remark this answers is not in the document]
skottmckay: I want to keep the legacy ai.onnx.contrib out of all the new infrastructure.

onnxruntime_extensions/tools/add_pre_post_processing_to_transformers.py
skottmckay: I can imagine you might have some helpers in the script that uses the Step that know about a specific tokenizer and can extract the necessary info from it. Possibly just a case of moving that logic up a level instead of doing it inside the Step.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
skottmckay: Can we have a flag in the ctor which chooses between a single input or 5 and branch on that in the code?

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
skottmckay: A Dict is a little opaque. Could we maybe define a dataclass with the 3 fields so it's obvious what's required?

onnxruntime_extensions/tools/pre_post_processing/step.py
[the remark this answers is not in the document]
wejoncy: Fixed. And I have add a UT to test it.

The reason here is that all outputs generated by the last DebugStep will be filter out.

Current solution is return a empty graph.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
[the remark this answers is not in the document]
wejoncy: Got it.

onnxruntime_extensions/tools/pre_post_processing/step.py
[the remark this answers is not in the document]
wejoncy: Indeed.  num_inputs=4 will lead to a error when built connections.

Yeah, I agree, custom_func is more using for developers.

onnxruntime_extensions/tools/pre_post_processing/step.py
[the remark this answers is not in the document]
wejoncy: Fixed. Have added UTs to test it.
```
        How it works when a string of DebugSteps are added,
            [_next,_debug1,_debug2]---->[_next_next,_debug1_debug,_debug2_debug]---->[_next_next_next,_debug1_debug_debug,_debug2_debug_debug]
        
```

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
wejoncy: Done.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
wejoncy: When we create inputs for `PrePostProcessor`, the function `create_named_value` required uers given a name for the input_value_info, for example
```
create_named_value("text",.....) # required
create_named_value("nbest_size",.....) #optional 
create_named_value("add_bos",.....) #optional 
create_named_value("add_eos",.....) #optional 
create_named_value("reverse",.....) #optional 
```

Is it enough for the purpose?

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
wejoncy: Done. Add a new class for passing  tokenizer parameters

onnxruntime_extensions/tools/add_pre_post_processing_to_transformers.py
wejoncy: Have seperated out into tool and tutorial for bert_e2e models

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,508 @@
...
+                self.input_names[0] = inp
+            elif "start_logits" in inp:
+                self.input_names[1] = inp
+            else:
+                self.input_names[2] = inp
+        for inp in graph_input_names:
+            if self.input_names[2] in inp:
+                self.input_names[2] = inp
+        is_input_names_in_graph = [inp in graph_input_names for inp in self.input_names]
+        assert all(is_input_names_in_graph), f"Input names {self.input_names} not in graph inputs {graph_input_names}"
```
skottmckay, line 421: Why does this get re-ordered inside the implementation vs. using an IOMapEntry to move inputs around if/when needed? 

Would have expected each step to simply define the inputs in the order it wants as we can't predict all the future use cases.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,508 @@
...
+    def _create_graph_for_step(self, graph: onnx.GraphProto, onnx_opset: int):
+        graph_input_names = [inp.name for inp in graph.output]
+
+        # need to reorder input_names by similar to  "input_ids_1", "start_logits", "end_logits"
+        o_inputs = graph_input_names.copy()
+        for inp in o_inputs:
+            if "input_ids_1" in inp:
+                self.input_names[0] = inp
+            elif "start_logits" in inp:
+                self.input_names[1] = inp
```
skottmckay, line 414: When steps are connected the input names are adjusted to match the predecessor graph, which also involves taking into account any IoMapEntry info. As such I don't think you can expect hard coded names to match, or for the order to match the order of graph.output (e.g. the previous graph may have 6 outputs and we're only using a subset for this step).

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,508 @@
...
+            self._get_input_type_and_shape_strs(graph, 3) if self.input_names[3] in graph_input_names else ("bool", "0")
+        )
+        input_type_str4, input_shape_str4 = (
+            self._get_input_type_and_shape_strs(graph, 4) if self.input_names[4] in graph_input_names else ("bool", "0")
+        )
+        input_type_str5, input_shape_str5 = (
+            self._get_input_type_and_shape_strs(graph, 5) if self.input_names[5] in graph_input_names else ("bool", "0")
+        )
+
+        output_shape_str = "batch_size,max_seq_len"
```
skottmckay, line 95: I don't think it's safe to hardcode arbitrary names for dimensions as they're essentially variable names that must have the same value throughout the entire graph. to do that you'd have to guarantee that no other place used the same name for a different value.

always use the dimension name from the input that uses the dynamic dimension where possible. if the dynamic dimension has no matching input value either use an empty string so the rank is correct but it's an unnamed dynamic dimension, or figure out a safe way to come up with a unique name (potentially based on input dim names). 

also in this case 'batch_size' seems inappropriate as the input to the step isn't batched and it's more the number of sentences isn't it so 'num_sentences' is possibly a better name (although I'd probably append  f'_step_{step_num}' to it to make it far more unique.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,508 @@
...
+            {{
+                {build_output_imp()}
+            }}
+            """
+        )
+
+        return converter_graph
+
+
+class BertTokenizerQATaskDecoder(Step):
```
skottmckay, line 386: nit: the 'Task' in a 'Step' seems a superfluous. Would BertTokenizerQADecoder be fine?

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,508 @@
...
+            name: Optional name of step. Defaults to 'SequenceClassify'
+
+        """
+        super().__init__(["logits"], ["index"], name)
+
+    def _create_graph_for_step(self, graph: onnx.GraphProto, onnx_opset: int):
+        graph_input_names = [inp.name for inp in graph.output]
+
+        is_input_names_in_graph = [inp in graph_input_names for inp in self.input_names]
+        assert all(is_input_names_in_graph), f"Input names {self.input_names} not in graph inputs {graph_input_names}"
```
skottmckay, line 488: Step.connect should have wired things up and self.input_names should always be present in graph.output so that individual Steps don't have to repeat that check (i.e. it should be done once in a central place instead of in every single Step._create_graph_for_step).

test/test_tools_add_pre_post_processing_to_model.py
```diff
@@ -170,8 +181,144 @@ def test_pytorch_superresolution(self):
...
+    def build_testModel_and_get_ref_output_for_tokenizer(
+        self, tokenizer_type: str, output_model: Path, pre_steps: List[pre_post_processing.steps.Step] = []
+    ):
+        # import transformers
+        import onnx
+
+        create_named_value = pre_post_processing.utils.create_named_value
+        SentencePieceTokenizer = pre_post_processing.steps.SentencePieceTokenizer
+        TokenizerParam = pre_post_processing.steps.TokenizerParam
+        BertTokenizer = pre_post_processing.steps.BertTokenizer
```
skottmckay, line 196: Given it's test code could we just import * from pre_post_processing.steps and not need these lines of code that seems to just be about removing the 'pre_post_processing.steps.' prefix.

test/test_tools_add_pre_post_processing_to_model.py
```diff
@@ -170,8 +181,144 @@ def test_pytorch_superresolution(self):
...
+        tokenizer_step = BertTokenizer
+        input_text = ("This is a test sentence",)
+        inputs = [create_named_value("inputs", onnx.TensorProto.STRING, ["sentence_length"])]
+
+        if tokenizer_type == "sentencePiece":
```
skottmckay, line 201: nit: Can we make the casing of the names consistent? Maybe 'SentencePiece', 'BertTokenizer' and 'HFBertTokenizer'?

test/test_tools_add_pre_post_processing_to_model.py
```diff
@@ -170,8 +181,144 @@ def test_pytorch_superresolution(self):
...
+            pipeline.add_post_processing([pre_post_processing.steps.BertTokenizerQATaskDecoder(tokenizer_args)])
+            input_model = onnx.load(os.path.join(test_data_dir, "../bert_qa_decoder_base.onnx"))
+        else:
+            input_model = onnx.ModelProto()
+            input_model.opset_import.extend([onnx.helper.make_operatorsetid("", 16)])
+        new_model = pipeline.run(input_model)
+        onnx.save_model(new_model, output_model)
+        return input_text, ref_output
+
+    def test_sentencePiece_tokenizer(self):
```
skottmckay, line 248:
```suggestion
    def test_sentencepiece_tokenizer(self):
```

tutorials/bert_e2e.py
```diff
@@ -0,0 +1,206 @@
...
+import onnxruntime_extensions
+
+# for tokenizer
+import transformers
+import numpy as np
+import onnxruntime
+
+
+# avoid loading model from hugging-face multiple times, it's time consuming
+@functools.lru_cache
```
skottmckay, line 22: Does huggingface not cache locally itself like torchvision does?

tutorials/bert_e2e.py
```diff
@@ -0,0 +1,206 @@
...
+        onnx_config = transformers.models.mobilebert.MobileBertOnnxConfig(config, "sequence-classification")
+        text = ("Hello, my dog is cute",)
+    else:
+        raise ValueError(f"{model_name} is not supported yet.")
+    return tokenizer, model, onnx_config, text
+
+
+def export_backbone(model_name: str, bert_onnx_model: Path):
+    """
+    To export onnx model from huggingface. This model usally has inputs "input_ids", "attention_mask", "token_type_ids",
```
skottmckay, line 50:
```suggestion
    To export onnx model from huggingface. This model usually has inputs "input_ids", "attention_mask", "token_type_ids",
```

tutorials/bert_e2e.py
```diff
@@ -0,0 +1,206 @@
...
+    model_path = args.model_path.resolve(strict=True)
+    canonized_name = re.sub(r"[^a-zA-Z0-9]", "_", args.model_type) + ".onnx"
+
+    if model_path.is_dir():
+        model_path = model_path / canonized_name
+
+    new_model_path = model_path.with_suffix(".with_pre_post_processing.onnx")
+
+    add_pre_post_processing_to_transformers(args.model_type, model_path, new_model_path)
+    verify_results_for_e2e_model(new_model_path, args.model_type, model_path)
```
skottmckay, line 201: nit: Given we're calling this with all 3 args we don't need a default value for new_model_path do we? If not, could we make the order of args args.model_type, model_path, new_model_path?

tutorials/bert_e2e.py
```diff
@@ -0,0 +1,206 @@
...
+    if len(text) == 2:
+        inputs = dict(
+            inputs=np.array([[i] for i in text]),
+        )
+    else:
+        inputs = dict(
+            inputs=np.array([i for i in text]),
+        )
+    real_outputs = session.run([i.name for i in session.get_outputs()], inputs)
+    matched_idx = [i for i, o in enumerate(session.get_outputs()) if list(ref_map_out.keys())[0] in o.name][0]
```
skottmckay, line 134: Can we simplify and provide the list of output names in the call to 'run'? ORT will return the outputs in that order, so if we do that in both calls to 'run' the order will match.

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
```diff
@@ -38,18 +38,13 @@ def imagenet_preprocessing(model_source: ModelSource = ModelSource.PYTORCH):
     # These utils cover both cases of typical pytorch/tensorflow pre-processing for an imagenet trained model
     # https://github.com/keras-team/keras/blob/b80dd12da9c0bc3f569eca3455e77762cf2ee8ef/keras/applications/imagenet_utils.py#L177
 
-    steps = [
-        Resize(256),
-        CenterCrop(224, 224),
-        ImageBytesToFloat()
-    ]
+    steps = [Resize(256), CenterCrop(224, 224), ImageBytesToFloat()]
```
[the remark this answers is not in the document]
wejoncy: Black format ....

Need `# fmt: off`?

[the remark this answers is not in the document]
skottmckay: I think the conclusion was that for ort-ext 'black' wasn't required as there wasn't agreement on its usefulness.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
wejoncy: Sounds good. Thanks.

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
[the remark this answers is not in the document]
wejoncy: Got it. Will restroe it.

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
```diff
@@ -162,18 +152,79 @@ def superresolution(model_file: Path, output_file: Path, output_format: str, onn
...
+    if model_name in ["google/mobilebert-uncased", "csarron/mobilebert-uncased-squad-v2"]:
+        # if two queries required
+        inputs = [create_named_value("inputs", onnx.TensorProto.STRING, [2, "sentence_length"])]
+    else:
+        inputs = [create_named_value("inputs", onnx.TensorProto.STRING, ["sentence_length"])]
+
+    pipeline = PrePostProcessor(inputs)
+    tokenizer_args = TokenizerParam(
+        vocab_or_file=vocab_file,
+        do_lower_case=True,
```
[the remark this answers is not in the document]
wejoncy: Sure.

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
```diff
@@ -194,7 +252,7 @@ def main():
         "--model_source",
         type=str,
         required=False,
-        choices=["pytorch", "tensorflow"],
+        choices=["pytorch", "tensorflow","huggingface"],
         default="pytorch",
```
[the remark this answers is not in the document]
wejoncy: Deleted.

onnxruntime_extensions/tools/pre_post_processing/step.py
```diff
@@ -154,23 +141,25 @@ def _get_input_type_and_shape_strs(self, graph: onnx.GraphProto, input_num: int)
...
     def __init__(self, num_inputs: int = 1, name: Optional[str] = None):
         """
         Initialize Debug step
+        How it works when a string of DebugSteps are added,
+            [_next,_debug1,_debug2]---->[_next_next,_debug1_debug,_debug2_debug]---->[_next_next_next,_debug1_debug_debug,_debug2_debug_debug]
+        
         Args:
-            num_inputs: Number of inputs from previous Step to make graph outputs.
+            num_inputs: Number of inputs from previous Step to make graph outputs. Devs can set any number of inputs to be debugged.
+                (named inputs are not supported though). This class will handle it if the number of inputs is less than the number.
```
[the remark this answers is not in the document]
wejoncy: Okay.
However, if we want to handle a chain of `Debug`,  we still need part of logic to handle it.

onnxruntime_extensions/tools/pre_post_processing/step.py
```diff
@@ -185,8 +174,24 @@ def _create_graph_for_step(self, graph: onnx.GraphProto, onnx_opset: int):
         output_debug_str = ""
         nodes_str = ""
 
+        # don't have to handle the debug node again
+        non_debug_input_names = [inp.name for inp in graph.output if not inp.name.endswith("_debug")]
+        tag_debug_input_names = [inp.name for inp in graph.output if inp.name.endswith("_debug")]
```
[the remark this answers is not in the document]
wejoncy: You are right. Removed.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,508 @@
...
+            Decode the input_ids to text
+        Args:
+            tokenizer_map: some essential infos, If you export tokenizer from hugging-face,
+            such as "tokenizer = transformers.AutoTokenizer.from_pretrained(model_name)"
+            You may need to provide the following:
+                tokenizer_param = TokenizerParam(vocab=tokenizer.vocab, #vocab is dict or file_path)
+            name: Optional name of step. Defaults to 'BertTokenizerQATaskDecoder'
+
+        """
+        # Bugs for default joins in onnxruntime_extensions/tools/pre_post_processing/pre_post_processor.py:187
```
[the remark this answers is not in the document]
wejoncy: In the pose-process stage, pre-process can produce graph outputs, but thsi join operation only joined onnx-model's outputs. However, post-process requires inputs produced by pre-processing.

tutorials/bert_e2e.py
wejoncy: Sure.

tutorials/bert_e2e.py
wejoncy: I would say no. It gives me a new model each time.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
wejoncy: Let me add a prefix, fetching the current timestamp.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,508 @@
...
+    def _create_graph_for_step(self, graph: onnx.GraphProto, onnx_opset: int):
+        graph_input_names = [inp.name for inp in graph.output]
+        for inp in self.input_names:
+            assert inp in graph_input_names
+
+        input_type_str0, input_shape_str0 = self._get_input_type_and_shape_strs(graph, 0)
+        output_shape_str = "batch_size,max_seq_len"
+        assert input_type_str0 == "string"
+
+        assert len(input_shape_str0.split(",")) in [1, 2], "support either one (classification) or two sentences (QA)"
```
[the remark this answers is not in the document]
wejoncy: All inputs are setting as [batch, num_sentences] now.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,508 @@
...
+        return converter_graph
+
+
+class BertTokenizerQATask(Step):
+    def __init__(self, name: Optional[str] = None):
+        """
+        Brief:
+            Just duplicate input_ids for decoder(TokenizerDecoder). For tasks like 'BertTokenizerQATaskDecoder' which need to use the same input_ids for decoder.
+            However, input_ids has its consumers, it will merged and removed in the next step. So we need to duplicate one.
+            The new output 'input_ids_1' will be kept as a new output in graph.
```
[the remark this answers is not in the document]
wejoncy: Removed.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
wejoncy: Removed.

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
```diff
@@ -162,18 +152,79 @@ def superresolution(model_file: Path, output_file: Path, output_format: str, onn
...
+    and output the result in text format for model like QA.
+
+    Args:
+        input_model_file (Path): the model file need to be updated.
+        output_model_file (Path): where to save the final onnx model.
+        onnx_opset (int): default 16. the opset version to use for the final model.
+    """
+    onnx_model = onnx.load(str(input_model_file.resolve(strict=True)))
+    # construct graph input for different tasks
+    if model_name in ["google/mobilebert-uncased", "csarron/mobilebert-uncased-squad-v2"]:
```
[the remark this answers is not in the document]
wejoncy: Yeah. TaskType can workaround it.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,508 @@
...
+class SentencePieceTokenizer(Step):
+    def __init__(self, tokenizer_param: TokenizerParam, name: Optional[str] = None):
+        """
+        Brief:
+            SentencePieceTokenizer is a bit special here. Most likely, users only want to use it like Bert-Tokenizer which has one "Text" input,
+            we support taking the other parameter as optional and use the default value such as the same usage in Hugging-Face.
+
+        Args:
+            tokenizer_map: some essential infos, If you export tokenizer from hugging-face,
+            such as "tokenizer = transformers.AutoTokenizer.from_pretrained(model_name)"
```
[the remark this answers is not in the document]
wejoncy: Refactored the description.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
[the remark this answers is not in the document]
wejoncy: The drawbacks are: Tokenizer implementations are not really support batch-encode/decode.
I think we should make the model align with the Bert backbone, with Batch-size support. But we can description we only support Batch-size=1 at the moment.

test/test_tools_add_pre_post_processing_to_model.py
```diff
@@ -158,24 +185,189 @@ def test_pytorch_superresolution(self):
...
         # whether avx512 is used or not. MacOS seems to be slightly worse though (max of 2)
-        diffs = np.absolute(expected.astype(np.int32) - result.astype(np.int32))
+        diffs = np.absolute(expected.astype(
+            np.int32) - result.astype(np.int32))
         total = np.sum(diffs)
-        print(f'Max diff:{diffs.max()} Total diffs:{total}')
+        print(f"Max diff:{diffs.max()} Total diffs:{total}")
         self.assertTrue(diffs.max() < 3 and total < (result.size / 1000))
 
+    def build_testModel_and_get_ref_output_for_tokenizer(self, tokenizer_type: str, output_model: Path):
```
skottmckay, line 210:
```suggestion
    def build_test_model_and_get_ref_output_for_tokenizer(self, tokenizer_type: str, output_model: Path):
```

test/test_tools_add_pre_post_processing_to_model.py
```diff
@@ -158,24 +185,189 @@ def test_pytorch_superresolution(self):
...
+                 pre_post_processing.utils.IoMapEntry("BertTokenizer", producer_idx=0, consumer_idx=2)])
+            ])
+            input_model = onnx.load(os.path.join(
+                test_data_dir, "../bert_qa_decoder_base.onnx"))
+        else:
+            input_model = onnx.ModelProto()
+            input_model.opset_import.extend(
+                [onnx.helper.make_operatorsetid("", 16)])
+        new_model = pipeline.run(input_model)
+        onnx.save_model(new_model, output_model)
```
skottmckay, line 277: It feels like this part could be a separate function that is called by the tokenizer specific tests if refactored a little, and all the tokenizer specific logic to create tokenizer_impl could be in the specific tests. i.e. having a big if/elif/elif for all the tokenizers in a shared function seems unnecessary. 

The pipeline create/run code could be a function that takes args for pre-processor and post-processor instances and model. Don't add post-processing if post-processor instance was None. If no model provided create the empty one.

That would also mean the HfBertTokenizer_with_decoder specific logic on lines 265 to 271 would be in test_hfBert_tokenizer rather than shared code.

test/test_tools_add_pre_post_processing_to_model.py
```diff
@@ -158,24 +185,189 @@ def test_pytorch_superresolution(self):
...
+                                 "CPUExecutionProvider"])
+
+        result = s.run(
+            None, {s.get_inputs()[0].name: np.array([[*input_text]])})
+
+        self.assertEqual(np.allclose(result[0], ref_output[0]), True)
+        self.assertEqual(np.allclose(result[1], ref_output[2]), True)
+        self.assertEqual(np.allclose(result[2], ref_output[1]), True)
+
+    def test_hfBert_tokenizer(self):
```
skottmckay, line 317:
```suggestion
    def test_hfbert_tokenizer(self):
```

test/test_tools_add_pre_post_processing_to_model.py
```diff
@@ -158,24 +185,189 @@ def test_pytorch_superresolution(self):
...
+
+        self.assertEqual(result[0][0], ref_output[0][0])
+
+    # Corner Case
+    def test_debug_step(self):
+        import onnx
+
+        create_named_value = pre_post_processing.utils.create_named_value
+
+        # multiple DebugSteps are stringed together
```
skottmckay, line 358: nit: this doesn't say what the expected behavior is and it's somewhat hard to infer what we're expecting. we take an existing model, create a new input name but don't add any pre-processing (not sure what happens to that input - is it just a no-op that doesn't become a model output?), and connect up 3 Debug steps with the original model output. 

Can we just add Debug steps in post-processing and say each one should a) pass through a value to the next step and b) create a duplicate output; so for each Debug step we expect 1 additional model output? If so, the check to see whether it works should be based on how many outputs the model originally had and the number of debug steps. 

If that is the case something like 
```
# each Debug step adds a new model output
post_processing = [pre_post_processing.Debug(), pre_post_processing.Debug(), pre_post_processing.Debug()]
...
self.assertEqual(len(new_model.graph.output), len(input_model.graph.output) + len(post_processing))
```

would convey a lot of that information to the developer about the intention of the test.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,357 @@
...
+            add_eos: bool, optional (default = False)
+                  Please see more detail explanation in 
+                  https://www.tensorflow.org/text/api_docs/python/text/SentencepieceTokenizer#args
+
+            name: Optional name of step. Defaults to 'SentencePieceTokenizer'
+
+        """
+        super().__init__(
+            ["inputs", "nbest_size", "alpha", "add_bos", "add_eos",
+                "reverse"], ["input_ids", "attention_mask"], name
```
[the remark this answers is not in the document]
wejoncy: AutoPep8 has the default limit 79.  I changed it to 120 and it works good now.

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
```diff
@@ -162,21 +162,119 @@ def superresolution(model_file: Path, output_file: Path, output_format: str, onn
     onnx.save_model(new_model, str(output_file.resolve()))
 
 
+NLPTaskType = enum.Enum(
+    value="NLPTaskType",
+    names=[
+        ("TokenClassification", 1),
+        ("token-classification", 1),  # alias name
```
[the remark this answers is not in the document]
wejoncy: It's Not necessary. Removed.

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
```diff
@@ -211,13 +313,32 @@ def main():
...
+                 "NextSentencePrediction"],
+        required=False,
+        help="The downstream task for NLP model.",
+    )
+
+    parser.add_argument(
+        "--vocab_file",
+        type=Path,
+        required=False,
+        help="Tokenizer model file for BertTokenizer or sentencePieceTokenizer.",
```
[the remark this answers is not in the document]
wejoncy: NLP task is working with TEXT input, so I believe these models will require a vocab file.
Different tokenizers has different vacab.model/vocab files.

onnxruntime_extensions/tools/pre_post_processing/step.py
```diff
@@ -60,7 +63,20 @@ def apply(self, graph: onnx.GraphProto, checker_context: onnx.checker.C.CheckerC
...
+            if preserver.producer == self:
+                preserver.is_active = True
+                idx = preserver.producer_idx
+                preserver.output = graph_for_step.output[idx].name
+            elif preserver.consumer == self:
+                preserver.is_active = False
+
+        # IOEntryValuePreserver, connect the output to the next step, So we explicitly add the output to the graph
+        additional_outputs = [i.output for i in preserved_outputs if i.is_active]
+        result = self.__merge(graph, graph_for_step, additional_outputs)
```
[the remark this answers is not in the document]
wejoncy: Sounds good.

onnxruntime_extensions/tools/pre_post_processing/step.py
```diff
@@ -180,32 +186,46 @@ def __init__(self, num_inputs: int = 1, name: Optional[str] = None):
...
-        output_debug_str = ""
-        nodes_str = ""
+        non_debug_input_names = [
+            name for name in self.input_names if not name.endswith("_debug")]
 
+        if self._num_inputs > len(non_debug_input_names):
+            raise ValueError(
+                f"Debug step requested {self._num_inputs} inputs, but graph only has {len(non_debug_input_names)}.")
+
+        debug_offset = len(self.input_names)
```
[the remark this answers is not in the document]
wejoncy: Each input will be treat as a new input now.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,357 @@
...
+        )
+        self._tokenizer_param = tokenizer_param
+        # python bool value (True/False) is not supported in c++, so we use 0/1 to represent bool
+        self._optional_args = [nbest_size, alpha,
+                               int(add_bos), int(add_eos), int(reverse)]
```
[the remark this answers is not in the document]
wejoncy: Replaced with named-args

test/test_tools_add_pre_post_processing_to_model.py
```diff
@@ -41,26 +44,34 @@
 class TestToolsAddPrePostProcessingToModel(unittest.TestCase):
     def test_pytorch_mobilenet(self):
         input_model = os.path.join(test_data_dir, "pytorch_mobilenet_v2.onnx")
-        output_model = os.path.join(test_data_dir, "pytorch_mobilenet_v2.updated.onnx")
+        output_model = os.path.join(
+            test_data_dir, "pytorch_mobilenet_v2.updated.onnx")
```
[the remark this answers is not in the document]
wejoncy: Have adjusted my pep8 config with max-lenth=120

test/test_tools_add_pre_post_processing_to_model.py
```diff
@@ -41,26 +44,34 @@
...
-                transforms.ToTensor(),
-                transforms.Normalize(mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225]),
-            ])
+            preprocess = transforms.Compose(
+                [
+                    transforms.Resize(256),
+                    transforms.CenterCrop(224),
+                    transforms.ToTensor(),
+                    transforms.Normalize(mean=[0.485, 0.456, 0.406], std=[
+                                         0.229, 0.224, 0.225]),
```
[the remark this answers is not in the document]
wejoncy: Fixed.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,357 @@
...
+
+        """
+        super().__init__(["inputs"], ["input_ids",
+                                      "attention_mask", "token_type_ids"], name)
+        self._tokenizer_param = tokenizer_param
+
+    def _create_graph_for_step(self, graph: onnx.GraphProto, onnx_opset: int):
+        input_type_str0, input_shape_str0 = self._get_input_type_and_shape_strs(graph, 0)
+
+        input_shape_0 = input_shape_str0.split(",")
```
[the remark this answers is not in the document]
wejoncy: Yes, we can.

The encoder model has batch-size dim in its input, so I think if we didn't provide a batch-dim in the tokenizer. It looks no aligned.

Have support the hidden 'batch-dim' already, if users didn't provide one when add pre-processing, we will add a Unsqueeze op inside.

test/test_tools_add_pre_post_processing_to_model.py
wejoncy: Some of logic in create a pipeline could be re-used by the four test functions.
I have moved the others into their own test function.

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,373 @@
...
+        alpha=1.0,
+        reverse=False,
+        add_bos=False,
+        add_eos=False,
+        name: Optional[str] = None,
+    ):
+        """
+        Brief:
+            SentencePieceTokenizer has actually 6 inputs in definition, but we allow user to provide only text input,
+            and make the others,"nbest_size", "alpha", "add_bos", "add_eos", "reverse"  optional.
```
skottmckay, line 41:
```suggestion
            and make the others, "nbest_size", "alpha", "add_bos", "add_eos", "reverse" optional.
```

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,373 @@
...
+        Args:
+            tokenizer_param: some essential infos to build a tokenizer
+            you can create a TokenizerParam object like:
+                tokenizer_param = TokenizerParam(vocab_size=tokenizer.vocab_size,
+                                    tweaked_bos_id=tokenizer.tweaked_bos_id)
```
skottmckay, line 46:
```suggestion
                tokenizer_param = TokenizerParam(vocab_size=tokenizer.vocab_size,
                                                 tweaked_bos_id=tokenizer.tweaked_bos_id)
```

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,373 @@
...
+        """
+        Brief: This step is used to convert the input text into the input_ids, attention_mask, token_type_ids.
+            It supports an input of a single string for classification models, or two strings for QA models.
+        Args:
+            tokenizer_param: some essential infos to build a tokenizer,
+            You can create a TokenizerParam like this:
+                tokenizer_param = TokenizerParam(vocab=tokenizer.vocab, #vocab is dict or file_path,
+                                    strip_accents = True or False (Optional),
+                                    do_lower_case = True or False (Optional),
+                                    )
```
skottmckay, line 193:
```suggestion
                tokenizer_param = TokenizerParam(vocab=tokenizer.vocab, # vocab is dict or file_path
                                                 strip_accents = True or False (Optional),
                                                 do_lower_case = True or False (Optional)
                                                 )
```

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,373 @@
...
+    def __init__(self, tokenizer_param: TokenizerParam, name: Optional[str] = None):
+        """
+        Brief:
+            Decode the input_ids to text
+        Args:
+            tokenizer_param: some essential infos to build a tokenizer.
+            you can create a TokenizerParam object like:
+                tokenizer_param = TokenizerParam(vocab=tokenizer.vocab, #vocab is dict or file_path)
+            name: Optional name of step. Defaults to 'BertTokenizerQADecoder'
+
```
skottmckay, line 293:
```suggestion
            tokenizer_param: some essential info to build a tokenizer.
                you can create a TokenizerParam object like:
                    tokenizer_param = TokenizerParam(vocab=tokenizer.vocab, #vocab is dict or file_path)
            name: Optional name of step. Defaults to 'BertTokenizerQADecoder'
```

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -0,0 +1,373 @@
...
+        for attr in attrs:
+            token_model_attr.append(onnx.helper.make_attribute(attr, attrs[attr]))
+
+        node_idx = next(i for i, v in enumerate(converter_graph.node) if v.op_type == "BertTokenizerDecoder")
+        converter_graph.node[node_idx].attribute.extend(token_model_attr)
+
+        return converter_graph
+
+
+class ArgMax(Step):
```
skottmckay, line 341: nit: can move to general.py

test/test_tools_add_pre_post_processing_to_model.py
```diff
@@ -108,12 +112,13 @@ def orig_output():
                 transforms.Normalize(mean=[0.5, 0.5, 0.5], std=[0.5, 0.5, 0.5]),
             ])
             input_tensor = preprocess(input_image)
-            input_batch = input_tensor.unsqueeze(
-                0).detach().cpu().numpy()  # create a mini-batch as expected by the model
-            input_batch = np.transpose(input_batch, (0, 2, 3, 1))  # to NHWC format for TF input
+            input_batch = (input_tensor.unsqueeze(
+                0).detach().cpu().numpy())  # create a mini-batch as expected by the model
```
skottmckay, line 116:
```suggestion
            # create a mini-batch as expected by the model
            input_batch = (input_tensor.unsqueeze(0).detach().cpu().numpy())  
```

test/test_tools_add_pre_post_processing_to_model.py
```diff
@@ -174,8 +179,147 @@ def test_pytorch_superresolution(self):
...
+        new_model = pipeline.run(input_model)
+        onnx.save_model(new_model, output_model)
+        return
+
+    def test_sentencepiece_tokenizer(self):
+        output_model = (Path(test_data_dir) / "../sentencePiece.onnx").resolve()
+
+        input_text = ("This is a test sentence",)
+        ref_output = [np.array([[0, 3293, 83, 10, 3034, 149357, 2]]), np.array(
+            [[1, 1, 1, 1, 1, 1, 1]])]
```
skottmckay, line 214:
```suggestion
        ref_output = [np.array([[0, 3293, 83, 10, 3034, 149357, 2]]), 
                      np.array([[1, 1, 1, 1, 1, 1, 1]])]
```

test/test_tools_add_pre_post_processing_to_model.py
```diff
@@ -174,8 +179,147 @@ def test_pytorch_superresolution(self):
...
+        output_model = (Path(test_data_dir) / "../bert_tokenizer.onnx").resolve()
+        input_text = ("This is a test sentence",)
+        ref_output = [
+            np.array([[2, 236, 118, 16, 1566, 875, 643, 3]]),
+            np.array([[0, 0, 0, 0, 0, 0, 0, 0]]),
+            np.array([[1, 1, 1, 1, 1, 1, 1, 1]]),
+        ]
+        # tokenizer = transformers.AutoTokenizer.from_pretrained("lordtt13/emo-mobilebert")
+        tokenizer_parameters = TokenizerParam(vocab_or_file=os.path.join(
+            test_data_dir, "../bert.vocab"), do_lower_case=True)
```
skottmckay, line 244:
```suggestion
        tokenizer_parameters = TokenizerParam(vocab_or_file=os.path.join(test_data_dir, "../bert.vocab"), 
                                              do_lower_case=True)
```

tutorials/bert_e2e.py
```diff
@@ -0,0 +1,210 @@
...
+    and has tensor outputs.
+    """
+
+    # fix the seed so we can reproduce the results
+    transformers.set_seed(42)
+    tokenizer, model, onnx_config, text = get_tokenizer_and_model_from_huggingface(
+        model_name)
+
+    if bert_onnx_model and bert_onnx_model.exists():
+        print("Use cached onnx Model, skip re-exporting the backbone model.")
```
skottmckay, line 68:
```suggestion
        print("Using cached ONNX model, skipping re-exporting the backbone model.")
```

tutorials/bert_e2e.py
```diff
@@ -0,0 +1,210 @@
...
+        tmp_model_path = Path(tmpdir + "/" + canonized_name)
+        onnx_inputs, onnx_outputs = transformers.onnx.export(
+            tokenizer, model, onnx_config, 16, tmp_model_path)
+        shutil.copy(tmp_model_path, bert_onnx_model)
+        return tokenizer, bert_onnx_model, onnx_config
+
+
+def add_pre_post_processing_to_transformers(model_name: str, input_model_file: Path, output_model_file: Path):
+    """construct the pipeline for a end2end model with pre and post processing. The final model can take text as inputs
+    and output the result in text format for model like QA.
```
skottmckay, line 83:
```suggestion
    """Construct the pipeline for an end2end model with pre and post processing. 
    The final model can take text as inputs and output the result in text format for models like Q & A.
```

tutorials/bert_e2e.py
```diff
@@ -0,0 +1,210 @@
...
+
+    Args:
+        model_name (str): Which model to export in hugging-face, it determinate tokenizer and onnx model backbone.
```
skottmckay, line 86:
```suggestion
        model_name (str): Model to export from hugging-face. Used to infer tokenizer and onnx model backbone.
```

tutorials/bert_e2e.py
```diff
@@ -0,0 +1,210 @@
...
+    if not hasattr(tokenizer, "vocab_file"):
+        vocab_file = bert_onnx_model.parent / "vocab.txt"
+        import json
+        with open(str(vocab_file), 'w') as f:
+            f.write(json.dumps(tokenizer.vocab))
+    else:
+        vocab_file = tokenizer.vocab_file
+    tokenizer_type = 'BertTokenizer' if model_name != 'xlm-roberta-base' else 'SentencePieceTokenizer'
+    task_type = 'NextSentencePrediction' if model_name == 'google/mobilebert-uncased' else ''.join(
+        [i.capitalize() for i in onnx_config.task.split('-')])
```
skottmckay, line 101:
```suggestion
    task_type = ('NextSentencePrediction' if model_name == 'google/mobilebert-uncased' 
                 else ''.join([i.capitalize() for i in onnx_config.task.split('-')]))
```

tutorials/bert_e2e.py
```diff
@@ -0,0 +1,210 @@
...
+
+def verify_results_for_e2e_model(model_name: str, input_bert_model: Path, output_model_file: Path):
+    """
+    Args:
+        output_model_file: the onnx model which finalized and needs to be verified
+        model_name: the huggingface model name
+        input_bert_model: the onnx model which is generated by huggingface or user provide
+    """
+    tokenizer, hg_model, _, text = get_tokenizer_and_model_from_huggingface(
+        model_name)
```
skottmckay, line 116:
```suggestion
    tokenizer, hg_model, _, text = get_tokenizer_and_model_from_huggingface(model_name)
```

tutorials/bert_e2e.py
```diff
@@ -0,0 +1,210 @@
...
+        str(input_bert_model.resolve(strict=True)), providers=["CPUExecutionProvider"]
+    )
+    inputs = {key: value.detach().numpy()
+              for key, value in encoded_input.items()}
+    output_name_for_verify = session.get_outputs()[0].name
+    ref_outputs = session.run([output_name_for_verify], inputs)
+
+    # Load tokenizer op
+    session_options.register_custom_ops_library(
+        onnxruntime_extensions.get_library_path())
```
skottmckay, line 133:
```suggestion
    session_options.register_custom_ops_library(onnxruntime_extensions.get_library_path())
```

tutorials/bert_e2e.py
```diff
@@ -0,0 +1,210 @@
...
+          - MobileBert with different tasks
+          - XLM-Roberta with classification task   
+          
+        This tutorial does only server as a example of how to add pre/post processing to a transformer model.
+        It adds tokenizer( sentencePiece/Berttokenizer/HfbertTokenizer in pre-processing) 
+        and classify/Decoder (in postprocessing, serves for different downstream tasks) to the model.
+        
+        Export models from huggingface by default if there is not a onnx model provided.
+        NOTE: if provided a onnx model, you have to make sure your model is matched with the model_type in hugging-face.
+        As we are using the hugging-face tokenizer to do the pre-processing.
```
skottmckay, line 166:
```suggestion
        This tutorial provides an example of how to add pre/post processing to a transformer model.
        It can add a tokenizer (SentencePiece/Berttokenizer/HfbertTokenizer) for pre-processing,
        and a classifier/decoder for post-processing.
        
        Exports models from huggingface by default if an existing onnx model is not provided.
        NOTE: if providing a onnx model, you have to make sure your model is matched with the model_type in hugging-face as we are using the hugging-face tokenizer to do the pre-processing.
```

tutorials/bert_e2e.py
```diff
@@ -0,0 +1,210 @@
...
+        help="Model type.",
+    )
+
+    parser.add_argument(
+        "model_path",
+        type=Path,
+        help="""The onnx model path which needs pre-or-post processing update 
+                        or a dir-path where to save updated-model.
+                        If a onnx model path is provided, the updated model will be written in the sane directory
+                        with a suffix '.with_pre_post_processing.onnx'""",
```
skottmckay, line 190:
```suggestion
        help="""The path to an existing ONNX model or directory name to save a model exported from HuggingFace in.
                This model will be updated to add pre/post processing, and saved in the same location with the suffix 
                '.with_pre_post_processing.onnx'""",
```

tutorials/bert_e2e.py
```diff
@@ -0,0 +1,210 @@
...
+    model_path = args.model_path.resolve(strict=True)
+    canonized_name = re.sub(r"[^a-zA-Z0-9]", "_", args.model_type) + ".onnx"
+
+    if model_path.is_dir():
+        model_path = model_path / canonized_name
+
+    new_model_path = model_path.with_suffix(".with_pre_post_processing.onnx")
+
+    add_pre_post_processing_to_transformers(
+        args.model_type, model_path, new_model_path)
```
skottmckay, line 204:
```suggestion
    add_pre_post_processing_to_transformers(args.model_type, model_path, new_model_path)
```

onnxruntime_extensions/tools/pre_post_processing/step.py
```diff
@@ -180,32 +171,43 @@ def __init__(self, num_inputs: int = 1, name: Optional[str] = None):
         super().__init__(input_names, output_names, name)
 
     def _create_graph_for_step(self, graph: onnx.GraphProto, onnx_opset: int):
-        input_str = ""
-        output_str = ""
-        output_debug_str = ""
-        nodes_str = ""
+        if self._num_inputs > len(self.input_names):
+            raise ValueError(
+                f"Debug step requested {self._num_inputs} inputs, but graph only has {len(self.input_names)}.")
```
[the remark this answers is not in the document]
wejoncy: I meant to use graph.output.

This errror msg is more straight forward when user passed more inputs than actual when compared to raise exception in _get_input_type_and_shape_strs.

onnxruntime_extensions/tools/pre_post_processing/step.py
```diff
@@ -48,7 +48,9 @@ def connect(self, entry: IoMapEntry):
 
         self.input_names[entry.consumer_idx] = entry.producer.output_names[entry.producer_idx]
 
-    def apply(self, graph: onnx.GraphProto, checker_context: onnx.checker.C.CheckerContext):
+    def apply(self, graph: onnx.GraphProto, 
+              checker_context: onnx.checker.C.CheckerContext, 
+              additional_outputs: List[str]):
```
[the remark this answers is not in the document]
wejoncy: Done

skottmckay (approved):

PR #355, opened 2023-02-01
Fix sentencepiece build error

skottmckay:
Combined ORT + ort-ext build for Android fails due to protobuf mismatch. 

Move the setting of SPM_USE_BUILTIN_PROTOBUF to before the populate to fix.


wenbingl (approved):
I noticed this issue, and it worked for web-assembly because there is another definition their cmake.

https://github.com/microsoft/onnxruntime/blob/60290393f3d3afe240b5a5886a5da2b8e2dcb835/cmake/external/extensions.cmake#L30

BTW, Is this change enough for Android combined build?

skottmckay:
Yes. Combined build works with this (assuming exceptions enabled).

PR #358, opened 2023-02-08
security warning of sentencepiece

wenbingl:


cmake/externals/sentencepieceproject.cmake
```diff
@@ -1,20 +1,63 @@
...
+    set_target_properties(libprotoc PROPERTIES
+      FOLDER externals/google/protobuf)
+    set_target_properties(protoc PROPERTIES
+      FOLDER externals/google/protobuf)
+  endif()
 endif()
 
+# To avoid create a complicated logic to build protoc, especailly for mobile platforms, we use the pre-generated pb files
+# Uses the following command line in _deps/spm-src folder to generate the PB patch file if protobuf version is updated
+# git diff -- src/builtin_pb/* | out-file -Encoding utf8 <REPO-ROOT>\cmake\externals\sentencepieceproject.pb.patch
```
edgchen1, line 45:
```suggestion
# git diff -- src/builtin_pb/* | out-file -Encoding utf8 <REPO-ROOT>\cmake\externals\sentencepieceproject_pb.patch
```

.pipelines/OneBranch.Official.yml
```diff
@@ -44,7 +44,7 @@ extends:
       tsa:
         enabled: true
       binskim:
-        break: true # always break the build on binskim issues in addition to TSA upload
```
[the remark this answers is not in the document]
skottmckay: nit: can the comment be updated to mention it's currently `false` for this reason?

cmake/externals/sentencepieceproject.cmake
```diff
@@ -23,29 +66,22 @@ FetchContent_Declare(
 )
 FetchContent_GetProperties(spm)
 
-if(_ONNXRUNTIME_EMBEDDED)
-  set(SPM_USE_BUILTIN_PROTOBUF OFF CACHE INTERNAL "")
+set(SPM_USE_EXTERNAL_ABSL OFF CACHE BOOL "Use external absl" FORCE)
```
skottmckay, line 69: If this is an embedded build should we be pointing it to the absl library ORT is using (unless onnxruntime_DISABLE_ABSEIL is on)?

cmake/externals/sentencepieceproject.cmake
```diff
@@ -1,20 +1,63 @@
 # spm is abbreviation of sentencepiece to meet the path length limits on Windows
+if(NOT _ONNXRUNTIME_EMBEDDED)
+  # If extensions wasn't built in ORT, we create fetchcontent the same 3rd party library as ORT
+  # So extensions is always consitent on the 3rd party libraries whether its build in ORT or not
```
skottmckay, line 4:
```suggestion
  # So extensions is always consistent on the 3rd party libraries whether its build in ORT or not
```

cmake/externals/sentencepieceproject.cmake
```diff
@@ -1,20 +1,63 @@
...
+    FOLDER externals/google/protobuf)
+  if(NOT CMAKE_SYSTEM_NAME STREQUAL "Android")
+    set_target_properties(libprotoc PROPERTIES
+      FOLDER externals/google/protobuf)
+    set_target_properties(protoc PROPERTIES
+      FOLDER externals/google/protobuf)
+  endif()
 endif()
 
+# To avoid create a complicated logic to build protoc, especailly for mobile platforms, we use the pre-generated pb files
```
skottmckay, line 43:
```suggestion
# To avoid creating complicated logic to build protoc, especially for mobile platforms, we use the pre-generated pb files
```

skottmckay: Something like https://marketplace.visualstudio.com/items?itemName=EWoodruff.VisualStudioSpellCheckerVS2022andLater is very handy for highlighting spelling errors in comments.

cmake/externals/sentencepieceproject.cmake
wenbingl: this flag is misleading, and it doesn't work as what we expected.
https://github.com/google/sentencepiece/blob/c5a49eb5de2867f4f31aca0abc1c70c35605a5f6/CMakeLists.txt#L142

Actually, to make abseil external library work, it requires another big patch on SPM library and may be do it later.

.pipelines/OneBranch.Official.yml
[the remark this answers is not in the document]
wenbingl: We should take it as the first priority right now and track it as a work item.
Will turn it back on as soon as it is done.

wenbingl:
@skottmckay @edgchen1 , any more comment?
We need to merge this PR ASAP to fix the ci pipeline which is broken now.

test/test_tools_add_pre_post_processing_to_model.py
```diff
@@ -36,7 +37,7 @@
 #     assert(len(labels) == 1000 if is_pytorch else 1001)
 #     return labels
 
-
+@unittest.skipIf(LooseVersion(ort.__version__) < LooseVersion("1.13"), "only supported in ort 1.13 and above")
```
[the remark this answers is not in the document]
edgchen1: also does this mean that the pre_post_processing tools only officially support use with ORT 1.13+?

@skottmckay

cmake/externals/sentencepieceproject.cmake
```diff
@@ -1,20 +1,63 @@
 # spm is abbreviation of sentencepiece to meet the path length limits on Windows
+if(NOT _ONNXRUNTIME_EMBEDDED)
+  # If extensions wasn't built in ORT, we create fetchcontent the same 3rd party library as ORT
```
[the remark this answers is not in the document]
wenbingl: Yes, so it will meet the static linking requirement.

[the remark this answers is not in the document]
edgchen1: thanks for clarifying. may be good to be more specific in this comment too :)

CMakeLists.txt
```diff
@@ -87,18 +93,19 @@ if(NOT CC_OPTIMIZE)
...
+  set(CMAKE_EXE_LINKER_FLAGS  "${CMAKE_EXE_LINKER_FLAGS} /DYNAMICBASE")
+  check_cxx_compiler_flag(-guard:cf HAS_GUARD_CF)
+  if (HAS_GUARD_CF)
+    set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} /guard:cf")
+    set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /guard:cf")
+    set(CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO} /guard:cf")
+    set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} /guard:cf")
+    set(CMAKE_C_FLAGS_MINSIZEREL "${CMAKE_C_FLAGS_MINSIZEREL} /guard:cf")
+    set(CMAKE_CXX_FLAGS_MINSIZEREL "${CMAKE_CXX_FLAGS_MINSIZEREL} /guard:cf")
+    set(CMAKE_EXE_LINKER_FLAGS  "${CMAKE_EXE_LINKER_FLAGS} /guard:cf")
```
skottmckay, line 106: Can we just set CMAKE_C_FLAGS and CMAKE_CXX_FLAGS if config is not DEBUG? 

Do we also need to limit when we set the linker flags or it's find to set those for debug builds as well?

skottmckay (approved):
Signing off to get it checked in and the pipelines working. Can refine cmake setup later if needed.

cmake/externals/sentencepieceproject.cmake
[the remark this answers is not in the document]
wenbingl: actually, PB is only the first step, and ABSEIL could be the next.
The idea to make ort combined build is consistent with our package, like 3rd party library, and no diverted code path.

CMakeLists.txt
wenbingl: The code copied from ORT, and it has to be proved to work well for binskim. So let's take it a solution

PR #359, opened 2023-02-09
re_enable exceptions if some modules are included in  building

wejoncy:


cmake/noexcep_ops.cmake
```diff
@@ -1,20 +1,32 @@
...
-if (OCOS_ENABLE_BLINGFIRE)
-    message(FATAL_ERROR "BLINGFIRE operator needs c++ exceptions support")
-endif()
-if (OCOS_ENABLE_SPM_TOKENIZER)
-    message(FATAL_ERROR "SPM_TOKENIZER operator needs c++ exceptions support")
-endif()
-if (OCOS_ENABLE_CV2 || OCOS_ENABLE_OPENCV_CODECS || OCOS_ENABLE_VISION)
-    message(FATAL_ERROR "the operators depending on opencv needs c++ exceptions support")
-endif()
+macro(TryReEnableExceptionsIfNeed OCOS_ENABLE_CPP_EXCEPTIONS)
```
wenbingl, line 1: Re-enabling seems a weird thing.
We can add a new OCOS_EXCEPTION_MODE as 'auto', 'on', 'off'. In the auto mode, it will try to disable exception as much as possible, if some ops are dependent on C++ exceptions, it will automatically turn on C++ exceptions.

natke:
> BTW, which product requires it?

This is for the combined minimal build scenario. See this PR for reference: https://github.com/microsoft/onnxruntime/pull/14540

cmake/noexcep_ops.cmake
skottmckay: Can we define the problem to solve as:
- we are building as part of ORT (_ONNXRUNTIME_EMBEDDED), 
- ORT has exceptions disabled (ORT_NO_EXCEPTIONS), 
- we need to ability to build ort-ext with exceptions enabled (override the OCOS_ENABLE_CPP_EXCEPTIONS value) so the code in ort-ext can provide the try/catch barrier between code that requires exceptions and may throw, and ORT.

We could potentially limit this re-enabling of exceptions in ort-ext to builds that include dependencies that require exceptions.

skottmckay: We also need a #define so the implementation in ort-ext can catch the exception but not rethrow it when ORT has exceptions disabled.

e.g. something like this would provide a layer that avoids passing an exception up to ORT, but needs the ifdef to know when it can rethrow and when it can't. also needs the infrastructure to make the try/catch a no-op in a build with exceptions disabled

```cpp
    OrtCustomOp::KernelCompute = [](void* op_kernel, OrtKernelContext* context) {
      OCOS_TRY {
        static_cast<TKernel*>(op_kernel)->Compute(context);
      }
      OCOS_CATCH(const Exception& ex) {
        OCOS_HANDLE_EXCEPTION([&]() {
#ifdef OCOS_CONTAIN_EXCEPTIONS
          // TODO: Update CustomOp API to allow KernelCompute (and CreateKernel) to return a status otherwise we have
          // to abort(). just log in this example code
          std::cerr << "Exception in KernelCompute: " << ex.Message() << "\n";
#else
          // rethrow
          throw;
#endif
        });
      }
```

wenbingl: There are already some macros in existing code https://github.com/microsoft/onnxruntime-extensions/blob/36a9c991a20a9fd9eb49909117e2528d64fdeda5/includes/onnxruntime_customop.hpp#L40

The problem is to try to re-enable the exception for these libraries which cannot be compiled without exceptions. I think another solution is to create a macro to wrap any API from 3rd party library.

skottmckay: I expect we also need to wrap the RegisterCustomOps code with the same error handling - basically any entry point that ORT can call needs something similar to what ORT's API_IMP_BEGIN/API_IMPL_END does (conditional try/catch that adjusts based on whether exceptions are enabled/disabled/propagated). My example expanded that macro to show what it was doing, but I'd hope for something similar to the ORT implementation in [onnxruntime_c_api.cc](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/onnxruntime_c_api.cc).

If we do the calls in the CustomOpBase class that can throw (can probably limit that to KernelCreate and KernelCompute) and RegisterCustomOps does that cover all code in the custom op implementations?

Do we do any initialization of static globals in 3rd party libraries outside of that? 

My initial thought was that trying to wrap calls to 3rd party libraries at the custom op level would require wrapping code in more places as you'd have to check the implementation of each custom op that uses a 3rd party libary that throws. Would that not the case? 

Is there benefit of doing it on the custom op level vs. doing it in CustomOpBase/RegisterCustomOps?

wenbingl: Most of operators are already implemented as C++ exceptions free. so, it should be kept as it was. Then let's deal with operators with 3rd party library C++ exceptions. we can wrap APIs calling to be exceptions free. If it is longsome, we can apply the macro in the higher-level function.

One missing part here I want to discuss more is how we configure C++ exceptions build flag for the project. In no-exceptional case, the ideal case is only the required C++ source files and libraries will be compiled with exception flag.  We should avoid enabling exceptions for all source files to make sure unintentional exceptions happen on the files that don't really depend on C++ exceptions.

PR #361, opened 2023-02-15
Possible fix for "FetchContent" builds, always use ONNXRUNTIME_PKG_DIR when defined.

matheusgomes28:
I'm not sure what the purpose of `_ONNXRUNTIME_EMBEDDED` is, but the if branch checking it's existence in `ext_ortlib.cmake` uses the path `${CMAKE_SOURCE_DIR}/incluude/onnxruntime/core/session` to attempt to find libs and headers for ORT.

When using ` `${CMAKE_SOURCE_DIR}/../include/onnxruntime/core/session`, unless you have the ONNX library in located in `include/onnxruntime/core/session` above the project root (a bit of a random place?), it will fail to find headers and libs.

This will cause the variable `ONNXRUNTIME_PKG_DIR` to be ignored, but the user can set that variable to the root of ONNX runtime, causing it to actually find the lib.

My suggestion is to always use `${ONNXRUNTIME_PKG_DIR}` when defined, as users like myself, can pull prebuilt packages (or build from source), find the libs and headers with CMake find package and then simply define `ONNXRUNTIME_PKG_DIR` prior to calling `FetchContent`.

I've left the previous branch in there, as a secondary check in case there's some need. But from a glance, we can just remove that branch as it's not really doing anything useful?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -154,7 +154,7 @@ include(FetchContent)
 
 # PROJECT_IS_TOP_LEVEL is available until 3.21
 get_property(not_top DIRECTORY PROPERTY PARENT_DIRECTORY)
-if(not_top)
+if(not_top AND ONNXRUNTIME_ROOT)
   set(_ONNXRUNTIME_EMBEDDED TRUE)
 endif()
 include(ext_ortlib)


wenbingl:
_ONNXRUNTIME_EMBEDDED  is a flag when ONNXRuntime build with -use-extensions flag, and ort-extensions will be compiled as ORT internal static library and linked.

I guess your projects take ort-extensions as a sub project which makes _ONNXRUNTIME_EMBEDDED be defined. And it triggers a buggy behavior of ort-extensions. It can be fixed by adding an ONNXRUNTIME flag check, like if ONNXRUNTIME_ROOT is defined here. https://github.com/microsoft/onnxruntime-extensions/blob/36a9c991a20a9fd9eb49909117e2528d64fdeda5/CMakeLists.txt#L157

wenbingl:
> > _ONNXRUNTIME_EMBEDDED is a flag when ONNXRuntime build with -use-extensions flag, and ort-extensions will be compiled as ORT internal static library and linked.
> > I guess your projects take ort-extensions as a sub project which makes _ONNXRUNTIME_EMBEDDED be defined. And it triggers a buggy behavior of ort-extensions. It can be fixed by adding an ONNXRUNTIME flag check, like if ONNXRUNTIME_ROOT is defined here.
> > https://github.com/microsoft/onnxruntime-extensions/blob/36a9c991a20a9fd9eb49909117e2528d64fdeda5/CMakeLists.txt#L157
> 
> Thanks for replying. And yes, I am using `FetchContent_Declare` and `FetchContent_MakeAvailable` to pull this project. The `not_top` flag is set as it's the result of querying CMake's `PARENT_DIRECTORY` property. So it's kind of a limbo when the project is being compiled as part of a different project with `FertchContent_*`.
> 
> Also, checking if `ONNXRUNTIME` is set at that line won't do anything because `ONNXRUNTIME` is set by[ finding the library `onnxruntime`](https://github.com/microsoft/onnxruntime-extensions/blob/36a9c991a20a9fd9eb49909117e2528d64fdeda5/CMakeLists.txt#L511), which itself depends on [`ONNXRUNTIME_LIB_DIR` defined in `ext_ortlib.cmake`](https://github.com/microsoft/onnxruntime-extensions/blob/36a9c991a20a9fd9eb49909117e2528d64fdeda5/cmake/ext_ortlib.cmake#L3) that is using `_ONNXRUNTIME_EMBEDDED`.
> 
> By inverting the "if" checks in this PR, users can still figure that they can set `ONNXRUNTIME_PKG_DIR` to find the prebuild libraries, and you also get the `_ONNXRUNTIME_EMBEDDED` behaviour if it's not set.
> 
> @wenbingl let me know if that still sounds like a bad solution, I'm happy to think of an alternative.

here is where _ONNXRUNTIME_EMBEDDED start. https://github.com/microsoft/onnxruntime/blob/b1abb8c656c597bf221bd85682ae3d9e350d9aba/cmake/external/extensions.cmake#L41

the _ONNXRUNTIME_EMBEDDED means ort-extensions is embedded into ONNXRUNTIME source tree and be built as its sub module. So cmake encounters the line of _not_top_, the **ONNXRUNTIME_ROOT** should be already defined. And in your case, ONNXRUNTIME_ROOT is never defined, and _ONNXRUNTIME_EMBEDDED will be false.

matheusgomes28:
Gotcha, pushed the changes required (reverted the changes to the `ext_onnx.cmake` file).

wenbingl (approved):

wenbingl:
Thanks for the fixing!

PR #362, opened 2023-02-16
Fix opset 18 issues and bug due to ORT Resize issue

skottmckay:
- Fix Split(18) requiring num_outputs.
- Calculate `sizes` in Resize instead of using the simpler `scales`
  - ORT implementation does not round correctly when applying scales
- Update center crop to use float so we are more accurate in choosing the crop area.
- Fix minor issue with Debug step by only adding values that are altered to the renaming graph inputs.
- Update unit tests expected output due to the change in Resize using sizes instead of scales.
- Crop e2e example input so before/after image covers same area.


test/test_tools_add_pre_post_processing_to_model.py
```diff
@@ -143,12 +143,15 @@ def test_pytorch_superresolution(self):
...
-        expected_output_image_path = os.path.join(test_data_dir, "test_superresolution.expected.png")
-
-        add_ppp.superresolution(Path(input_model), Path(output_model), "png")
+        onnx_opset = 16
+        from packaging import version
+        if version.parse(ort.__version__) >= version.parse("1.14.0"):
+            onnx_opset = 18
+            expected_output_image_path = os.path.join(test_data_dir, "test_superresolution.expected.opset18.png")
+        else:
+            expected_output_image_path = os.path.join(test_data_dir, "test_superresolution.expected.png")
```
edgchen1, line 152: nit: can we move `onnx_opset = 16` here for consistency unless there's a good reason to set it outside the if?

tutorials/superresolution_e2e.py
```diff
@@ -94,6 +94,19 @@ def add_pre_post_processing(output_format: str = "png"):
     add_ppp.superresolution(Path(ONNX_MODEL), Path(ONNX_MODEL_WITH_PRE_POST_PROCESSING), output_format, onnx_opset)
 
 
+def _center_crop(img: Image):
```
edgchen1, line 97: naming nit: how about `_center_crop_to_square`? with "crop" I was expecting dimensions

edgchen1 (dismissed):

onnxruntime_extensions/tools/pre_post_processing/steps/vision.py
```diff
@@ -139,6 +139,11 @@ def _create_graph_for_step(self, graph: onnx.GraphProto, onnx_opset: int):
         output_shape_str = f"YCbCr_ppp_{self.step_num}_h, YCbCr_ppp_{self.step_num}_w"
         assert input_type_str == "uint8"
 
+        split_attr = "axis = -1"
+        if onnx_opset >= 18:
+            # Split now requires the number of outputs to be specified even though that can be easily inferred...
+            split_attr += ", num_outputs = 3"
```
YUNQIUGUO, line 145: thanks for the fix!

interesting to know that Split-18 requires num_outputs as an additional attribute now.

YUNQIUGUO (dismissed):

wejoncy (approved):

edgchen1 (approved):

PR #363, opened 2023-02-17
Fix the current valid binSkim error and warnings.

wenbingl:


CMakeLists.txt
```diff
@@ -94,6 +94,11 @@ if(NOT CC_OPTIMIZE)
 endif()
 
 if (MSVC)
+  check_cxx_compiler_flag(-Qspectre HAS_QSPECTRE)
```
edgchen1, line 97: nit: instead of checking "-option" and then passing "/option" can we use the same style? e.g., consistently using "-option"

edgchen1 (dismissed):

CMakeLists.txt
wenbingl: Not sure if check_cxx_compiler_flag can work with /Qspectre styple.
check_cxx_compile_flag is a cmake function which requires a general flag style. While CMKAE_C_FLAGS are the native compiler so the style is '/flag' for MSVC. so there is a subtle difference.

edgchen1: FWIW, MSVC allows either / or - to specify an option.
https://learn.microsoft.com/en-us/cpp/build/reference/compiler-options?view=msvc-170

CMakeLists.txt
```diff
@@ -94,6 +101,14 @@ if(NOT CC_OPTIMIZE)
 endif()
 
 if (MSVC)
+  if (MSVC)
+    check_cxx_compiler_flag(-sdl HAS_SDL)
+  endif()
```
edgchen1, line 106: can this be moved closer to where HAS_SDL is used?

wenbingl: No, a lot of SDL error actually happened on 3rd libraries and even some code in ort-extensions itself. So right now, I can only turn it on lib by lib, and there would be multiple places to add SDL flag.

CMakeLists.txt
wenbingl: I guess, it may be only used for latest MSVC. For better compatibility, let's still keep it in the current way.

CMakeLists.txt
edgchen1, line 106:
```suggestion
  check_cxx_compiler_flag(-sdl HAS_SDL)
```

edgchen1 (approved):

PR #365, opened 2023-02-22
Initial RobertaTokenizer implementation

sayanshaw24:
### For reference: [HuggingFace Implementation](https://github.com/huggingface/transformers/blob/v4.24.0/src/transformers/models/roberta/tokenization_roberta.py)

**Validation**
- [x] Added sample test cases under `test_robertatok.py` and cross-validated with HuggingFace pretrained RobertaTokenizerFast.
- [x] Extensions building correctly using `python setup.py develop`


operators/tokenizer/bpetokenizer.hpp
```diff
@@ -20,6 +20,7 @@
 #include "nlohmann/json.hpp"
 #include "clip_tokenizer.hpp"
 #include "gpt2_tokenizer.hpp"
+#include "roberta_tokenizer.hpp"
```
wenbingl, line 23: I cannot remember why these header files need to be included here? Can we remove them?

operators/tokenizer/roberta_tokenizer.cc
```diff
@@ -0,0 +1,251 @@
...
+
+#include "nlohmann/json.hpp"
+#include "bpetokenizer.hpp"
+#include "string_tensor.h"
+#include "unicode.h"
+
+std::list<std::pair<int, int>> offset_map;
+
+// Note: the following logic comes from CPython: unicodetype_db.h (_PyUnicode_IsWhitespace)
+bool IsWithinUnicodeSpace(char32_t ch) {
```
wenbingl, line 29: Can we dedup this function?

operators/tokenizer/roberta_tokenizer.cc
```diff
@@ -0,0 +1,251 @@
...
+    auto cur_input = std::move(seg_id.first);
+    // Note: keep ptr to make sure the string_view is valid in the following process
+    const char32_t* ptr = cur_input.c_str();
+    regcmp.Set(ptr);
+
+    int offset = 0;
+    std::list<std::pair<int, int>> offset_mapping;
+
+    // Add offset mapping for BOS token
+    offset_mapping.push_back(std::make_pair(0, 0));
```
wenbingl, line 125: So, offset-mapping is only available for Roberta tokenizer, can we make it be available for other tokenizer as well?

operators/tokenizer/bpetokenizer.hpp
sayanshaw24: We include the header files of the tokenizers that use bpe tokenization here so we can just include this file (bpetokenizer.hpp) for all bpe-based tokenizers. Unfortunately, we cannot remove them and instead just include bpetokenizer.hpp in the header files included here, such as roberta_tokenizer.hpp, because then when building we get 'struct' type redefinition errors for the customop primitives.

operators/tokenizer/roberta_tokenizer.cc
sayanshaw24: Yes, I believe so. I will go ahead and add this for other tokenizers once we have finalized the roberta tokenizer.

operators/tokenizer/roberta_tokenizer.cc
sayanshaw24: I recall I suggested this when we were cleaning up the CLIP tokenizer code - issue was that each time a tokenizer obj that uses this is built, it contains that primitive already so we get an error like the following: `error LNK2005: "bool __cdecl IsWithinUnicodeSpace(char32_t)" (?IsWithinUnicodeSpace@@YA_N_U@Z) already defined in ocos_operators.lib(gpt2_tokenizer.obj)`, so you suggested to just change its name and keep it in the .cc files for the time being.

Given this, what would be the best way to dedup it now?

wenbingl (approved):

PR #367, opened 2023-02-23
Disable C++ exception only for these applicable operators

wenbingl:


CMakeLists.txt
```diff
@@ -393,8 +395,6 @@ endif()
 if(OCOS_ENABLE_MATH)
   target_include_directories(ocos_operators PUBLIC ${dlib_SOURCE_DIR})
   list(APPEND OCOS_COMPILE_DEFINITIONS ENABLE_MATH)
-
-  # The dlib matrix implementation is all in the headers, no library compiling needed.
```
skottmckay, line 397: Is this comment no longer relevant? 

FWIW I find explanatory comments like this to be very helpful.

CMakeLists.txt
```diff
@@ -451,6 +449,19 @@ target_include_directories(ocos_operators PRIVATE ${GSL_INCLUDE_DIR})
 list(APPEND ocos_libraries Microsoft.GSL::GSL)
 
 list(REMOVE_DUPLICATES OCOS_COMPILE_DEFINITIONS)
+target_compile_definitions(noexcep_operators PRIVATE ${OCOS_COMPILE_DEFINITIONS})
+if(NOT OCOS_ENABLE_CPP_EXCEPTIONS)
+  if(MSVC)
+    get_target_property(_target_cxx_flags noexcep_operators COMPILE_OPTIONS)
+    list(REMOVE_ITEM _target_cxx_flags "/EHsc")
+    list(APPEND _target_cxx_flags "/EHs-c-")
```
skottmckay, line 457: out of interest is there an advantage to doing REMOVE + APPEND instead of REPLACE?

CMakeLists.txt
```diff
@@ -539,10 +550,12 @@ foreach(nf ${NO_USE_FILES})
...
 
 # test section
-if(OCOS_ENABLE_CTEST AND OCOS_ENABLE_SELECTED_OPLIST)
-  # currently the tests don't handle operator exclusion cleanly.
-  message(WARNING "Due to usage of OCOS_ENABLE_SELECTED_OPLIST excluding operators the tests are unable to be built and run")
-elseif(OCOS_ENABLE_CTEST AND NOT OCOS_ENABLE_SELECTED_OPLIST)
+if(OCOS_ENABLE_CTEST)
+  if (OCOS_ENABLE_SELECTED_OPLIST)
+    # currently the tests don't handle operator exclusion cleanly.
+    message(FATAL "Due to usage of OCOS_ENABLE_SELECTED_OPLIST excluding operators the tests are unable to be built and run")
```
skottmckay, line 556: nit: message could tell user how to fix by saying "don't set OCOS_ENABLE_CTEST when using OCOS_ENABLE_SELECTED_OPLIST" instead of just saying it's broken.

or alternatively we could print a warning and unset OCOS_ENABLE_CTEST

skottmckay (approved):

CMakeLists.txt
```diff
@@ -212,27 +212,20 @@ if(OCOS_ENABLE_RE2_REGEX)
...
-macro(standardize_output_folder bin_target)
-  set_target_properties(${bin_target} PROPERTIES
-    RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin"
-    LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib"
-    ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib"
-    PDB_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin")
-endmacro()
-
 # ### scan all source files
+set(TARGET_SRC_NOEXCEPTION)
```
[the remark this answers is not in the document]
wenbingl: Yes, I will do it another small PR.

CMakeLists.txt
wenbingl: agree, but this information causes misleading since the one source file actually added in the above.

CMakeLists.txt
wenbingl: REPLACE requires regular-expression, but I am not very familiar it and think it may need some escaping char for the flag.

CMakeLists.txt
wenbingl: OCOS_ENABLE_CTEST is a cmake option and we should not update the value.
Will refine the message in another PR for that.

PR #368, opened 2023-02-23
Add ability to prevent exception propagation if building as part of ORT when ORT has exceptions disabled

skottmckay:
Add ability to prevent exception propagation with top level try/catch hander macros.

If combined build with ORT has exceptions disabled in ORT but ort-ext has an operator that requires exceptions (e.g. SentencePiece), we enable exceptions in ort-ext but prevent them propagating up via try/catch in the entry points that ORT can call. These are:
  - RegisterCustomOps
  - CustomOpBase constructor 
  - CustomOpBase Compute

Removed some places in CustomOpApi that threw if OpKernelInfo* was nullptr but standardizing all kernels to store the OpKernelInfo provided in the ctor. Having one standard CreateKernel meant a lot of overrides could be removed.

Added unit tests
  - need to validate on more platforms and add CI for build where we don't want to allow exceptions to propagate


operators/cv2/imgcodecs/imdecode.hpp
```diff
@@ -33,26 +33,22 @@ struct KernelImageDecoder : BaseKernel {
     // Decode the image
     const std::vector<int32_t> encoded_image_sizes{1, static_cast<int32_t>(encoded_image_data_len)};
     const cv::Mat encoded_image(encoded_image_sizes, CV_8UC1,
-                            const_cast<void*>(static_cast<const void*>(encoded_image_data)));
+                                const_cast<void*>(static_cast<const void*>(encoded_image_data)));
```
skottmckay, line 36: Suggest hiding whitespace diffs as clang-format has run when saving the edits causing lots of diffs. 
Other whitespace diffs are from conforming to 120 char limit but those might not get hidden.

image

CMakeLists.txt
```diff
@@ -158,8 +164,39 @@ if(OCOS_ENABLE_SELECTED_OPLIST)
...
+  if(MSVC)
+    string(REGEX REPLACE "/EHs-c-" "/EHsc" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+  else()
+    string(REPLACE "-fno-exceptions" "-fexceptions" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+    string(REPLACE "-fno-unwind-tables" "" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+    string(REPLACE "-fno-asynchronous-unwind-tables" "" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+  endif()
+
+  # the ort-ext code has to provide a barrier between the exception enabled custom op code and ORT.
+  add_compile_definitions(OCOS_CONTAIN_EXCEPTIONS)
```
skottmckay, line 193: OCOS_PREVENT_EXCEPTION_PROPAGATION may be clearer

pyop/pykernel.h
```diff
@@ -60,10 +60,6 @@ struct PyCustomOpFactory : OrtW::CustomOpBase<PyCustomOpFactory, PyCustomOpKerne
     op_type_ = op;
   }
 
-  void* CreateKernel(const OrtApi& api, const OrtKernelInfo* info) const {
-    return CreateKernelImpl(api, info, opdef_->obj_id, opdef_->attrs);
```
wenbingl, line 64: there are some extra arguments here (opdef_->obj_id...), doesn't this removal cause some issues?

skottmckay: Good catch. Fixed.

CMakeLists.txt
```diff
@@ -158,8 +164,39 @@ if(OCOS_ENABLE_SELECTED_OPLIST)
...
+
+  if (NOT OCOS_ENABLE_CPP_EXCEPTIONS)
+    message(WARNING "Enabling C++ exception support as custom ops included in the build require.")
+    set(OCOS_ENABLE_CPP_EXCEPTIONS ON)
+  endif()
+
+  # undo the flags that ORT has set to disable exceptions.
+  # see https://github.com/microsoft/onnxruntime/blob/b1abb8c656c597bf221bd85682ae3d9e350d9aba/cmake/adjust_global_compile_flags.cmake#L160-L169
+  if(MSVC)
+    string(REGEX REPLACE "/EHs-c-" "/EHsc" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
```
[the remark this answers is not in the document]
skottmckay: I copied from the ORT code [here](https://github.com/microsoft/onnxruntime/blob/abb51ec97599adee62c6c90f950a719ee778b974/cmake/adjust_global_compile_flags.cmake#L161)

Will try REPLACE as that seems simpler.

CMakeLists.txt
```diff
@@ -158,8 +164,39 @@ if(OCOS_ENABLE_SELECTED_OPLIST)
   include(_selectedoplist)
 endif()
 
+set(_OCOS_EXCEPTIONS_REQUIRED OFF)
```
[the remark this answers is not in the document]
skottmckay: I believe the leading '_' is an internal value not expected to be set directly by the user. i.e. isn't in the list of public options specified around line 43.

Do you have an alternative naming suggestion? I was defaulting to OFF mainly so the name of the value set is visible here and letting SetExceptionsRequired update to the correct value. AFAIK cmake doesn't allow returning a value from a function so something like `set(_OCOS_EXCEPTIONS_REQUIRED SetExceptionsRequired())` isn't possible. That said, noexceps_ops just went away so the check will most likely become local to this fine when I merge that in.

includes/exceptions.h
```diff
@@ -0,0 +1,98 @@
...
+#if defined(__ANDROID__)
+  __android_log_print(ANDROID_LOG_ERROR, "onnxruntime-extensions", "Exception in %s line %d: %s", file, line, msg);
+#else
+  std::cerr << "Exception in " << file << " line " << line << ": " << msg << std::endl;
+#endif
+}
+#endif
+
+#ifdef OCOS_NO_EXCEPTIONS
+#define ORTX_CXX_API_THROW(string, code)                                         \
```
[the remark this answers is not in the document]
skottmckay: It would be good to make all the naming/namespaces consistent in a separate PR. We have `OrtW` (not sure what that means), ORTX_, OCOS_ and various other things in places. We should agree on a macro prefix and namespace and update everything to use those. 

I think that's a separate PR though as it will touch a lot of files.

includes/onnxruntime_customop.hpp
```diff
@@ -100,18 +62,44 @@ template <typename TOp, typename TKernel>
...
-    OrtCustomOp::CreateKernel = [](const OrtCustomOp* this_, const OrtApi* api, const OrtKernelInfo* info) { return static_cast<const TOp*>(this_)->CreateKernel(*api, info); };
-    OrtCustomOp::GetName = [](const OrtCustomOp* this_) { return static_cast<const TOp*>(this_)->GetName(); };
-
-    OrtCustomOp::GetExecutionProviderType = [](const OrtCustomOp* this_) { return static_cast<const TOp*>(this_)->GetExecutionProviderType(); };
+    OrtCustomOp::CreateKernel = [](const OrtCustomOp* this_, const OrtApi* api, const OrtKernelInfo* info) {
+      void* result = nullptr;
+      API_IMPL_BEGIN
+      result = static_cast<const TOp*>(this_)->CreateKernel(*api, *info);
+      API_IMPL_END
+      return result;
```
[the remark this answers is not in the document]
skottmckay: Probably. I was contemplating adding new CreateKernel/Compute functions to the ORT CustomOp to allow returning a status when I set it up this way, but as we're just re-throwing or calling abort() it may be fine (if the compiler knows an `abort()` means no return is required).

includes/exceptions.h
```diff
@@ -0,0 +1,98 @@
...
+  std::string message_;
+  OrtErrorCode code_;
+};
+
+#if defined(OCOS_NO_EXCEPTIONS) || defined(OCOS_CONTAIN_EXCEPTIONS)
+inline void PrintFinalMessage(const char* file, int line, const char* msg) {
+#if defined(__ANDROID__)
+  __android_log_print(ANDROID_LOG_ERROR, "onnxruntime-extensions", "Exception in %s line %d: %s", file, line, msg);
+#else
+  std::cerr << "Exception in " << file << " line " << line << ": " << msg << std::endl;
```
wejoncy, line 37: Would fprintf(stderr, ....) be safer?

skottmckay: Typically in C++ using IO streams  is safer and the recommended approach as all the arguments are type checked at compile type.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#slio3-prefer-iostreams-for-io

There's no performance reason here to use printf.

includes/onnxruntime_customop.hpp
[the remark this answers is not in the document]
skottmckay: Yeah nah. It's not smart enough to treat the 'abort()' as a return path so generates a warning (at least with MSVC)

CMakeLists.txt
```diff
@@ -158,12 +164,76 @@ if(OCOS_ENABLE_SELECTED_OPLIST)
...
+    OCOS_ENABLE_BLINGFIRE OR
+    OCOS_ENABLE_SPM_TOKENIZER OR
+    (OCOS_ENABLE_CV2 OR OCOS_ENABLE_OPENCV_CODECS OR OCOS_ENABLE_VISION))
+    set(_OCOS_EXCEPTIONS_REQUIRED ON)
+endif()
+
+# Special case an embedded build with ORT exceptions disabled but custom ops that require exceptions.
+# Allow using an override so we can do a direct build of ort-ext in a CI without having to embed it in an ORT build.
+set(_OCOS_PREVENT_EXCEPTION_PROPAGATION OFF)
+if (_OCOS_PREVENT_EXCEPTION_PROPAGATION_OVERRIDE)
```
skottmckay, line 179: This allows building with _OCOS_PREVENT_EXCEPTION_PROPAGATION_OVERRIDE=ON to simulate an embedded build with exceptions disabled so that the test_exceptions.cc tests can be run. Will look at setting up a CI with this tomorrow unless there's a reason to prefer a full embedded build in the CI.

includes/exceptions.h
skottmckay: @wenbingl Please let me know how I can re-create the warning from using std::cerr. Can change to printf if we can't fix that warning.

CMakeLists.txt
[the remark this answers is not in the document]
skottmckay: Things are inlined now so this shouldn't be an issue

includes/exceptions.h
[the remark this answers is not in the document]
edgchen1: ok, sounds good

CMakeLists.txt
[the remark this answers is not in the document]
edgchen1: > Do you have an alternative naming suggestion?

not really, just didn't quite understand the naming conventions yet. now I'm also wondering when the `[_]OCOS` prefix should be used.

> AFAIK cmake doesn't allow returning a value from a function

FWIW, we could have function output variables by providing an output variable name parameter, e.g., SetExceptionsRequired(_OCOS_EXCEPTIONS_REQUIRED).

includes/ocos.h
```diff
@@ -18,28 +18,29 @@ constexpr const char* c_OpDomain = "ai.onnx.contrib";
...
 
   template <class T>
-  bool TryToGetAttribute(const char* name, T& value);
+  bool TryToGetAttribute(const char* name, T& value) const noexcept;
 
   template <class T>
-  T TryToGetAttributeWithDefault(const char* name, T default_value) {
+  T TryToGetAttributeWithDefault(const char* name, T default_value) const noexcept {
     T& result = default_value;
     TryToGetAttribute(name, result);
```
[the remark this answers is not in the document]
skottmckay: I agree. I don't think modifying the default value be expected. I would lean towards making it `const T default_value` to make that more explicit as well.

CMakeLists.txt
```diff
@@ -158,8 +164,80 @@ if(OCOS_ENABLE_SELECTED_OPLIST)
   include(_selectedoplist)
 endif()
 
+set(_OCOS_EXCEPTIONS_REQUIRED OFF)
```
wenbingl, line 167: Given we only disable exception flags for the non-exception operation, do we still need _OCOS_EXCEPTIONS_REQUIRED since this flag requires the maintenance of the build option list.

CMakeLists.txt
```diff
@@ -158,8 +164,80 @@ if(OCOS_ENABLE_SELECTED_OPLIST)
...
+    string(REPLACE "-fno-exceptions" "-fexceptions" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+    string(REPLACE "-fno-unwind-tables" "" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+    string(REPLACE "-fno-asynchronous-unwind-tables" "" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+  endif()
+
+  # the ort-ext code has to provide a barrier between the exception enabled custom op code and ORT.
+  add_compile_definitions(OCOS_PREVENT_EXCEPTION_PROPAGATION)
+endif()
+
 if(NOT OCOS_ENABLE_CPP_EXCEPTIONS)
```
wenbingl, line 209: Do we still need to report the error here? since C++ exceptions can be partially disabled for some operators now.

wenbingl:
@skottmckay , do you still need to refine this PR? 
I will check it later today and approve and merge it if there is no major issue, then cut a release branch.

CMakeLists.txt
skottmckay: Possibly not, but not sure about one aspects of the setup though.

We're building some operators with and some without exceptions.
We have ORTX_CXX_API_THROW to provide a consistent way to throw, but that is dependent on OCOS_NO_EXCEPTIONS which is not operator specific. 
Operators have no way to report errors in construction or compute with the current ORT custom op API having no ability to return a status, so AFAIK they MUST throw (or log+abort if execeptions disabled) for any error.

What is the expected operator implementation for something that compiles without exceptions? The string operators currently use ORTX_CXX_API_THROW but I think are in the operators we build with exceptions disabled so not sure how that is working.

skottmckay:
> @skottmckay , do you still need to refine this PR? I will check it later today and approve and merge it if there is no major issue, then cut a release branch. have a good weekend!

Maybe some minor things once I test more builds. I want to add something to the CIs but that doesn't necessarily have to block the release if manual testing is OK for checkin. Can we look to cut the release branch first Monday so I can do more testing/validation over the w/e and on my Monday.

CMakeLists.txt
skottmckay: Is the current setup if exceptions are disabled as follows:
- set OCOS_NO_EXCEPTIONS 
  - which causes all calls to ORTX_CXX_API_THROW to log+abort
- compile the noexcep_operators operators with adjusted flags to disable exceptions
  - this code can use ORTX_CXX_API_THROW which will result in log+abort
- compile other operations with exceptions enabled
  -  3rd party code can call `throw`

For the operators that throw in a build with exception disabled what is the expected behavior? Propagate exception up (no call to `abort`), or something like OCOS_API_IMPL_END should catch the exception and log+abort so all operators have consistent behavior?

includes/exceptions.h
wenbingl: Some warning like this,
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.35.32215\include\ostream(360,1): warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc

But, I just figure out defining a _HAS_EXCEPTIONS=0 can remove the warnings from MSVC, but not sure if it is officially supported and how about other compilers,

CMakeLists.txt
wenbingl: My initial thinking is simple:
If OCOS_ENABLE_CPPEXCEPTION is OFF, then all code in ort-extensions will NOT throw any exception since ORTX-THROW will abort immediately if something unexpected is happening.
But try-catch macro in API function is effective and it actually just only catch the exception from 3rd party libraries.

There is one thing missing here is that BaseKernel should have an exception-free superclass, which will be the base class for non-exception operators. So if only non-exception operators are compiled, there actually no try-catch code in whole compiling source code, I guess the binary size is still minimal in that case.

skottmckay: Updated to also log+abort in OCOS_API_IMPL_END if exceptions are disabled.

Not sure it's worth the complexity of a BaseKernel layer without testing to see if the binary size saving justifies it and there a use case that requires it.

skottmckay:
> > @skottmckay , do you still need to refine this PR? I will check it later today and approve and merge it if there is no major issue, then cut a release branch. have a good weekend!
> 
> Maybe some minor things once I test more builds. I want to add something to the CIs but that doesn't necessarily have to block the release if manual testing is OK for checkin. Can we look to cut the release branch first Monday so I can do more testing/validation over the w/e and on my Monday.

PR is ready to be checked in.

wenbingl (approved):

PR #369, opened 2023-02-24
Ignore inputs that are missing from the vocab in CLIP tokenizer

matheusgomes28:
Hi all,

I came around the issue of unrecognised words in the vocab map when using the ClipTokenizer from OnnxExtensions. For example, if an input contains `\n` or other missing inputs in the vocab, the call to `bbpe_tokenizer_->GetEncoding(boundary + "</w>")` would cause a crash as we're dereferencing an end iterator there.

I added some code to ignore the input if it's not present in the vocab file. Let me know if that looks ok, or if it will cause repercussions elsewhere. Happy to discuss alternatives to fixing this crash.


matheusgomes28:

wenbingl:
@matheusgomes28 , what's the current behavior of Huggingface tokenizer implementation? it is also ignored. Just want to confirm that.

matheusgomes28:
> @matheusgomes28 , what's the current behavior of Huggingface tokenizer implementation? it is also ignored. Just want to confirm that.

I will check what the behaviour is. It doesn't crash on the Python huggyface implementation, so I guess they either ignore it or add some default token instead. Again, I'm not sure what the best default token would be (as it depends on the model's vocab), so IMO the best option would be to ignore invalid inputs.

Anyway, I'll update this with the Huggy Face behaviour tomorrow :)

wenbingl:
> > @matheusgomes28 , what's the current behavior of Huggingface tokenizer implementation? it is also ignored. Just want to confirm that.
> 
> @wenbingl I will check what the behaviour is. It doesn't crash on the Python huggyface implementation, so I guess they either ignore it or add some default token instead. Again, I'm not sure what the best default token would be (as it depends on the model's vocab), so IMO the best option would be to ignore invalid inputs.
> 
> Anyway, I'll update this with the Huggy Face behaviour tomorrow :)

most tokenizer supports the unknown-id, something like <UNK>. It might be a good candidate for the word/char missing in the  vocab.

sayanshaw24:
Closing and reopening here with updates due to access issues for merge conflicts: https://github.com/microsoft/onnxruntime-extensions/pull/462.

PR #372, opened 2023-02-28
Fix the build breaks the release pipeline and some C++ warnings

wenbingl:


operators/tokenizer/bpetokenizer.hpp
```diff
@@ -1,11 +1,12 @@
 // Licensed under the MIT License.
 // Partial code comes from other Microsoft employee.
 #pragma once
+#include "ocos.h"
 #include "ustring.h"
 
 #include <regex>
+#include <list>
```
sayanshaw24, line 8: Perfect - was going to say this is what caused the ci failures in the previous run, should be fine now.

sayanshaw24 (dismissed):
Thanks for adding this.

PR #375, opened 2023-03-02
port the fixing from the release branch

wenbingl:
The fixing solves the problem of the removal build-id from release build, while it should be kept in nightly build.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ def load_vsdevcmd():
     if os.environ.get(VSINSTALLDIR_NAME) is None:
         stdout, _ = subprocess.Popen([
             'powershell', ' -noprofile', '-executionpolicy', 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -150,15 +150,22 @@ def read_version():
...
+    # special handling for Onebranch building
+    if os.getenv('BUILD_SOURCEBRANCHNAME', "").startswith('rel-'):
+        return version_str
+
     # is it a dev build or release?
+    rel_br, cid = read_git_refs() if os.path.isdir(
+        os.path.join(TOP_DIR, '.git')) else (True, None)
+
+    if rel_br:
+        return version_str
```
sayanshaw24, line 162: So to confirm, in addition to build_id, we are also removing cid from the release branch correct?

wenbingl: Yes, if any release branch information retrieved, we will remove the build id or other information from the version string.

sayanshaw24 (approved):

PR #377, opened 2023-03-06
Added offset mapping support for CLIPTokenizer

sayanshaw24:


operators/tokenizer/bpetokenizer.hpp
```diff
@@ -261,6 +261,29 @@ class VocabData {
     return it->second;
   }
 
+  const std::string DecodeEncoding(int value) {
+    for (auto& it : vocab_map_) {
```
wenbingl, line 265: this looks very inefficient.

test/test_cliptok.py
```diff
@@ -1,10 +1,11 @@
 import unittest
 import numpy as np
+import numpy.lib.recfunctions as nlr
 import onnxruntime as _ort
 
 from pathlib import Path
 from onnx import helper, onnx_pb as onnx_proto
-from transformers import CLIPTokenizer
+from transformers import CLIPTokenizerFast
```
wenbingl, line 8: why switch to the *Fast version?
nonFast version seems to be more formal

sayanshaw24: normal version does not support offset mapping which we need for test validation

operators/tokenizer/bpetokenizer.hpp
sayanshaw24: correct - best solution to this is keep a 1:1 mapping (vocab string, id) from K->V and V->K using 2 maps, but that would be storage inefficient. wanted to check in with you before adding that to confirm we want to prioritize compute?

wenbingl: you may log the used token for one utterance so that you can build a small map based on the used token only.

sayanshaw24: so this was the original goal but the issue is after we perform byte-level bpe we are only left with input ids (so we cannot get the tokens to build the map without decoding).

For example, with input "Hey Cortana", we do not know that the tokens are ["hey", "cor", "tana"] (cannot know that cortana was split as cor-tana) until we perform bbpe and decode the input ids [2189, 851, 21396]. So then we have to find the token from the input id to get the offset mapping. Does that make sense (correct me if I am wrong)?

test/test_cliptok.py
wenbingl: If there is no official support in Normal version, and if there is no good solution for reserve vocabulary looking up, we may not push for this feature.

sayanshaw24: sounds good, we can keep offset mapping on a case-by-case implementation based on customer demand in that case, like with RobertaTokenizer.

test/test_cliptok.py
```diff
@@ -81,15 +90,5 @@ def test_tokenizer(self):
         self._run_tokenizer(["9 8 7 - 6 5 4 - 3 2 1 0"])
         self._run_tokenizer(["One Microsoft Way, Redmond, WA"])
 
-    def test_converter(self):
```
sayanshaw24, line 84: Temporarily removed converter check as model from PyOrtFunction is outdated therefore it would throw errors (model output is 2 instead of expected 3) and this new CLIPTokenizer with offset mapping output needs to be checked in first.

Will add back and update check with offset mapping in different PR once this PR is checked in.

wenbingl: use unittest.skip, instead of deletion

sayanshaw24: Got it, forgot about this trick, thanks! Will use this in the future.

PR #380, opened 2023-03-15
Pre and Post processing example for openAI Whisper model

wenbingl:


onnxruntime_extensions/_cuops.py
```diff
@@ -372,6 +372,24 @@ def get_outputs(cls):
         ]
 
 
+class StftNorm(CustomOp):
+    @classmethod
+    def get_inputs(cls):
+        return [
+            cls.io_def('pcm_wave', onnx_proto.TensorProto.FLOAT, [None]),
```
guschmue, line 379: onnx has a definition for stft here:
https://github.com/onnx/onnx/blob/main/docs/Operators.md#STFT

Can we use this signature and names and not invent new ones ?

guschmue (changes requested):

onnxruntime_extensions/_cuops.py
wenbingl: I don't see any benefit of why the signature has to be same. Even the signature are same, the op domain and the exporting script are still different, you can expert there is no change switch.

since ONNX already has STFT operator, it doesn't make sense to duplicate another one in ort-extensions. The reason of this op here is efficiency. Fusing with NORM operation can increase efficiency and it also has a clear physics definition magnitudes of the frequency response.

operators/math/dlib/stft_norm.hpp
```diff
@@ -0,0 +1,124 @@
...
+      }
+    }
+  }
+
+ private:
+  bool return_magnitude_;
+  int64_t onesided_;
+};
+
+struct CustomOpStft : OrtW::CustomOpBase<CustomOpStft, KernelStft> {
```
wenbingl, line 83: TODO: need a unit test for this op.

onnxruntime_extensions/_cuops.py
guschmue: I can run this thing in 300ms in javascript so efficiency is not the problem (reads as: if you do it in 250ms or 300ms doesn't matter if the total inference time is 10sec).
What I mean is there is a STFT in onnx defined - don't invent your own, it is hard to get such thing removed later.
You can either fix the existing STFT or temporarily use a custom op to work around it but I think it should be temporary and for that the signatures should be the same.

wenbingl: ONNX STFT signature also added now, https://github.com/microsoft/onnxruntime-extensions/blob/39a32f68c7163a24581372e9be80a05478302504/test/test_audio.py#L20

but there is no end-to-end example for that now since I am not sure if there is any reliable solution to export it, or you share one if you have it?

In the long run, ONNX STFT op has to be updated or to be deprecated to be more friendly torch exporter and remove the complex data type workaround.

operators/math/math.cc
```diff
@@ -12,6 +13,8 @@ FxLoadCustomOpFactory LoadCustomOpClasses_Math =
                         CustomOpNegPos,
 #ifdef ENABLE_DLIB
                         CustomOpInverse,
+                        CustomOpStft,
```
guschmue, line 16: not sure if for ort-web or mobile we a single define ENABLE_DLIB is sufficient - I think we only want Stft

operators/math/math.cc
```diff
@@ -12,6 +13,8 @@ FxLoadCustomOpFactory LoadCustomOpClasses_Math =
...
+                        CustomOpStftNorm,
 #endif
                         CustomOpSegmentExtraction,
```
guschmue, line 19: for sure we don't want SegmentExtraction & co when including Stft

tutorials/whisper_e2e.py
```diff
@@ -0,0 +1,115 @@
...
+    def __init__(self):
+        super().__init__()
+        self.window = torch.hann_window(N_FFT)
+        self.mel_filters = torch.from_numpy(librosa.filters.mel(sr=SAMPLE_RATE, n_fft=N_FFT, n_mels=N_MELS))
+
+    def forward(self, audio_pcm: torch.Tensor):
+        pad_len = N_SAMPLES - audio_pcm.shape[0]
+        audio_pcm = torch.nn.functional.pad(audio_pcm, (0, pad_len), mode='constant', value=0)
+        audio_pcm = audio_pcm.unsqueeze(0)
+        stft_norm = CustomOpStftNorm.apply(audio_pcm, N_FFT, HOP_LENGTH, self.window)
```
guschmue, line 53: again, I have trouble understanding StftNorm when onnx has a Stft.

guschmue (changes requested):

onnxruntime_extensions/_cuops.py
guschmue: yes, I can share a script.
I have trouble with StftNorm in the presents of Stft in onnx but I would have far less trouble if StftNorm 
would be log_mel_spectrum().

skottmckay: > I don't see any benefit of why the signature has to be same. 

Naive question - what is the benefit of making it different? 

If it's the same (attributes and inputs/outputs match including shapes) the custom op can override the ORT implementation of an ONNX kernel to provide a different implementation. You'd also get the ONNX shape inferencing which will help with constant folding and allocation planning. Looked like StftNorm produces a different output shape though so I assume it's not a drop in replacement for ONNX STFT.

cmake/externals/dlib.cmake
```diff
@@ -1,6 +1,8 @@
 FetchContent_Declare(dlib
     GIT_REPOSITORY https://github.com/davisking/dlib.git
-    GIT_TAG        v19.22
+    # there is non an official tag which supports STFT,
+    # choose a relatively stable commit id for that.
+    GIT_TAG        a12824d42584e292ecb3bad05c4b32c2015a7b89
```
skottmckay, line 5: This might be problematic when we have to switch to using a URL/URL_HASH based setup as we need a zip from a release.

wenbingl: A not-well-known feature is github can also produce zip for any commit id, which is the way how ort integrate onnx and ort-extensions now.

onnxruntime_extensions/_cuops.py
wenbingl: followed up with some internal discussion.

operators/math/dlib/stft_norm.hpp
```diff
@@ -0,0 +1,125 @@
...
+  const char* op_name_ = "STFT";
+};
+
+struct CustomOpStftNorm : CustomOpStft {
+ public:
+  CustomOpStftNorm() {
+    with_norm_ = true;
+    op_name_ = "StftNorm";
+  }
+};
```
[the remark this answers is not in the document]
wenbingl: followed up with some internal discussion.

natke (commented):
Should this PR be named something like Support pre and post processing for the whisper model? Since there is an end to end script.

guschmue (approved):

test/test_audio.py
pranavsharma, line 1: Missing copyright/licensing header for this and other Python files in this PR.

wenbingl: Thanks for the reminder, I will add them into in another PR.
BTW, Is there any file in ORT that can be the template for copyright info? I saw there were some different styles.

pranavsharma: https://docs.opensource.microsoft.com/releasing/general-guidance/copyright-headers/

PR #382, opened 2023-03-21
built-in bounding box op

wejoncy:
- Built-in Bounding-box drawing support
- LetterBox padding/Crop support
- No Text drawing


operators/vision/draw_bounding_box.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+#pragma once
+
+#include "ocos.h"
+#include "onnxruntime_c_api.h"
+#include "string_utils.h"
+
+#include <cstdint>
+
+namespace ort_extensions {
```
wenbingl, line 12: may we unify the root namespace name as OrtX, more consistent with Ort style?
@skottmckay

skottmckay: We should definitely unify.

ORT follows the Google C++ Style Guide so namespace should be lowercase with '_' separator, and the name should be clear and not unnecessarily abbreviated.

https://google.github.io/styleguide/cppguide.html#Namespace_Names
https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

`ort_extensions` seems to fit those rules. I think the most you could shorten that is to `ort_ext` but that's possibly getting into unnecessarily abbreviated territory.

operators/vision/draw_bounding_box.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include "ocos.h"
+#include "string_utils.h"
+
+#include <cstdint>
+
+namespace ort_extensions {
+struct DrawBoundingBox : BaseKernel {
+  DrawBoundingBox(const OrtApi& api, const OrtKernelInfo& info) : BaseKernel(api, info) {
+    thickness_ = TryToGetAttributeWithDefault<int64_t>("thickness", 4);
+    num_classes_ = static_cast<int32_t>(TryToGetAttributeWithDefault<int64_t>("num_classes", 10));
```
skottmckay, line 14: I expected we'd configure the number of colours to use. If doing output by score, that value is the maximum number of boxes we draw. If the output is by class, that value is the maximum number of classes we can draw results for.

operators/vision/draw_bounding_box.hpp
```diff
@@ -0,0 +1,68 @@
...
+    auto need_sort = TryToGetAttributeWithDefault<int64_t>("sort_by_score", 0);
```
skottmckay, line 15: If we don't sort by score, we're not picking the top x classes/results which seems like an unexpected way to output results.

operators/vision/draw_bounding_box.cc
```diff
@@ -0,0 +1,191 @@
...
+              const uint8_t* color, int64_t thickness) {
+  // -------(1)------
+  // |              |
+  //(2)            (4)
+  // |              |
+  // -------(3)------
+  int64_t x_start = static_cast<int64_t>(box[0]);
+  int64_t y_start = static_cast<int64_t>(box[1]);
+  int64_t x_end = static_cast<int64_t>(box[2]);
+  int64_t y_end = static_cast<int64_t>(box[3]);
```
skottmckay, line 79: Should these use round instead of truncating via static_cast?

operators/vision/draw_bounding_box.cc
```diff
@@ -0,0 +1,191 @@
...
+  x_end += half_thickness;
+  y_end += half_thickness;
+
+  x_start = std::max(x_start, 0L);
+  y_start = std::max(y_start, 0L);
+  x_end = std::min(x_end, image.height - 1);
+  y_end = std::min(y_end, image.width - 1);
+
+  if (x_end - x_start < thickness || y_end - y_start < thickness) {
+    // "invalid thickness, The box is too small to draw.";
```
skottmckay, line 93: Silent failures are typically bad and hard to debug. Should this throw so the user can fix how the operator is being used?

operators/vision/draw_bounding_box.cc
```diff
@@ -0,0 +1,191 @@
...
+  // |              |
+  //(2)            (4)
+  // |              |
+  // -------(3)------
+  int64_t x_start = static_cast<int64_t>(box[0]);
+  int64_t y_start = static_cast<int64_t>(box[1]);
+  int64_t x_end = static_cast<int64_t>(box[2]);
+  int64_t y_end = static_cast<int64_t>(box[3]);
+
+  auto half_thickness = (thickness - 1) / 2;
```
skottmckay, line 81: Why -1?

If thickness is 0 or 1 this gives a value of 0. Is that intended?

Would we avoid potential invalid data access and simplify things if we always drew the lines on the inside of the box vs. trying to center the lines on the box?

onnxruntime_extensions/tools/pre_post_processing/steps/vision.py
```diff
@@ -604,3 +606,124 @@ def __init__(self, has_batch_dim: bool = False, name: Optional[str] = None):
         """
         perms = [0, 3, 1, 2] if has_batch_dim else [2, 0, 1]
         super().__init__(perms, name)
+
+
+class DrawBoundingBox(Step):
+    """
+    Draw boxes on BGR image at given position.
+    Input shape: {height, width, 3<BGR>}
+    Output shape: {height, width, 3<BGR>}
```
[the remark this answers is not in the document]
wejoncy: Sure.

onnxruntime_extensions/tools/pre_post_processing/steps/vision.py
```diff
@@ -604,3 +606,124 @@ def __init__(self, has_batch_dim: bool = False, name: Optional[str] = None):
...
+    """
+
+    def __init__(self, thickness: int = 4, num_classes: int = 10, sort_by_score=False, colour_by_classes=False, name: Optional[str] = None):
```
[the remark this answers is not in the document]
wejoncy: Do we need to support different format? like `xyxy` or `xywh`

operators/vision/draw_bounding_box.hpp
wejoncy: Yes, we have 10 of predefined colors to use.
For now, I didn't use `num_classes_`  too much.

operators/vision/draw_bounding_box.hpp
wejoncy: If we have a few class, so it doesn't make sense to sort classes right?  For example, the first class has 7 object, and the second has 4 object, we can't say the first class is priority over the second.

onnxruntime_extensions/tools/pre_post_processing/steps/vision.py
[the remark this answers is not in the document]
skottmckay: I'd say take a look at some relevant models in pytorch and tensorflow and cover the format/s they use. Probably the latest yolo versions, ssd mobilenet from torch and tf, Faster R-CNN, maybe [ultraface](https://github.com/Linzaer/Ultra-Light-Fast-Generic-Face-Detector-1MB) as we use it in the C# examples would be a good starting set. We can add others later if/when needed.

operators/vision/draw_bounding_box.cc
wejoncy: I would expect 'thickness >=1'.  Not sure how I should hanlde this if thickness==0, to return directly or failed?

The reason of "thickness - 1" :
if we want to draw a point in x=2, and we have thickness=2,  then we can draw  the point at {1,2} or {2,3}.  Not sure which one is the best.
  \|-----------------------------------\|
  \|-==--------------------------------\|
  \|--==-------------------------------\|

I don't think drawing the lines on the inside of the box would simplify this and avoid invalid data access. Because we have  to process the boundary whenever. trying to center the lines on the box will make it looks better instead.

operators/vision/draw_bounding_box.cc
wejoncy: Personally, I would we can draw as much as possible boxes and one box failure shouldn't hold down the whole procedure.
But I agree with that silent failures is hard to debug.

operators/vision/draw_bounding_box.cc
```diff
@@ -0,0 +1,191 @@
...
+  }
+}
+
+void draw_line_vertical(ImageView& image, int64_t point_offset, int64_t line_length, const uint8_t* color,
+                        int64_t thickness) {
+  auto* point_start = image.data + point_offset;
+  auto* point_end = point_start + line_length * image.width * image.channels;
+  for (; point_start < point_end; point_start += image.width * image.channels) {
+    for (int64_t i = 0; i < thickness; ++i) {
+      // no boundary check
```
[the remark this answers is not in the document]
wejoncy: Sure. Done.

operators/vision/draw_bounding_box.hpp
wejoncy: Got it. Will do the sort before drawing boxes.

operators/vision/draw_bounding_box.cc
skottmckay: Can we just set the thickness to x_end - s_start or y_end - y_start? The line is so thick it will fill the box, so rather than overfilling the box, just fill it. Hopefully that equates to we draw the box, there's no error, and because the thickness or box is insane the user just gets a solid block.

wejoncy: Yes. It's a acceptable compromise

operators/vision/draw_bounding_box.cc
skottmckay: thickness == 0 is throw in ctor I would expect. It's invalid usage of the op.

Regarding invalid data access, the std::min/std::max adjustment should prevent that. Sorry - overlooked that part.

I'd favour pushing the box slightly out - so for thickness 2 and a box from 2,2, to 8,8 I'd draw on pixels 1 and 2 for left and top and pixels 8 and 9 for bottom and right. 

Might be clearer to call it `auto offset = thickness / 2;` as well given that value doesn't truly represent half the thickness.

The user has no reference point to compare how we draw the boxes to, so 'looks better' seems like it would be very hard to quantify/prove.

operators/vision/draw_bounding_box.hpp
wenbingl: any story about this Ort in this C++ api?
https://github.com/microsoft/onnxruntime/blob/2ee822d4831d2043e3d588d601264ed96cd72bf3/include/onnxruntime/core/session/onnxruntime_cxx_api.h#L45, 
And I remember I also saw this pattern in ort codebase for multiple times and take it as de facto standard

ort-ext is not good since it neither be short enough for convenience, nor precise of full name.

operators/vision/draw_bounding_box.cc
wejoncy: >The user has no reference point to compare how we draw the boxes to, so 'looks better' seems like it would be very hard to quantify/prove
hmm, It's ture.

onnxruntime_extensions/tools/pre_post_processing/steps/vision.py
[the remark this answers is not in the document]
wejoncy: Yolov8 used xywh, and yolov5 used xyxy.
I would support both.

operators/vision/draw_bounding_box.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include "ocos.h"
+#include "string_utils.h"
+
+namespace ort_extensions {
+
+enum class BoundingBoxFormat {
+  XYWH,
+  XYXY,
+  Center_XYWH,
+};
```
skottmckay, line 15: Can we link to some documentation both here and in the Step explaining exactly what these formats are so that a new user who is not familiar with bounding boxes can more easily understand? The Keras documentation would be fine if that's the best we can find.

operators/vision/draw_bounding_box.cc
```diff
@@ -0,0 +1,248 @@
...
+  gsl::span<uint8_t> data;
+  int64_t height;
+  int64_t width;
+  int64_t channels;
+};
+
+// To represent Boxes, [num of boxes, box_info]
+class BoxArray {
+ private:
+  void SortBoxesByScore(const float* data) {
```
skottmckay, line 41: Use gsl::span instead of raw pointers.

operators/vision/draw_bounding_box.cc
```diff
@@ -0,0 +1,248 @@
...
+                return first[4] > second[4];
+              });
+  }
+
+  [[nodiscard]] int64_t ShapeAtDim(int64_t dim) const {
+    return shape_[dim];
+  }
+
+ public:
+  BoxArray(const std::vector<int64_t>& shape, const float* data, BoundingBoxFormat bbox_mode) : shape_(shape), bbox_mode_(bbox_mode) {
```
skottmckay, line 58: nit: line length 

Most IDEs have a plugin or built in capacity to display a vertical line at 120 chars. If you set that up it's very clear when a line is too long. e.g. VS has the Editor Guidelines plugin and VS Code has the editor.rulers setting.

operators/vision/draw_bounding_box.cc
```diff
@@ -0,0 +1,248 @@
...
+ private:
+  void SortBoxesByScore(const float* data) {
+    boxes_by_score_.resize(NumBoxes());
+    for (size_t i = 0; i < boxes_by_score_.size(); ++i) {
+      boxes_by_score_[i] = gsl::make_span(data + i * ShapeAtDim(1), ShapeAtDim(1));
+    }
+
+    std::sort(boxes_by_score_.begin(), boxes_by_score_.end(),
+              [](const gsl::span<const float>& first, const gsl::span<const float>& second) {
+                return first[4] > second[4];
```
skottmckay, line 49: nit: ShapeAtDim() doesn't seem to provide much value given it's a private method and the code could just be shape_[dim] without losing any readability.

nit: avoid magic numbers. define a constant for the index of the score in the data so '4' has some context.

operators/vision/draw_bounding_box.cc
```diff
@@ -0,0 +1,248 @@
...
+  int64_t height;
+  int64_t width;
+  int64_t channels;
+};
+
+// To represent Boxes, [num of boxes, box_info]
+class BoxArray {
+ private:
+  void SortBoxesByScore(const float* data) {
+    boxes_by_score_.resize(NumBoxes());
```
skottmckay, line 42: nit: as you're going to populate each entry it's more efficient to use `reserve`.

operators/vision/draw_bounding_box.cc
```diff
@@ -0,0 +1,248 @@
...
+  } else if (bbox_mode == BoundingBoxFormat::XYWH) {
+    point_x1 = box[0];
+    point_x2 = box[1];
+    point_x3 = box[0] + box[2];
+    point_x4 = box[1] + box[3];
+  }
+  int64_t x_start = static_cast<int64_t>(std::rintf(point_x1));
+  int64_t y_start = static_cast<int64_t>(std::rintf(point_x2));
+  int64_t x_end = static_cast<int64_t>(std::rintf(point_x3));
+  int64_t y_end = static_cast<int64_t>(std::rintf(point_x4));
```
skottmckay, line 140: Why rintf? Doesn't that make it less deterministic than `round` as the current rounding mode is used, and it could be harder to debug an issue?

operators/vision/draw_bounding_box.cc
```diff
@@ -0,0 +1,248 @@
...
+  thickness = std::clamp<int64_t>(thickness, 1, std::min(x_end - x_start, y_end - y_start));
+  if (x_end - x_start < thickness || y_end - y_start < thickness) {
+    // "invalid box, It's invalid to draw a point to represent a box.";
+    return;
+  }
+
+  if (x_start + thickness >= image.height || y_start + thickness >= image.width) {
+    // "invalid box, It's invalid to draw a point to represent a box.";
+    return;
+  }
```
skottmckay, line 162: These should throw. A silent failure is going to lead to users creating a GH Issues saying they asked for X boxes and only Y were drawn and it will be time consuming to figure out why.

operators/vision/draw_bounding_box.cc
```diff
@@ -0,0 +1,248 @@
...
+  // line  (4) --|
+  DrawLineInVertical(image, x_start, y_end - thickness, x_end - x_start, color, thickness);
+}
+
+void DrawBoxesForNumClasses(ImageView& image, const BoxArray& boxes, int64_t thickness) {
+  std::unordered_map<float, size_t> color_used;
+
+  for (size_t i = 0; i < boxes.NumBoxes(); ++i) {
+    const auto box = boxes.GetBox(i);
+    if (color_used.find(box[5]) == color_used.end()) {
```
skottmckay, line 182: we should have a constant providing context instead of directly using the value '5'

operators/vision/draw_bounding_box.cc
```diff
@@ -0,0 +1,248 @@
...
+  // x,y, x/w y/h, score, class
+  if (dimensions_box.size() != 2 || dimensions_box[1] != 6) {
+    ORTX_CXX_API_THROW("[DrawBoundingBoxes] requires rank 2 input and the last dim should be 6.", ORT_INVALID_ARGUMENT);
+  }
+  BoxArray boxes(dimensions_box, ort_.GetTensorData<float>(input_box), bbox_mode_);
+  int64_t image_size = dimensions_bgr[0] * dimensions_bgr[1] * dimensions_bgr[2];
+
+  // Setup output & copy to destination
+  // can we reuse the input buffer?
+  const std::vector<int64_t> output_dims(dimensions_bgr);
```
skottmckay, line 225: If we're not changing any values why do we copy the dims?

operators/vision/draw_bounding_box.cc
```diff
@@ -0,0 +1,248 @@
...
+  OrtValue* output_value = ort_.KernelContext_GetOutput(context, 0,
+                                                        output_dims.data(),
+                                                        output_dims.size());
+
+  auto* output_data = ort_.GetTensorMutableData<uint8_t>(output_value);
+  const auto* input_data = ort_.GetTensorData<uint8_t>(input_bgr);
+  // TODO: support batch in python side.
+  // So we are hard-coding batch size to 1 for now.
```
skottmckay, line 233: What's the use case where we'd be doing batched inferences?

test/test_tools_add_pre_post_processing_to_model.py
```diff
@@ -320,6 +320,28 @@ def test_debug_step(self):
 
         self.assertEqual(len(new_model.graph.output), len(input_model.graph.output) + len(post_processing))
 
+    def test_make_border_and_draw_box(self):
```
skottmckay, line 323: This a great start but I think there's a lot more test coverage needed.

Some test cases that comes to mind
- boxes that share a border with the image (need to test top, bottom, left and right
- box that goes off image (e.g. user did something wrong but we should still draw the part of the box that is valid for the image)
- colour by classes with more classes than colours
- colour by score with more results than colours
- overlapping boxes
  - how do the colours look? should we for example figure start drawing the lowest scoring results once we determine the full set of boxes to draw so that the lines for the best results are done last and are on top?

operators/vision/draw_bounding_box.cc
wejoncy: Actually, I am confused why all pre-post steps do only support one batch.  It limits the ability to run multi-batches in mobile.

Does it have any motivations behind?

onnxruntime_extensions/tools/pre_post_processing/steps/vision.py
```diff
@@ -604,3 +616,148 @@ def __init__(self, has_batch_dim: bool = False, name: Optional[str] = None):
...
+                    "XYXY" (xmin ymin xmax ymax)  All values in the XYXY format should be absolute pixel values.
+                    "XYWH" (xmin ymin width height) 
+                    "CENTER_XYWH" (x_center, y_center, width, height) 
+                                  All values in the CENTER_XYWH format should be absolute pixel values.
+
+
+            thickness: Thickness of the box edge
+            num_colours: Number of colours to use
+                         We have 10 of predefined colours, if `num_colours` is greater than 10, we will use 
+                         the loop-around strategy to get the colours.
```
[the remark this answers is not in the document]
wejoncy: Fixed.

onnxruntime_extensions/tools/pre_post_processing/steps/vision.py
```diff
@@ -604,3 +616,148 @@ def __init__(self, has_batch_dim: bool = False, name: Optional[str] = None):
...
+                i64_0 = Constant <value = int64[1] {{0}}>()
+                const_val = Constant <value = uint8[1] {{{self.fill_value_}}}> ()
+                image_shape = Shape ({self.input_names[0]})
+                h,w,c = Split <{split_input_shape_attr}> (image_shape)
+                hw = Concat <axis = 0> (h, w)
+                pad_hw = Sub (target_size, hw)
+                half_pad_hw = Div (pad_hw, i64_2)
+                another_half_pad_hw = Sub (pad_hw, half_pad_hw)
+                pad_value = Concat <axis = 0> (half_pad_hw, i64_0,another_half_pad_hw,i64_0)
+                {self.output_names[0]} = Pad({self.input_names[0]}, pad_value, const_val)
```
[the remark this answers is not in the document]
wejoncy: `axes` is a 1-D tensor also, we need `concat` anyway.

operators/vision/draw_bounding_box.hpp
wejoncy: Sure.

operators/vision/draw_bounding_box.cc
wejoncy: Sure.

operators/vision/draw_bounding_box.cc
wejoncy: Got it.

operators/vision/draw_bounding_box.cc
wejoncy: Used const reference already.
I want to use a separate variable to represent it.

onnxruntime_extensions/tools/pre_post_processing/steps/vision.py
```diff
@@ -604,3 +616,148 @@ def __init__(self, has_batch_dim: bool = False, name: Optional[str] = None):
...
         perms = [0, 3, 1, 2] if has_batch_dim else [2, 0, 1]
         super().__init__(perms, name)
+
+
+class DrawBoundingBoxes(Step):
+    """
+    Draw boxes on BGR image at given position.
+    Input shape: <uint8_t>{height, width, 3<BGR>}
+    boxes: <float>{num_boxes, 6<xmin, ymin, xmax, ymax, score, class>}
+        The coordinates should have been scaled back to the original image size
```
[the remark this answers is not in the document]
wejoncy: Yeah, we should make it more clear.

test/test_tools_add_pre_post_processing_to_model.py
wejoncy: > overlapping boxes

Yeah, The higher-score box will be drawn latter and has better visibility.

operators/vision/draw_bounding_box.hpp
skottmckay: IIRC we had to use `Ort` as the namespace for the public C++ API couldn't clash with the existing `onnxruntime` namespace used in all the other C++ code. 

I'm not aware of any complaints about the convenience of the `onnxruntime` namespace and `ort_extensions` is a pretty similar length. I don't find myself typing it very often - mainly a couple of times when adding a new class, once in the header and once in the .cc, and auto-complete usually means only a few characters actually need to be typed. 

If we want to be consistent with the style guide and internal ORT code that would be `ort_extensions` wouldn't it?

operators/vision/draw_bounding_box.cc
```diff
@@ -0,0 +1,267 @@
...
+    {{0, 255, 0}},      // Green
+    {{255, 0, 0}},      // Blue
+    {{255, 255, 0}},    // Cyan
+    {{255, 0, 255}},    // Magenta
+    {{0, 0, 128}},      // Dark Red/Maroon
+    {{0, 128, 0}},      // Dark Green/Lime
+    {{128, 0, 0}},      // Dark Blue/Navy
+    {{0, 0, 0}},        // Black
+    {{128, 128, 128}},  // Gray
+}};
```
skottmckay, line 31: 'Lime' is 0, 255, 0 according to https://www.htmlcsscolor.com/hex/00ff00 
Yellow and Orange are easily recognized colours that are not used and might be better than black and grey.

Suggested list/names (not ordered):
Red: 0, 0, 255
Yellow: 0, 255, 255
Lime: 0, 255, 0
Cyan: 255, 255, 0
Blue: 255, 0, 0
Magenta: 255, 0, 255
Orange: 0, 128, 255
Maroon: 0, 0, 128
Green: 0, 128, 0
Navy: 128, 0, 0

onnxruntime_extensions/tools/pre_post_processing/steps/vision.py
```diff
@@ -604,3 +619,149 @@ def __init__(self, has_batch_dim: bool = False, name: Optional[str] = None):
...
+                    "XYXY" (xmin ymin xmax ymax)  All values in the XYXY format should be absolute pixel values.
+                    "XYWH" (xmin ymin width height) 
+                    "CENTER_XYWH" (x_center, y_center, width, height) 
+                                  All values in the CENTER_XYWH format should be absolute pixel values.
+
+
+            thickness: Thickness of the box edge
+            num_colours: Number of colours to use
+                         We support 10 predefined colours and the other classes more than 10 wouldn't be drawn.
+                         colors are [red, green, blue, cyan, magenta, maroon, lime, navy, black]
```
skottmckay, line 651: Only 9 entries in the list.

operators/vision/draw_bounding_box.cc
```diff
@@ -0,0 +1,267 @@
...
+              });
+  }
+
+ public:
+  BoxArray(const std::vector<int64_t>& shape, gsl::span<const float> data, BoundingBoxFormat bbox_mode)
+      : shape_(shape), bbox_mode_(bbox_mode) {
+    SortBoxesByScore(data);
+  }
+
+  [[nodiscard]] gsl::span<const float> GetBox(size_t index) const {
```
skottmckay, line 66: nit: I understand using nodiscard when it's success/failure status being returned and if the calling code ignores that an error may be ignored. not clear why it's necessary here though.

operators/vision/draw_bounding_box.cc
```diff
@@ -0,0 +1,267 @@
...
+    assert(index < boxes_by_score_.size());
+    return boxes_by_score_[index];
+  }
+
+  [[nodiscard]] int64_t NumBoxes() const {
+    return shape_[0];
+  }
+
+  auto BBoxMode() const {
```
skottmckay, line 75: nit: using 'auto' in the public interface of a class is somewhat painful for the reader. they have to go find the definition of the member variable being returned in order to figure out the type. i.e. you're unnecessarily forcing them to work harder to understand the interface of your class.

operators/vision/draw_bounding_box.cc
```diff
@@ -0,0 +1,267 @@
...
+      color_used.emplace(box[kBoxClassIndex], color_used.size());
+    }
+    box_reverse.emplace_back(i, static_cast<int64_t>(color_used[box[kBoxClassIndex]]));
+  }
+
+  // A class which has higher score will be drawn on the top of the image.
+  std::sort(box_reverse.begin(), box_reverse.end(),
+            [](const std::pair<size_t, int64_t>& first_, const std::pair<size_t, int64_t>& second_) {
+              return first_.second < second_.second;
+            });
```
skottmckay, line 202: Can we use a reverse iterator instead of sorting?

operators/vision/draw_bounding_box.cc
skottmckay: In mobile scenarios when would you have a batch of inferences to perform at once?

Video maybe, but most things (took a photo, selected some text, recorded some audio) I would expect involved a batch size of 1 given the user cannot create input fast enough. 

And if you want to use NNAPI or CoreML you need to fix the batch size anyway. 

I think that might leave a scenario with a model with a fixed batch size that's processing video, and if we need to support that we could always use a Scan op around the preprocessing to handle the batch.

operators/vision/draw_bounding_box.cc
wejoncy: We are sorting this by colors, because smaller color index has higher priority.  And the `box_reverse` is sorted by score, so this is essential.

operators/vision/draw_bounding_box.cc
wejoncy: > And if you want to use NNAPI or CoreML you need to fix the batch size anyway.

I see.
CPU EP can't provide any gains for multi-batches.

skottmckay: CPU EP can support batches and there would most like be a performance gain. 

But need a concrete use case before we start considering whether it's worth implementing batch support in the pre/post processing, and how we want to do it. If the requirements of the theoretical use case can be met by using a Scan or Loop and the individual pre/post processing steps do not need to be batch aware, that would be significantly less work to develop/test/maintain.

onnxruntime_extensions/tools/pre_post_processing/steps/vision.py
```diff
@@ -604,3 +621,140 @@ def __init__(self, has_batch_dim: bool = False, name: Optional[str] = None):
...
+                    "XYXY" (xmin ymin xmax ymax)  All values in the XYXY format should be absolute pixel values.
+                    "XYWH" (xmin ymin width height) 
+                    "CENTER_XYWH" (x_center, y_center, width, height) 
+                                  All values in the CENTER_XYWH format should be absolute pixel values.
+
+
+            thickness: Thickness of the box edge
+            num_colours: Number of colours to use
+                         We support 10 predefined colours and the other classes more than 10 wouldn't be drawn.
+                         colors are [red, Yellow, Lime, cyan, blue, magenta, Orange, Maroon, Green, Navy]
```
skottmckay, line 653:
```suggestion
                         colors are [Red, Yellow, Lime, Cyan, Blue, Magenta, Orange, Maroon, Green, Navy]
```

operators/vision/draw_bounding_box.cc
```diff
@@ -0,0 +1,263 @@
...
+  thickness = std::clamp<int64_t>(thickness, 1, std::min(x_end - x_start, y_end - y_start));
+  if (x_end - x_start < thickness || y_end - y_start < thickness) {
+    ORTX_CXX_API_THROW("box is too small to draw", ORT_INVALID_ARGUMENT);
+    return;
+  }
+
+  if (x_start + thickness >= image.height || y_start + thickness >= image.width) {
+    ORTX_CXX_API_THROW("box is too small to draw", ORT_INVALID_ARGUMENT);
+    return;
+  }
```
skottmckay, line 167: Given we clamp the thickness do we need the tests here? If the thickness is == the width or height it's just going to be a solid filled in box isn't it? so we can draw it even though it's not that useful. but this seems like an edge case where the user specifies a really large thickness so drawing a filled in box vs. throwing might be a better outcome.

onnxruntime_extensions/tools/pre_post_processing/steps/vision.py
```diff
@@ -370,6 +381,12 @@ def _create_graph_for_step(self, graph: onnx.GraphProto, onnx_opset: int):
             # Split now requires the number of outputs to be specified even though that can be easily inferred...
             split_input_shape_attr += f", num_outputs = {len(dims)}"
             split_new_sizes_attr += ", num_outputs = 2"
+        
+        # Resize-18 has the attribute "not_larger/not_smaller" to specify the resize policy, however
+        # we want to support older opsets as well. 
+        ratio_resize_func = "ReduceMax"
+        if self.policy_ == "not_larger":
+            ratio_resize_func = "ReduceMin"
```
skottmckay, line 389: should we also check the policy == 'not_smaller' so we can return an error if the user incorrectly tries to use an unsupported policy that the ONNX Resize op defines?

operators/vision/draw_bounding_box.cc
wejoncy: There are two cases which will hit it.
-  Box has zero areas. Say [x1,y1,x2,y2] = [0,0,0,0] or even [1,1,0,0].
- Thickness is too large. So, we are drawing nothing on the image.

Shall we allow it by default?

skottmckay (approved):

PR #383, opened 2023-03-21
Improve offset map algorithm for RobertaTokenizer

sayanshaw24:
Use byte list in bpe algorithm to determine post-bpe token lengths for offset mapping.


wenbingl (approved):
Just to confirm, this is just a code refinement, nothing fixed?

PR #385, opened 2023-03-23
Add an audio decoder custom op for whisper end-to-end processing

wenbingl:


onnxruntime_extensions/_cuops.py
```diff
@@ -388,6 +388,19 @@ def get_outputs(cls):
         ]
 
 
+class AudioDecoder(CustomOp):
+    @classmethod
+    def get_inputs(cls):
+        return [
+            cls.io_def('audio_stream', onnx_proto.TensorProto.UINT8, [1, None])
```
sayanshaw24, line 395: not extremely important but at some point for all custom ops we should probably denote what the shapes represent for better readability, would help when modifying them in the future.

operators/audio/audio_decoder.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+  enum {
+    DEFAULT_STREAM = 0,
+    WAV_STREAM,
+    MP3_STREAM,
+    FLAC_STREAM
+  };
+
+  int ReadStreamFormat(OrtKernelContext* context, const int64_t* p_data) {
+    int stream_format = DEFAULT_STREAM;
```
sayanshaw24, line 30: is there anyway we can do this with size_t instead? (for consistency because we try to avoid using int because of overflow issues)

operators/audio/audio_decoder.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+  size_t GetInputTypeCount() const {
+    return 2;
+  }
+
+  ONNXTensorElementDataType GetInputType(size_t index) const {
+    return index == 0 ? ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT8 : ONNX_TENSOR_ELEMENT_DATA_TYPE_STRING;
+  }
+
+  OrtCustomOpInputOutputCharacteristic GetInputCharacteristic(size_t index) const {
```
sayanshaw24, line 145: should we add an explicit unit test for the optional inputs?

sayanshaw24 (dismissed):

onnxruntime_extensions/_cuops.py
wenbingl: Yes, the string of name is better.
1 was intentional since the operator now only support batch size = 1

operators/audio/audio_decoder.hpp
wenbingl: this int is used only for enumuration, and no algorithmic operation with other value, so no worry for overflow.

operators/audio/audio_decoder.hpp
wenbingl: Yes, will do.
Just thinking about how we can do it with signatures in _cuops.py.

operators/audio/audio_decoder.hpp
```diff
@@ -0,0 +1,157 @@
...
+      } else if (marker[0] == char(-1) && marker[1] == char(-13)) {
+        stream_format = MP3_STREAM;
+      } else {
+        ORTX_CXX_API_THROW("[AudioDecoder]: Unknown audio stream format", ORT_INVALID_ARGUMENT);
+      }
+    }
+    return stream_format;
+  }
+
+  template<typename _CLS_AUDIO, typename _FN_DECODER>
```
edgchen1, line 63: nit: identifier names starting with an underscore and uppercase letter are reserved - we shouldn't use them.
https://en.cppreference.com/w/cpp/language/identifiers

operators/audio/audio_decoder.hpp
```diff
@@ -0,0 +1,157 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
```
edgchen1, line 2:
```suggestion
// Licensed under the MIT License.

#pragma once

```

operators/audio/audio_decoder.hpp
```diff
@@ -0,0 +1,157 @@
...
+      } else {
+        ORTX_CXX_API_THROW("[AudioDecoder]: Unknown audio stream format", ORT_INVALID_ARGUMENT);
+      }
+    }
+    return stream_format;
+  }
+
+  template<typename _CLS_AUDIO, typename _FN_DECODER>
+  static size_t DrReadFrames(std::list<std::vector<float>>& frames, _FN_DECODER fx, _CLS_AUDIO& obj) {
+    const size_t default_chunk_size = 4096;
```
edgchen1, line 65: this is the number of frames, right? how about a name like `default_frame_count`?

operators/audio/audio_decoder.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+    OrtTensorDimensions input_dim(ort_, input);
+    if (!((input_dim.size() == 1) || (input_dim.size() == 2 && input_dim[0] == 1))) {
+      ORTX_CXX_API_THROW("[AudioDecoder]: Expect input dimension [n] or [1,n].", ORT_INVALID_ARGUMENT);
+    }
+
+    auto stream_format = ReadStreamFormat(context, p_data);
+
+    int64_t total_frames = 0;
+    std::list<std::vector<float>> lst_frames;
```
edgchen1, line 96: why `std::list`? I think generally `std::vector` is preferable.

operators/audio/audio_decoder.hpp
```diff
@@ -0,0 +1,157 @@
...
+      std::vector<float> buf;
+      buf.resize(default_chunk_size * obj.channels);
+      auto n_frames = fx(&obj, default_chunk_size, buf.data());
+      if (n_frames <= 0) {
+        break;
+      }
+      auto n_samples = n_frames * obj.channels;
+      total_frames += n_samples;
+      buf.resize(n_samples);
+      frames.emplace_back(buf);
```
edgchen1, line 78:
```suggestion
      frames.emplace_back(std::move(buf));
```

operators/audio/audio_decoder.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+    for (;;) {
+      std::vector<float> buf;
+      buf.resize(default_chunk_size * obj.channels);
+      auto n_frames = fx(&obj, default_chunk_size, buf.data());
+      if (n_frames <= 0) {
+        break;
+      }
+      auto n_samples = n_frames * obj.channels;
+      total_frames += n_samples;
```
edgchen1, line 76: I'm confused by the names here. why are we incrementing the frame count by a number of samples?

operators/audio/audio_decoder.hpp
edgchen1: actually, since we just copy each element (float vector) to the output, could we just use a single float vector to store all frame data?

operators/audio/audio_decoder.hpp
```diff
@@ -0,0 +1,157 @@
...
+      }
+      total_frames = DrReadFrames(lst_frames, drmp3_read_pcm_frames_f32, mp3_obj);
+
+    } else if (stream_format == FLAC_STREAM) {
+      drflac* flac_obj = drflac_open_memory(p_data, input_dim.Size(), nullptr);
+      if (flac_obj == nullptr) {
+        ORTX_CXX_API_THROW("[AudioDecoder]: unexpected error on FLAC stream.", ORT_RUNTIME_EXCEPTION);
+      }
+      total_frames = DrReadFrames(lst_frames, drflac_read_pcm_frames_f32, *flac_obj);
+      drflac_close(flac_obj);
```
edgchen1, line 111: avoid possible leak on exception. gsl::finally returns a helpful RAII type that runs something at end of scope.
```suggestion
      auto flac_obj_closer = gsl::finally([flac_obj]() { drflac_close(flac_obj); });
      total_frames = DrReadFrames(lst_frames, drflac_read_pcm_frames_f32, *flac_obj);
```

operators/audio/audio_decoder.hpp
```diff
@@ -0,0 +1,157 @@
+// Copyright (c) Microsoft Corporation.
```
edgchen1, line 1: can some of the implementation details be put into a .cc file?

wenbingl: The big advantage of using header file is we don't need to do if/else in CMake file, plus function declaration and implementation are unified, i.e. the function prototype can be changed just in one place.

operators/audio/audio_decoder.hpp
wenbingl: because we don't know the buf size of final decoding stream, not using a single float vector is to avoid the big memory allocation overhead. so does list vs. vector.

operators/audio/audio_decoder.hpp
wenbingl: the frame is the same of n_sample,

operators/audio/audio_decoder.hpp
edgchen1: > we don't need to do if/else in CMake file

not quite following what this if/else in CMake is - wouldn't it just be adding another file?

a disadvantage of doing this is adding unnecessary stuff for whoever includes this file. for example, macro definitions like DR_FLAC_IMPLEMENTATION and inclusion of external library headers. IMO, the encapsulation is worth splitting things into both a header and implementation file.

operators/audio/audio_decoder.hpp
edgchen1: "frame" and "sample" seem to be separate units. is this not the case?

operators/audio/audio_decoder.hpp
edgchen1: ah, I see.

is it possible to query the frame/sample (I don't know the right word) size of the whole stream beforehand and avoid repeated dynamic memory allocations and copies?

e.g., with APIs like these:
https://github.com/mackron/dr_libs/blob/4b3d07849537ce0b71b22180c0b1335eacc6e9be/dr_mp3.h#L440-L458

operators/audio/audio_decoder.hpp
wenbingl: check line: 47, this format value needs to be changed, which cannot be supported by enum type.

operators/audio/audio_decoder.hpp
wenbingl: good to know, updated

operators/audio/audio_decoder.hpp
wenbingl: If we add more audio custom ops in the future, and all of them implemented in headers file, we don't need to glob cc files according to different pattern, like what we did for tokenizer files and they looks a little messy.

.hpp extension file in the projects is not indented to be included in the multiple places, while .h files are common which are designed to be included by other header files.

operators/audio/audio_decoder.hpp
wenbingl: mp3 doesn't have a simple and light-weight to cal the final decompressed data size, like reading the value from the header. And these get_frame_count APIs needs decoding the frame firstly which makes it look heavy. So, it is not unworthy of adding a dependency on these APIs.

I have increased the decoding buf size to 256K from 4K(which seems to be small), the memory allocation should be less frequent and should not be a problem now.

sayanshaw24 (dismissed):

operators/audio/audio_decoder.hpp
edgchen1: > .hpp extension file in the projects is not indented to be included in the multiple places, while .h files are common which are designed to be included by other header files.

I was not aware of this convention. is it documented anywhere?

operators/audio/audio_decoder.hpp
```diff
@@ -0,0 +1,163 @@
...
+      buf.resize(data_size);
+      frames.emplace_back(std::move(buf));
+    }
+
+    return total_buf_size;
+  }
+
+  void Compute(OrtKernelContext* context) {
+    const OrtValue* input = ort_.KernelContext_GetInput(context, 0);
+    const int64_t* p_data = ort_.GetTensorData<int64_t>(input);
```
edgchen1, line 92: why `int64_t*` here? input 0 looks like it should be uint8 from CustomOpAudioDecoder::GetInputType()

operators/audio/audio_decoder.hpp
edgchen1: ok. I guess this is fine for the initial implementation, but I think there might be some potential performance improvement here.

operators/audio/audio_decoder.hpp
```diff
@@ -0,0 +1,163 @@
...
+      GetTensorMutableDataString(api_, ort_, context, ov_format, str_format);
+      const char* formats[] = {"default",
+                               "wav",
+                               "mp3",
+                               "flac"};
+      for (auto fmt : formats) {
+        if (str_format[0] == fmt) {
+          break;
+        }
+        stream_format++;
```
edgchen1, line 47: I think that using an enumeration type conveys more meaning than an int. this effectively maps from a string value to the enumeration value and doesn't have to be implemented with an integer increment.

operators/audio/audio_decoder.hpp
wenbingl: Good finding! Because all following usage are all type casted, the compiler didn't report issue.

operators/audio/audio_decoder.hpp
wenbingl: not sure you are suggesting there is any new C++ feature that get type string from enum class. But I just noticed we could static initialize a std::map now, which makes the code look more modern C++ style now.

operators/audio/audio_decoder.hpp
```diff
@@ -0,0 +1,168 @@
...
+
+  enum class AudioStreamType {
+    kDefault = 0,
+    kWAV,
+    kMP3,
+    kFLAC
+  };
+
+  AudioStreamType ReadStreamFormat(OrtKernelContext* context, const uint8_t* p_data) {
+    const std::map<std::string, AudioStreamType> format_mapping = {
```
edgchen1, line 35: nit: could be static
```suggestion
    static const std::map<std::string, AudioStreamType> format_mapping = {
```

edgchen1 (dismissed):

operators/audio/audio_decoder.hpp
edgchen1: > not sure if you are suggesting there is any new C++ feature that get type string from enum class

I'm not aware of this, it would be nice though.

but yeah, I was thinking of a map lookup. actually with the small number of entries we can probably get by with scanning an array of pairs too, but map is ok.

PR #387, opened 2023-03-27
add the onnxruntime 1.14 release into the CI pipeline

wenbingl:


.pipelines/ci.yml
```diff
@@ -21,22 +21,22 @@ jobs:
...
-          ort.version: '1.12.0'
-        py38-1110:
+          ort.version: '1.13.1'
+        py38-1121:
           python.version: '3.8'
           torch.version: 'torch==1.11.0+cpu torchvision==0.12.0+cpu torchaudio==0.11.0+cpu -f https://download.pytorch.org/whl/cpu/torch_stable.html'
-          ort.version: '1.11.0'
-        py37-1100:
+          ort.version: '1.12.1'
+        py37-1110:
```
sayanshaw24, line 36: should this be `py37-1111` (for version 1.11.1)? if so, it also needs to be changed everywhere else `py37-1110` is used.

wenbingl: Good finding. But given the pipeline was successfully running for a while, let's keep it as it was, and it will be dropped in the next version.

sayanshaw24: yes, sounds good!

sayanshaw24 (approved):

PR #389, opened 2023-03-31
Added optional outputs for GPT2, CLIP and Roberta Tokenizers

sayanshaw24:


operators/tokenizer/roberta_tokenizer.cc
wenbingl, line 1: Can you add some unit test cases as well?

sayanshaw24: Yep working on it, trying to ensure functionality is as expected first.

sayanshaw24: Done!

PR #393, opened 2023-04-10
Fix ORT combined build nlohmann_json fetch.

edgchen1:
Use dependency name 'nlohmann_json' which is the same name that ORT uses.

https://github.com/microsoft/onnxruntime/blob/4a676b011a71f5a1ddacdac566f5579f0e132976/cmake/external/onnxruntime_external_deps.cmake#L154-L159

Then the FetchContent calls in both ORT and ORT Extensions won't try to fetch this dependency as two separate dependencies and complain about an existing target created by the first fetch.

### DIFF
--- a/cmake/externals/json.cmake
+++ b/cmake/externals/json.cmake
@@ -1,11 +1,11 @@
-FetchContent_Declare(json
+FetchContent_Declare(nlohmann_json
   GIT_REPOSITORY https://github.com/nlohmann/json.git
   GIT_TAG v3.10.5)
 
 set(JSON_BuildTests OFF CACHE INTERNAL "")
 
-FetchContent_GetProperties(json)
-if(NOT json_POPULATED)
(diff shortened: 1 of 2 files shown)


snnn (approved):

cmake/externals/json.cmake
```diff
@@ -1,11 +1,11 @@
-FetchContent_Declare(json
+FetchContent_Declare(nlohmann_json
   GIT_REPOSITORY https://github.com/nlohmann/json.git
   GIT_TAG v3.10.5)
```
wenbingl, line 3: @snnn, @edgchen1 Did any of you know what happens if the version mismatch?

wenbingl (approved):

cmake/externals/json.cmake
snnn: The one who appears first will get used.

PR #395, opened 2023-04-12
Initial work for ort extensions nuget package

sayanshaw24:
Started `.nuspec` file generator.

**TODO**:
1. Generate `.nuspec` file with all necessary files needed for nuget package.
3. Create `.props` file for consuming nuget package
4. Call `nuget pack` using nuget.exe CLI to create the nuget package.
5. Possibly add code signing? (will be required for 1P customers as per Sunghoon)
6. Publish nuget package

### DIFF
--- a/tools/nuget/generate_nuspec_for_native_nuget.py
+++ b/tools/nuget/generate_nuspec_for_native_nuget.py
@@ -1,0 +1,150 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import argparse
+import os
+import re
+import sys
+from pathlib import Path
+
+def parse_arguments():
+    parser = argparse.ArgumentParser(
+        description="Create ORT Extensions nuget spec script", usage=""
+    )
+    # Main arguments
+    parser.add_argument("--package_name", required=True, help="ORT package name. Eg: ort-extensions-nuget")
+    parser.add_argument("--package_version", required=True, help="ORT package version. Eg: 0.7.0")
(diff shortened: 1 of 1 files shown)


tools/nuget/generate_nuspec_for_native_nuget.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def generate_repo_url(line_list, repo_url, commit_id):
+    line_list.append('<repository type="git" url="' + repo_url + '"' + ' commit="' + commit_id + '" />')
+
+def generate_dependencies(line_list):
+    dependencies_list = ['<group targetFramework="native">']
+
+    # Update dependencies as needed here:
+    dependencies_list.append('<dependency id="blingfire" version="0831265c1aca95ca02eca5bf1155e4251e545328"/>')
```
wenbingl, line 65: ort-extensions dll statically linked any dependency so it doesn't depend any 3rd party DLL here.

sayanshaw24: perfect, thanks! will remove them in the next commit.

wenbingl: Before completing this script, can we check in nuspec file of amd64 native binary only?

sayanshaw24: sure - was initially motivated to write the script because generating the files list is easier with a script because we can do it recursively, would take a lot longer to just manually add the files list.

PR #397, opened 2023-04-13
[object detection ppp] YoLo as example

wejoncy:


onnxruntime_extensions/tools/pre_post_processing/steps/vision.py
```diff
@@ -760,3 +760,221 @@ def _create_graph_for_step(self, graph: onnx.GraphProto, onnx_opset: int):
...
+        return converter_graph
+
+
+class LinearMapBox(Step):
+    """
+    Mapping boxes coordinate to scale in original image.
+    The coordinate of boxes from detection model is relative to the input image of network, 
+    image is scaled and padded/cropped. So we need to do a linear mapping to get the real coordinate of original image.
+    input:
+        ori_img: original image decoded from jpg/png<uint8_t>[H, W, 3<BGR>]
```
skottmckay, line 909: Missing the boxes as input 0.

nit: 'ori_img' is unnecessarily abbreviated. At least 'orig_img' but 'original_image' would be better. Prefer descriptive names unless there's a good reason not to. more developers will spend time reading than code than typing the name (and autocomplete should handle most of that for you anyway).

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
```diff
@@ -162,6 +162,90 @@ def superresolution(model_file: Path, output_file: Path, output_format: str, onn
...
+
+    model_input_shape = model.graph.input[0].type.tensor_type.shape
+    model_output_shape = model.graph.output[0].type.tensor_type.shape
+
+    # Should we use 640*640 as default input shape if no specific shape provided?
+    w_in = model_input_shape.dim[-1].dim_value if model_input_shape.dim[-1].HasField("dim_value") else 640
+    h_in = model_input_shape.dim[-2].dim_value if model_input_shape.dim[-1].HasField("dim_value") else 640
+
+    # We have no idea if we need to transpose the output or not.
+    # To run the model and get the real shape is the only way to know the concrete shape.
```
skottmckay, line 193: The model may define the concrete shape (at least the exported yolov8 I have does) so we could check for dim_value's in the output shape.

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
```diff
@@ -162,6 +162,90 @@ def superresolution(model_file: Path, output_file: Path, output_format: str, onn
...
+    """
+    model = onnx.load(str(model_file.resolve(strict=True)))
+    inputs = [create_named_value("image", onnx.TensorProto.UINT8, ["num_bytes"])]
+
+    model_input_shape = model.graph.input[0].type.tensor_type.shape
+    model_output_shape = model.graph.output[0].type.tensor_type.shape
+
+    # Should we use 640*640 as default input shape if no specific shape provided?
+    w_in = model_input_shape.dim[-1].dim_value if model_input_shape.dim[-1].HasField("dim_value") else 640
+    h_in = model_input_shape.dim[-2].dim_value if model_input_shape.dim[-1].HasField("dim_value") else 640
```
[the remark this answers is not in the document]
wejoncy: Will add args used to pass input shape.

onnxruntime_extensions/tools/pre_post_processing/steps/vision.py
```diff
@@ -760,3 +760,221 @@ def _create_graph_for_step(self, graph: onnx.GraphProto, onnx_opset: int):
         )
 
         return converter_graph
+
+
+class SplitOutBoxAndScore(Step):
+    r"""
+    Split the output of the model into boxes and scores. This step will also handle the optional object score.
```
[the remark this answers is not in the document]
wejoncy: >optionally skip the obj slot if necessary right?
Sort of. But we can't skip it, we have to do `conf(cls) = conf(obj) * conf(cls)`.  Right now I used a tricky solution with `Where` Op to settle the dynamic of having  "conf(obj)" or not. `conf(obj) =where(0_or_1, obj_or_none, 1)`.

Doing this along with NMS makes the text-IR really huge and complicated.  Isolate this step makes NMS step focus on doing pure NMS stuff.

onnxruntime_extensions/tools/pre_post_processing/steps/vision.py
```diff
@@ -760,3 +760,221 @@ def _create_graph_for_step(self, graph: onnx.GraphProto, onnx_opset: int):
...
+
+class LinearMapBox(Step):
+    """
+    Mapping boxes coordinate to scale in original image.
+    The coordinate of boxes from detection model is relative to the input image of network, 
+    image is scaled and padded/cropped. So we need to do a linear mapping to get the real coordinate of original image.
+    input:
+        ori_img: original image decoded from jpg/png<uint8_t>[H, W, 3<BGR>]
+        scaled_img: scaled image, but without padding/crop[<uint8_t>[H1, W1, 3<BGR>]
+        net_in_img: scaled image and with padding/crop[<uint8_t>[H2, W3, 3<BGR>]
```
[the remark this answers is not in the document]
wejoncy: We have to do `OFFSET` as Lettbox Step padded/cropped the resized image. It means that coordinates are centered on cropped/padded image.
The point with coordinates (0, 0) should be based on resized image instead of cropped image.

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
wejoncy: Would it be better to assume model has a symbolic shape? Will check output_shape if there is and perverse the backward shape-infer.

onnxruntime_extensions/tools/pre_post_processing/steps/vision.py
```diff
@@ -760,3 +760,221 @@ def _create_graph_for_step(self, graph: onnx.GraphProto, onnx_opset: int):
...
+                {self.output_names[0]} = Identity (boxes_o)
+                {self.output_names[1]} = Identity (scores_o)
+
+            }}
+            """
+        )
+        return converter_graph
+
+
+class NMS(Step):
```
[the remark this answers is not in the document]
wejoncy: Done. Thanks

onnxruntime_extensions/tools/pre_post_processing/steps/vision.py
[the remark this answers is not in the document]
wejoncy: Added

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
```diff
@@ -162,6 +162,90 @@ def superresolution(model_file: Path, output_file: Path, output_format: str, onn
...
+        (LinearMapBox(),
+         [
+            utils.IoMapEntry("ConvertImageToBGR", producer_idx=0, consumer_idx=1),
+            utils.IoMapEntry("Resize", producer_idx=0, consumer_idx=2),
+            utils.IoMapEntry("LetterBox", producer_idx=0, consumer_idx=3),
+        ]),
+        (DrawBoundingBoxes(mode='CENTER_XYWH', num_classes=num_classes),
+         [
+            utils.IoMapEntry("ConvertImageToBGR", producer_idx=0, consumer_idx=0),
+            utils.IoMapEntry("LinearMapBox", producer_idx=0, consumer_idx=1),
```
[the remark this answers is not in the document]
wejoncy: Will add more comments to explain how does the process work.

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
```diff
@@ -162,6 +162,123 @@ def superresolution(model_file: Path, output_file: Path, output_format: str, onn
...
+            import numpy as np
+            session = onnxruntime.InferenceSession(str(model_file), providers=["CPUExecutionProvider"])
+            input_name = session.get_inputs()[0].name
+            inp = {input_name: np.random.rand(1, 3, w_in, h_in).astype(dtype=np.float32)}
+            outputs = session.run(None,  inp)[0]
+            assert len(outputs.shape) == 3 and outputs.shape[0] == 1, "model output shape is not (1, n, m)"
+            if outputs.shape[1] < outputs.shape[2]:
+                need_transpose = True
+        except:
+            pass
```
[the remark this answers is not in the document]
wejoncy: Make sense. Error out or warning for those indeterminate models.

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
```diff
@@ -162,6 +162,134 @@ def superresolution(model_file: Path, output_file: Path, output_format: str, onn
...
+    output_shape = [model_output_shape.dim[i].dim_value if model_output_shape.dim[i].HasField("dim_value") else -1
+                    for i in [-2, -1]]
+    if output_shape[0] != -1 and output_shape[1] != -1:
+        need_transpose = output_shape[0] < output_shape[1]
+    elif len(model.graph.input) > 1:
+        print(""" warning!!
+            The model has more than one inputs and the output shape is not fixed.
+            Output shape would be assumed to be [num_boxes, num_classes+4(reg)].
+            If this behavior is not expected, please run the onnx shape inference first to \
+            infer the output shape then run this script again.""")
```
skottmckay, line 219: Not quite following. We're trying to infer from the output shape if a transpose is needed. How does the number of inputs affect that?

Is it valid to continue? We don't know if we need to transpose or not, so wouldn't we produce a broken model if it was actually required?

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
```diff
@@ -162,6 +162,134 @@ def superresolution(model_file: Path, output_file: Path, output_format: str, onn
...
+    else:
+        try:
+            import numpy as np
+            import onnxruntime
+        except ImportError:
+            raise ImportError(
+                "Please install onnxruntime and numpy to run this script. eg 'pip install onnxruntime numpy'")
```
skottmckay, line 226: nit: tell the user why they need to do that - i.e. we need to execute the model to determine the output shape in order to add the correct post-processing.

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
wejoncy: We don't know how to generate inputs for such a model with more than one input.

Yes, it's possible to  we produce a broken model. Should we just error out here?

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
```diff
@@ -162,6 +162,134 @@ def superresolution(model_file: Path, output_file: Path, output_format: str, onn
...
+    You should assure this model has only one input, and the input shape is [1, 3, h, w].
+    The model has either one or more outputs. 
+        If the model has one output, the output shape is [1,num_boxes, coor+(obj)+cls] 
+            or [1, coor+(obj)+cls, num_boxes].
+        If the model has more than one outputs, you should assure the first output shape is 
+            [1, num_boxes, coor+(obj)+cls] or [1, coor+(obj)+cls, num_boxes].
+    Note: (obj) means it's optional.
+
+    :param model_file: The model file path.
+    :param output_file: The output file path.
```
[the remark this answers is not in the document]
wejoncy: Sure.

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
```diff
@@ -213,17 +217,19 @@ def yolo_detection(model_file: Path, output_file: Path, output_format: str = 'jp
...
     elif len(model.graph.input) > 1:
         print(""" warning!!
-            The model has more than one inputs and the output shape is not fixed.
-            Output shape would be assumed to be [num_boxes, num_classes+4(reg)].
+            The model possesses more than one inputs, indicating we can't generate the input 
+ data and map to model inputs to execute model by onnx-runtime and we can't get concrete output shape.
+            Output shape will be assumed to be [num_boxes, num_classes+4(reg)].
             If this behavior is not expected, please run the onnx shape inference first to \
-            infer the output shape then run this script again.""")
+            infer the output shape out then run this script again.""")
```
skottmckay: I think it's best to keep this simple and remove the `elif` as I think it's adding unnecessary complication and confusion and not a lot of value.

I don't think we need to try and support or detect a model that unexpectedly has multiple required inputs in this predefined handler for yolo. If someone has such a model they can cut-and-paste the implementation and adjust as needed. At the most add an assert in the `else` to check num inputs is 1.

test/test_tools_add_pre_post_processing_to_model.py
```diff
@@ -590,7 +590,7 @@ def get_model_output():
         
     def test_FastestDet(self):
         # https://github.com/dog-qiuqiu/FastestDet
-        # a minor fix is to adapt output with yolo output format, including bounding box regression inside.
+        # a minor fix is to acommodate output with yolo output format, including bounding box regression inside.
```
skottmckay, line 593:
```suggestion
        # a minor fix is to accommodate output with yolo output format, including bounding box regression inside.
```

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
wejoncy: Sure.

skottmckay (approved):

PR #402, opened 2023-04-21
Added windows nuget work temporarily for testing

sayanshaw24:
Continued work with all other builds including mobile: https://github.com/microsoft/onnxruntime-extensions/compare/skottmckay/NugetWIP


nuget/targets/netstandard/Microsoft.ML.OnnxRuntime.Extensions.props
```diff
@@ -0,0 +1,80 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <ItemDefinitionGroup>
+    <ClCompile>
+      <AdditionalIncludeDirectories>$(MSBuildThisFileDirectory)../../build/native/include/;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
wenbingl, line 6: for desktop, we don't have any header files needs to be published.

nuget/targets/netstandard/Microsoft.ML.OnnxRuntime.Extensions.props
```diff
@@ -0,0 +1,80 @@
...
+    </ClCompile>
+  </ItemDefinitionGroup>
+
+  <ItemDefinitionGroup Condition="'$(PlatformTarget)' == 'ARM64'">
+    <Link>
+      <AdditionalDependencies>$(MSBuildThisFileDirectory)../../runtimes/win-arm64/native/ortextensions.lib;%(AdditionalDependencies)</AdditionalDependencies>
+    </Link>
+  </ItemDefinitionGroup>
+
+  <ItemDefinitionGroup Condition="'$(PlatformTarget)' == 'ARM'">
```
wenbingl, line 16: ARM for WIndows doesn't work well, consider to remove it.

nuget/targets/netstandard/Microsoft.ML.OnnxRuntime.Extensions.props
```diff
@@ -0,0 +1,80 @@
...
+    <Link>
+      <AdditionalDependencies>$(MSBuildThisFileDirectory)../../runtimes/win-arm/native/onnxruntime.lib;%(AdditionalDependencies)</AdditionalDependencies>
+    </Link>
+  </ItemDefinitionGroup>
+
+  <ItemDefinitionGroup Condition="'$(PlatformTarget)' == 'x64' OR ('$(PlatformTarget)' == 'AnyCPU' AND '$(Prefer32Bit)' != 'true')">
+   <Link>
+      <AdditionalDependencies>$(MSBuildThisFileDirectory)../../runtimes/win-x64/native/onnxruntime.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
wenbingl, line 24: why is this dependency required?

nuget/targets/netstandard/Microsoft.ML.OnnxRuntime.Extensions.props
```diff
@@ -0,0 +1,80 @@
...
+
+  <PropertyGroup>
+    <OnnxRuntimeBinary>$(MSBuildThisFileDirectory)../../runtimes/win-$(EnginePlatform)/native/ortextensions.dll</OnnxRuntimeBinary>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!-- TODO: Should be able to have a single entry and use something like OnnxRuntimeBinary in it.
+               Need to handle AnyCPU with Prefer32Bit.  -->
+    <!-- x64 -->
+    <None Include="$(MSBuildThisFileDirectory)..\..\runtimes\win-x64\native\ortextensions.dll"
```
wenbingl, line 49: is this 'runtime' folder needed?

nuget/NativeNuget.nuspec
```diff
@@ -0,0 +1,48 @@
...
+		2. iOS package: CocoaPods
+		3. PrePostProcessor tool for mobile model
+		4. Super-resolution model pre- / post- processing end-to-end examples
+	</releaseNotes>
+	<copyright>© Microsoft Corporation. All rights reserved.</copyright>
+	<tags>ONNX ONNXRuntime AI Machine Learning</tags>
+	<icon>ORT_icon_for_light_bg.png</icon>
+	<license type="file">LICENSE.txt</license>
+	<projectUrl>https://github.com/Microsoft/onnxruntime-extensions</projectUrl>
+	<repository type="git" url="https://github.com/Microsoft/onnxruntime-extensions.git" commit="98a303e639abff092f77295d39a31b5411d4f110" />
```
wenbingl, line 29: Either this commit id needs to be removed, or this .nuspec file will be generated dynamically.

nuget/NativeNuget.nuspec
```diff
@@ -0,0 +1,48 @@
...
+	</metadata>
+	<!-- assemble files in the /nuget directory and run nuget pack from -->
+	<files>
+		<file src="..\LICENSE.txt" target="LICENSE.txt" />
+		<file src="..\includes\*.h" target="build\native\include" />
+		<file src="..\includes\*.hpp" target="build\native\include" />
+		<!-- Windows, Linux, macOS via netstandard -->
+		<file src="..\build\Windows\Debug\bin\Debug\ortextensions.dll" target="runtimes\win-x64\native" />
```
wenbingl, line 37: should it be 'Release', instead of 'Debug'?

LICENSE.txt
wenbingl, line 1: move this file to nuget folder

ORT_icon_for_light_bg.png
wenbingl, line 1: move it to nuget folder?

nuget/NativeNuget.nuspec
wenbingl, line 1: Can you also add nuget.exe command into $projectdir\build.bat?

skottmckay: Needs a lot more infrastructure before that is considered. The nuget package combines builds from multiple platforms. Something has to create and assemble those so you can't just simply do a local build and generate a nuget package until the nuspec is dynamically generated to take into account which builds are available.

nuget/NativeNuget.nuspec
skottmckay: Commit id needs to be set by the CI before we release the package

nuget/NativeNuget.nuspec
skottmckay: This needs to be updated to match where the CI downloads the artifacts from the Windows build to. Current path is hardcoded from local testing.

nuget/targets/netstandard/Microsoft.ML.OnnxRuntime.Extensions.props
skottmckay: This nuget package provides a native binary so if someone wants to use that directly they need the headers.

This is consistent with what the ORT native nuget package includes.

nuget/targets/netstandard/Microsoft.ML.OnnxRuntime.Extensions.props
skottmckay: This could be trimmed to the required platforms. Setup was copied from what ORT supports.

nuget/targets/netstandard/Microsoft.ML.OnnxRuntime.Extensions.props
skottmckay: From what I can see it provides the exports info for the dll. Assume that is needed by C# given we include this in the ORT nuget package.

nuget/targets/netstandard/Microsoft.ML.OnnxRuntime.Extensions.props
skottmckay: This is the structure used in the ORT native package. 

https://learn.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks#architecture-specific-folders

nuget/targets/netstandard/Microsoft.ML.OnnxRuntime.Extensions.props
sayanshaw24: removed

nuget/targets/netstandard/Microsoft.ML.OnnxRuntime.Extensions.props
sayanshaw24: done

nuget/targets/netstandard/Microsoft.ML.OnnxRuntime.Extensions.props
sayanshaw24: not necessarily - but probably better to have it for the time being so it is easier to add to in the future when adding more support for other runtimes.

nuget/NativeNuget.nuspec
sayanshaw24: good catch, thanks! i believe this was because Scott was building with build_lib.bat rather than build.bat, updated now.

nuget/NativeNuget.nuspec
sayanshaw24: removed

nuget/NativeNuget.nuspec
wenbingl: build.bat is only for developer, and I agree the it would be much complicated to generate the release package, which will be done in some build/release pipeline.

wenbingl (approved):

PR #405, opened 2023-04-21
Remove unnecessary nupkg file and update nuspec

sayanshaw24:


snnn (approved):

wenbingl (approved):

nuget/NativeNuget.nuspec
```diff
@@ -3,14 +3,14 @@
...
-	<version>0.8.0</version>
+	<version>0.8.0-alpha</version>
 	<authors>Microsoft</authors>
 	<owners>Microsoft</owners>
 	<description>ONNX Runtime Extensions NuGet Package</description>
 	<!-- TODO: Update release notes prior to 0.8.0 release with new features here -->
 	<releaseNotes>
 		General
-		1. New custom operators: RobertaTokenizer, ClipTokenizer, EncodeImage, DecodeImage
+		1. New custom operators: Whisper, YOLO, RobertaTokenizer, ClipTokenizer, EncodeImage, DecodeImage
```
skottmckay, line 13: nit: The custom operator is to draw bounding boxes. Should be usable by any object detection model. YOLO is just a use case, so listing the operator name is probably more meaningful.

sayanshaw24: sounds good, i believe i'm supposed to get the updated release notes as needed from nat, will update them when i get them, YOLO was something i just added for the time being.

PR #407, opened 2023-04-26
Add a nuget test app

wenbingl:


test/nuget/ClipTokenizerTest/cliptok.onnx
sayanshaw24, line 1: can we use some existing model in test/data folder (like test_bert_tokenizer.onnx) and use that to test so we don't have to add another onnx file to the repo?

test/nuget/ClipTokenizerTest/Program.cs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                SessionOptions sessionOptions = new();
+                // Register the onnxruntime-extensions DLL for the custom ops.
+                sessionOptions.RegisterCustomOpLibraryV2(extensionsDllName, out handle);
+
+                var tokenizeSession = new InferenceSession(modelFilePath, sessionOptions);
+                var inputTensor = new DenseTensor<string>(new string[] { "This is a test string for tokenizer" }, new int[] { 1 });
+                var inputString = new List<NamedOnnxValue> { NamedOnnxValue.CreateFromTensor<string>("string_input", inputTensor) };
+                var tokens = tokenizeSession.Run(inputString);
+                var ids = tokens.First();
```
sayanshaw24, line 33: can you add an assert here and just hardcode the expected input ids from a python run so it functions as a test since it is classified as a test? or we could move this entire test app to the tutorials folder perhaps?

probably better and easier to do the first option.

test/nuget/ClipTokenizerTest/cliptok.onnx
wenbingl: we don't have .net CI pipeline, it would be make more sense to move it to tutorial folder.
I will do it later.

tutorials/demo4dotnet/ClipTokenizerTest/cliptok.onnx
sayanshaw24, line 1: should we move the cliptok model to test/data or some common folder so when we have an e2e python tutorial for clip too we can use the same model without confusion as to why it is under the demo4dotnet?

sayanshaw24 (approved):

tutorials/demo4dotnet/ClipTokenizerTest/cliptok.onnx
wenbingl: This demo project should be free to copy, it will bring some trouble to users if there is some data file dependency on other folders.

PR #408, opened 2023-04-26
[NuGet] All platform package pipeline

wejoncy:
Nuget pipeline for
- Windows x64
- Windows x86
- Windows arm
- Windows arm64
- Linux x64
- Linux arm64
- Macos x64
- Macos arm64
- Macos universe
- Android aar(all archs)
- iOS xcframework (all archs)

TODO:
- how to build extionsions for mac arm64?


.pipelines/templates/c-api-windows-cpu.yml
```diff
@@ -0,0 +1,105 @@
...
+
+  steps:
+    - checkout: self
+      clean: true
+      submodules: none
+
+    - task: BatchScript@1
+      displayName: 'build onnxruntime-extensions'
+      inputs:
+        filename: '.\build.bat'
```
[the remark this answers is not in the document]
wenbingl: @skottmckay TFSTRING and RE2 are added for tf2onnx converter at the initial time. and they don't bring too much footprint increasing.

.pipelines/templates/c-api-linux-cpu.yml
```diff
@@ -0,0 +1,72 @@
...
+  pool: ${{parameters.PoolName}}
+  steps:
+    - checkout: self
+      clean: true
+      submodules: none
+
+    - task: ShellScript@2
+      displayName: 'build onnxruntime-extensions for Linux'
+      inputs:
+        scriptPath: 'build.sh'
```
wenbingl, line 35: the same cmake configuration with Windows

.pipelines/templates/c-api-linux-cpu.yml
```diff
@@ -0,0 +1,72 @@
...
+
+    - template: set-package-version-variable-step.yml
+      parameters:
+        IsReleaseBuild: true
+        PackageVersionVariableName: ORT_EXTENSIONS_VERSION
+
+    - bash: |
+       cd $(Build.SourcesDirectory)
+
+       lib_target=$(Build.BinariesDirectory)/onnxruntime-extensions-linux-${{parameters.OrtExtensionsArch}}-$(ORT_EXTENSIONS_VERSION)/lib
```
wenbingl, line 47: Not sure if I missed, please make sure the .so file is 'stripped' before packing into NuGet package.

.pipelines/templates/c-api-windows-cpu.yml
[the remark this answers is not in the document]
skottmckay: What is the current binary size on the various platforms?

tools/ci_build/generate_nuspec_for_native_nuget.py
```diff
@@ -0,0 +1,462 @@
...
+
+
+def generate_owners(line_list, owners):
+    line_list.append("<owners>" + owners + "</owners>")
+
+
+def generate_description(line_list, package_name):
+    description = ""
+
+    if package_name == "Microsoft.AI.MachineLearning":
```
skottmckay, line 98: We only need to support Microsoft.ML.OnnxRuntime.Extensions don't we? I expected we did not need any logic here handling different package names/types so we can remove all that. is_cpu_package should always be true (that is for when ORT is building a GPU nuget)

tools/ci_build/generate_nuspec_for_native_nuget.py
```diff
@@ -0,0 +1,462 @@
...
+            "netstandard",
+            args.package_name + ".targets",
+        )
+        os.system(copy_command + " " + source_targets + " " + target_targets)
+        files_list.append("<file src=" + '"' + target_targets + '" target="build\\native" />')
+        files_list.append("<file src=" + '"' + target_targets + '" target="build\\netstandard1.1" />')
+        files_list.append("<file src=" + '"' + target_targets + '" target="build\\netstandard2.0" />')
+
+        # Process xamarin targets files
+        if args.package_name == "Microsoft.ML.OnnxRuntime.Extensions_xmamarin":
```
skottmckay, line 282: The Microsoft.ML.OnnxRuntime.Extensions package should include all the xamarin and maui targets. Trying to avoid multiple nuget packages as it greatly complicates usage.

tools/ci_build/generate_nuspec_for_native_nuget.py
```diff
@@ -0,0 +1,462 @@
...
+        pkg_name += cpu_arch
+    elif os == "osx":
+        pkg_name = "onnxruntime-extensions-osx-" + cpu_arch
+    return pkg_name
+
+
+# nuget_artifacts_dir: the directory with uncompressed C API tarball/zip files
+# files_list: a list of xml string pieces to append
+# This function has no return value. It updates files_list directly
+def generate_file_list(nuget_artifacts_dir, files_list, include_pdbs):
```
skottmckay, line 29: nit: it feels like we're greatly overcomplicating things by copying so much from the ORT script which needs to handle a lot more things like building multiple different packages. 

The ORT setup is extremely complicated and hard to understand. The generation script is run from a proj file in the /csharp directory, you have to go find the CI to figure out the args to run that with and how, it writes the nuspec in a build output dir from the ORT build, and packs using custom commands in the proj. None of that is intuitive or simple to debug. 

Would it be simpler to understand/maintain if we just had a script to update a checked in NativeNuget.nuspec like we have currently? We know exactly what files we want the CI to include and those are fixed. The script can update commit id, version, etc. as needed but it would be editing an existing file vs. trying to generate everything dynamically.

If someone wants to test locally I'm starting to think it's easier for them to edit a checked in nuspec file to update the 'src' paths for things they have, and to remove the entries in 'files' for things they don't. Also easy to directly run `nuget pack` once they have done so.

.pipelines/templates/c-api-windows-cpu.yml
```diff
@@ -0,0 +1,105 @@
...
+  default: ''
+
+- name: ort_build_pool_name
+  type: string
+  default: 'Win-CPU-2021'
+
+- name: UseIncreasedTimeoutForTests
+  displayName: Increase timeout for tests? Set it to false if you are doing an Onnx Runtime release.
+  type: boolean
+  default: false
```
skottmckay, line 35: Is this relevant here?

.pipelines/templates/ios_shared_lib_build.yml
```diff
@@ -0,0 +1,71 @@
...
+          python -m pip install cmake==3.25.0
+        displayName: "Install CMake 3.25.0"
+
+      - template: set-package-version-variable-step.yml
+        parameters:
+          IsReleaseBuild: ${{ parameters.IsReleaseBuild }}
+          PackageVersionVariableName: ORT_EXTENSIONS_POD_VERSION
+
+      - script: |
+          python ./tools/gen_selectedops.py ./tools/ios/package_ops.config
```
skottmckay, line 43: same comment here regarding the nuget build needing to include the same ops across all platforms

tools/android/build_aar.py
```diff
@@ -55,6 +55,7 @@ def build_for_abi(
 def build_aar(
     output_dir: Path,
     config: str,
+    build_target: str,
```
skottmckay, line 58: nit: confusing to have a file and function called `build_aar` that is building an AAR or a shared library.

tools/android/build_aar.py
```diff
@@ -198,6 +214,7 @@ def main():
     build_aar(
         output_dir=args.output_dir,
         config=args.config,
+        build_target=args.build_target,
```
skottmckay, line 217: The nuget package uses the aar. Why do we need to ability to just build the shared library if that isn't needed for nuget?

tools/ci_build/windows/extract_nuget_files.ps1
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# This file is used by Zip-Nuget Packaging NoContribOps Pipeline,Zip-Nuget-Java Packaging Pipeline
```
skottmckay, line 4: comment needs updating

.pipelines/templates/c-api-cpu.yml
```diff
@@ -0,0 +1,269 @@
...
+#  jobs:
+#  - template: ../android_packaging.yml
+#
+#- stage: iOS_Full_xcframework
+#  dependsOn: []
+#  jobs:
+#    - template: ../ios_packaging.yml
+
+
+- stage: NuGet_Packaging_CPU
```
skottmckay, line 126: If this config is responsible for assembling the nuget package for multiple different platforms, 'c-api-cpu.yml' doesn't really convey that. Can we rename to something more meaningful?

.pipelines/templates/c-api-ios-cpu.yml
```diff
@@ -0,0 +1,151 @@
+# packaging pipeline for iOS CocoaPods package
```
skottmckay, line 1: can we maybe call it build-ios-cocoapods-package.yml or something like that? would be good to consider all the yml file names and try and give them a meaningful name for what they do.

tools/ci_build/windows/extract_nuget_files.ps1
```diff
@@ -0,0 +1,69 @@
...
+
+# Re-construct a build directory that contains binaries from all the different platforms we're including
+# in the native ORT nuget package
+$nuget_artifacts_dir = "$Env:BUILD_BINARIESDIRECTORY\RelWithDebInfo\RelWithDebInfo\nuget-artifacts"
```
skottmckay, line 8: Is it possible to separate things out a little better so steps like this don't have a dependency on a build output dir from a local build?

i.e. copy ALL the required build artifacts to one place ($Env:BUILD_BINARIESDIRECTORY\build-artifacts) and extract them as needed to setup the directory structure the nuget pack needs (e.g. somewhere like $Env:BUILD_BINARIESDIRECTORY\nuget would be fine). 

That way in theory anything could use the build artifacts as they're all in one place and we cleanly separate out the nuget specific usage.

tools/ci_build/windows/extract_nuget_files.ps1
```diff
@@ -0,0 +1,69 @@
...
+Get-ChildItem $Env:BUILD_BINARIESDIRECTORY\nuget-artifact -Filter *.tar |
+Foreach-Object {
+ $cmd = "7z.exe x $($_.FullName) -y -o$nuget_artifacts_dir"
+ Write-Output $cmd
+ Invoke-Expression -Command $cmd
+}
+
+# copy android AAR.
+# for full build of onnxruntime-extensions Android AAR, there should only be one .aar file
+# called onnxruntime-extensions-android-x.y.z.aar but sanity check that
```
skottmckay, line 39: I don't think this check is relevant/needed here. It was for ORT where we potentially had 'full' and 'mobile' AAR files.

.pipelines/templates/android_shared_lib_build.yml
```diff
@@ -0,0 +1,70 @@
...
+        inputs:
+          versionSpec: "11"
+          jdkArchitectureOption: "x64"
+          jdkSourceOption: "PreInstalled"
+
+      - script: brew install coreutils ninja
+        displayName: Install coreutils and ninja
+
+      - script: |
+          python ./tools/gen_selectedops.py ./tools/android/package_ops.config
```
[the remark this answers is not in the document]
wejoncy: we want to build all ops and set OCOS_ENABLE_SELECTED_OPLIST=OFF, is that right?

.pipelines/templates/android_shared_lib_build.yml
```diff
@@ -0,0 +1,70 @@
...
+- name: AndroidABI
+  displayName: android abi,[arm64-v8a, x86_64, X86, armeabi-v7a]
+  values:
+  - 'x86_64'
+  - 'x86'
+  - 'arm64-v8a'
+  - 'armeabi-v7a'
+  type: string
+- name: Android_arch
+  displayName: same android abi, but in valid var
```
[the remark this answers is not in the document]
wejoncy: there is a hyphen in abi 'armeabi-v7a'. I have to use armeabi_v7a.
Will use a more clean name.

.pipelines/templates/c-api-android-cpu.yml
```diff
@@ -0,0 +1,100 @@
...
+  type: boolean
+  default: false
+stages:
+- stage: Android_Java_API_AAR_Packaging_Full
+  dependsOn: []
+  jobs:
+  - template: android_shared_lib_build.yml
+    parameters:
+      AndroidABI: 'x86_64'
+      Android_arch: 'x86_64'
```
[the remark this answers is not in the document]
wejoncy: ditto.  arm64-v8a vs arm64_v8a

.pipelines/templates/c-api-cpu.yml
```diff
@@ -0,0 +1,269 @@
+parameters:
+- name: RunOrtExtensionsTests
+  displayName: Run Tests?. There is no test exist for now
```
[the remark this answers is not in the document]
wejoncy: I think we don't have to run the test here, right? because we should run these tests somewhere, so I removed all tests in pipeline.

.pipelines/templates/c-api-cpu.yml
```diff
@@ -0,0 +1,269 @@
...
+      packageName: x86
+
+  - template: c-api-windows-cpu.yml
+    parameters:
+      DoCompliance: ${{ parameters.DoCompliance }}
+      DoEsrp: ${{ parameters.DoEsrp }}
+      stage_name_suffix: CPU_arm_${{ parameters.BuildVariant }}
+      buildArch: x64
+      msbuildPlatform: arm
+      packageName: arm
```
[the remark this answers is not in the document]
wejoncy: we are building all artifacts in a x64 machine. so I leave it here for reading.

tools/ci_build/generate_nuspec_for_native_nuget.py
wejoncy: I am not very clear what does this doing?
```
        # Process props file
        source_props = os.path.join(
            args.sources_path, "csharp", "src", "Microsoft.ML.OnnxRuntime", "targets", "netstandard", "props.xml"
        )
        target_props = os.path.join(
            args.sources_path,
            "csharp",
            "src",
            "Microsoft.ML.OnnxRuntime",
            "targets",
            "netstandard",
            args.package_name + ".props",
        )
```
Can I remove them all?

tools/ci_build/generate_nuspec_for_native_nuget.py
wejoncy: I don't have these files for now. So I would either remove this or have a condition to pass by this branch.

.pipelines/templates/c-api-linux-cpu.yml
wejoncy: Added in the seperate CXXFLAG

.pipelines/templates/android_shared_lib_build.yml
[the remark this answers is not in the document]
skottmckay: Currently the request is to build the libraries for the nuget package with `-DOCOS_ENABLE_CV2=OFF -DOCOS_ENABLE_BLINGFIRE=OFF` rather than selected ops list.

.pipelines/templates/c-api-cpu.yml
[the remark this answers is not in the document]
skottmckay: No. Do we need this parameter? If so the displayName is confusing as it seems to say we have no tests. 

I would suggest the displayName should just describe the parameter. You could add a comment above the default value of false explaining why it's false.

.pipelines/templates/android_shared_lib_build.yml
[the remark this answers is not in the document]
wejoncy: Got it.

tools/android/build_aar.py
wejoncy: The pipeline before is:
```
x86   (all in sequential)
  |
x64
 |
armv7
 |
armv8
  |
aar
```
and Now
```
x86     x64   armv7   armv8  (in parallel)
\          \         /            /
      \           |         /
              aar
```

To support build sharelib in different platform  in parallel, this script has to be support build lib/aar/pack_only

.pipelines/templates/c-api-ios-cpu.yml
wejoncy: Renamed to build-package-for-xxx.yml

tools/ci_build/windows/extract_nuget_files.ps1
wejoncy: unpack to ($Env:BUILD_BINARIESDIRECTORY\nuget-ext)

tools/ios/build_xcframework.py
```diff
@@ -179,6 +191,12 @@ def parse_args():
         required=True,
         help="Path to output directory.",
     )
+    parser.add_argument(
+        "--build_target",
+        default="xcframework",
+        choices=["lib", "xcframework","pack_xcframework"],
```
skottmckay, line 197:
```suggestion
        choices=["lib", "xcframework", "pack_xcframework"],
```

tools/ci_build/windows/extract_nuget_files.ps1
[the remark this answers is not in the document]
wejoncy: sure.

.pipelines/templates/mac-build-step-with-cache.yml
```diff
@@ -0,0 +1,60 @@
...
+  type: boolean
+  default: false
+
+- name: BuildStep
+  type: stepList
+
+# if WithCahe is false, the following parameters are not used.
+- name: Today
+  type: string
+  default: ""
```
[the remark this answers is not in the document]
wejoncy: Will remove it given that build is really fast.

.pipelines/templates/build-package-for-nuget.yml
```diff
@@ -0,0 +1,234 @@
...
+      displayName: Use Nuget 6.2.1
+      inputs:
+        versionSpec: 6.2.1
+
+    - task: PowerShell@2
+      displayName: 'Updating NuSpec'
+      inputs:
+        targetType: 'inline'
+        script: |
+          mv $(Build.BinariesDirectory)/nuget-ort-ext  $(Build.SourcesDirectory)/nuget-artifacts-ext
```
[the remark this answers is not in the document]
wejoncy: I am assuming `move` is a fast command. If we copy `nuget` folder from source-path to binary-path in reverse, we have to carefully handle the other files in source repo.

.pipelines/templates/android_shared_lib_build.yml
```diff
@@ -0,0 +1,66 @@
+# packaging pipeline for onnxruntime-extensions Android AAR package
```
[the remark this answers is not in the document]
wejoncy: Can do it in the next PRs. Setting ${{parameters.AdditionalBuildFlags}} to different flags `-DOCOS_ENABLE_CV2=OFF -DOCOS_ENABLE_BLINGFIRE=OFF` or `OCOS_ENABLE_SELECTED_OPLIST=ON`.

.pipelines/templates/build-package-for-android-aar.yml
```diff
@@ -0,0 +1,100 @@
...
+        targetType: 'inline'
+        script: |
+          set -e -x
+          python ./tools/android/build_aar.py \
+              --output-dir $(Build.BinariesDirectory)/android_aar \
+              --build_target pack_aar       \
+              --config $(buildConfig) \
+              --cmake-extra-defines OCOS_ENABLE_SELECTED_OPLIST=ON
+          VERSION=$(cat ./version.txt)
+          AAR_PATH="$(Build.BinariesDirectory)/android_aar/aar_out/$(buildConfig)/com/microsoft/onnxruntime/onnxruntime-extensions-android/${VERSION}/onnxruntime-extensions-android-${VERSION}.aar"
```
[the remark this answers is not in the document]
wejoncy: We have to keep it for now.
```
- task: DownloadPipelineArtifact@2
      inputs:
        artifact: 'onnxruntime-extensions-android-x86_64'
        targetPath: '$(Build.BinariesDirectory)/android_aar'
    - task: DownloadPipelineArtifact@2
      inputs:
        artifact: 'onnxruntime-extensions-android-x86'
        targetPath: '$(Build.BinariesDirectory)/android_aar'
    - task: DownloadPipelineArtifact@2
      inputs:
        artifact: 'onnxruntime-extensions-android-armeabi-v7a'
        targetPath: '$(Build.BinariesDirectory)/android_aar'
    - task: DownloadPipelineArtifact@2
      inputs:
        artifact: 'onnxruntime-extensions-android-arm64-v8a'
        targetPath: '$(Build.BinariesDirectory)/android_aar'
```
In the .pipelines/templates/build-package-for-android-aar.yml. 
`onnxruntime-extensions-android-x86_64` is a android package.`onnxruntime-extensions-x86_64` is a windows one?

.pipelines/templates/build-package-for-macosx.yml
```diff
@@ -0,0 +1,62 @@
...
+    parameters:
+      MacosArch: 'x86_64'
+      AdditionalBuildFlags: ${{ parameters.AdditionalBuildFlags }} 
+      WithCache: ${{ parameters.WithCache }}
+  # error   _png_do_read_transformations in liblibpng.a(pngrtran.c.o) not found for architecture arm64
+  # solution https://github.com/davisking/dlib/pull/2664/files
+
+  #- template: mac-cpu-packing-jobs.yml
+  #  parameters:
+  #    MacosArch: 'arm64'
```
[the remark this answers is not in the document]
wejoncy: I didn't find a way to fix it for now, so just workaround to comment it out.

tools/ios/build_xcframework.py
```diff
@@ -179,6 +191,20 @@ def parse_args():
         required=True,
         help="Path to output directory.",
     )
+    # Same as how android works
+    # This one is used in the ci pipeline for accelerating the build process,
+    # We have 4 archs to be build. It's in sequence by default, but we can build them in parallel.
+    # The parallel build works as:
+    #   1. build the lib files for each arch in different ci jobs
+    #   2. download all the lib files in tasks
+    #   2. pack the xcframework package in ci task
```
edgchen1, line 200: it's a good idea to allow parallelization of the framework builds. however, this needs some refinement for ios.

there is a framework (for one platform, for one arch), and then a fat framework (for one platform, for all archs), and then an xcframework (for all platform/archs).

so for example, you'd need to build each framework for one platform (e.g., iphonesimulator/x86_64 and iphonesimulator/arm64), then put all those together in one fat framework with `lipo` for the iphonesimulator platform. then, put the fat frameworks for each platform (iphonesimulator and iphoneos) together in one xcframework with `xcrun xcodebuild -create-xcframework ...`.

tools/ios/build_xcframework.py
```diff
@@ -179,6 +191,20 @@ def parse_args():
...
+    parser.add_argument(
+        "--build_target",
+        default="xcframework",
+        choices=["lib", "xcframework", "pack_xcframework"],
+        help="Build target. 'xcframework' builds the whole package. 'lib' builds the libraries only. \
+            'pack_xcframework' only does pack xcframework from existing lib files.",
```
edgchen1, line 206: nit: "target" has multiple meanings and it's not obvious to me what it means here.

just an idea for another name: `--mode` option with values `build_platform_arch_frameworks_only`, `pack_xcframework_only`, `build_xcframework` (default)

similar comment for build_aar.py

.pipelines/templates/android_shared_lib_build.yml
```diff
@@ -0,0 +1,70 @@
...
+        inputs:
+          targetType: 'inline'
+          script: |
+            set -e -x
+            python ./tools/android/build_aar.py \
+                --output-dir $(Build.BinariesDirectory)/android_so \
+                --build_target so       \
+                --abi ${{ parameters.AndroidABI }}       \
+                --config $(buildConfig) \
+                --cmake-extra-defines ${{parameters.AdditionalBuildFlags}}
```
edgchen1, line 55: from the name, it's not obvious that "AdditionalBuildFlags" are starting the value of `--cmake-extra-defines`. one could pass additional, unrelated options too after that.

maybe call it "CmakeExtraDefines" and pass it like `--cmake-extra-defines "${{ parameters.CmakeExtraDefines }}"`?

similar comment for other pipeline files

.pipelines/templates/ios_shared_lib_build.yml
```diff
@@ -0,0 +1,69 @@
+# packaging pipeline for onnxruntime-extensions ios cocoapods package
+parameters:
+- name: Platform
+  displayName: ios Platform
+  type: string
+- name: ios_arch
```
edgchen1, line 6: nit: can we have consistent naming for parameter names? e.g. `IosArch`

.pipelines/templates/ios_shared_lib_build.yml
edgchen1, line 1: on the name `ios_shared_lib_build.yml` - it's actually a static framework, not a shared library. consider `ios_framework_build.yml`

edgchen1: also can we have consistently named files? e.g., either all `my-template-file.yml` or `my_template_file.yml`

tools/ios/build_xcframework.py
edgchen1: FYI working on this in a branch: https://github.com/microsoft/onnxruntime-extensions/tree/edgchen1/split_xcframework_build

.pipelines/templates/build-package-for-windows.yml
```diff
@@ -0,0 +1,94 @@
...
+  steps:
+    - checkout: self
+      clean: true
+      submodules: none
+
+    - task: BatchScript@1
+      displayName: 'build onnxruntime-extensions'
+      inputs:
+        filename: '.\build.bat'
+        args: "-A ${{parameters.msbuildPlatform}} -DOCOS_ENABLE_CV2=OFF -DOCOS_ENABLE_BLINGFIRE=OFF"
```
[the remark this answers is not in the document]
wejoncy: refined to use "build_lib.bat/sh".

tools/android/build_aar.py
skottmckay, line 1: API level needs to be 21.

https://github.com/microsoft/onnxruntime-extensions/blob/5b89b79cc4de9e2174a11b3337c5d974345e9e62/tools/android/build_aar.py#LL169C1-L174C6

Otherwise you get this due to the mismatch with the ORT AARs minimum

Severity    Code    Description    Project    File    Line    Suppression State
Error    AMM0000    
    uses-sdk:minSdkVersion 21 cannot be smaller than version 24 declared in library D:\src\github\onnxruntime-inference-examples\mobile\examples\Maui\MauiVisionSample\MauiVisionSample\obj\Debug\net6.0-android\lp\152\jl\AndroidManifest.xml as the library might be using APIs not available in 21
    Suggestion: use a compatible library with a minSdk of at most 21,
        or increase this project's minSdk version to at least 24,
        or use tools:overrideLibrary="ai.onnxruntime.extensions" to force usage (may lead to runtime failures)

.pipelines/nuget.yml
```diff
@@ -1,107 +1,33 @@
...
+
+trigger:
+  branches:
+    exclude:
+      - "*"
+
+pr:
+  branches:
+    include:
+      - "*"
```
edgchen1, line 25: maybe running this for every PR is too frequent?

also I think for other pipelines we just specify the triggers in the web UI instead of the yml file. could do that for this as well to be consistent.

.pipelines/templates/build-package-for-macosx.yml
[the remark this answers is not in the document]
wejoncy: Sure.

tools/android/build_aar.py
wejoncy: Fixed.

.pipelines/nuget.yml
wejoncy: Yep

skottmckay (dismissed):

PR #409, opened 2023-04-26
Add nuget README and remove ort lib references from props

sayanshaw24:


nuget/targets/netstandard/Microsoft.ML.OnnxRuntime.Extensions.props
```diff
@@ -1,17 +1,5 @@
...
-
-  <ItemDefinitionGroup Condition="'$(PlatformTarget)' == 'x64' OR ('$(PlatformTarget)' == 'AnyCPU' AND '$(Prefer32Bit)' != 'true')">
-	<Link>
-	  <AdditionalDependencies>$(MSBuildThisFileDirectory)../../runtimes/win-x64/native/onnxruntime.lib;%(AdditionalDependencies)</AdditionalDependencies>
-	</Link>
-  </ItemDefinitionGroup>
-
-  <ItemDefinitionGroup Condition="'$(PlatformTarget)' == 'x86' OR ('$(PlatformTarget)' == 'AnyCPU' AND '$(Prefer32Bit)' == 'true')">
-	<Link>
-	  <AdditionalDependencies>$(MSBuildThisFileDirectory)../../runtimes/win-x86/native/onnxruntime.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
wenbingl, line 12: should it be ortextensions.lib?

wenbingl (approved):

nuget/targets/netstandard/Microsoft.ML.OnnxRuntime.Extensions.props
sayanshaw24: I believe RelWithDebInfo does not have a ortextensions.lib, that was in Release. Regardless, this was the old reference to the ort lib which we removed from the package and so we are removing it from the props file too.

wenbingl (approved):

skottmckay (approved):

PR #411, opened 2023-04-27
Infrastructure needed for nuget package that works on mobile platforms.

skottmckay:
Miscellaneous changes for the nuget package to be able to be created for mobile platforms:

- add targets/props for the additional platforms
- update nuspec with the various entries required
  - commented out temporarily pending filling in the paths from the CI where the builds will come from
- update android build to use shared library as Xamarin/MAUI load that
- update android build to target API level 21
  - matches ORT package. if there's a mismatch the build complains about incompatibility


nuget/targets/netstandard/Microsoft.ML.OnnxRuntime.Extensions.props
```diff
@@ -1,30 +1,7 @@
...
-	</Link>
-  </ItemDefinitionGroup>
-	
-  <PropertyGroup>
-    <EnginePlatform Condition="'$(Platform)' == 'Win32'">x86</EnginePlatform>
-    <EnginePlatform Condition="'$(Platform)' != 'Win32' AND '$(Platform)' != 'ARM64'">$(Platform)</EnginePlatform>
-  </PropertyGroup>
-
-  <PropertyGroup>
-    <OnnxRuntimeBinary>$(MSBuildThisFileDirectory)../../runtimes/win-$(EnginePlatform)/native/ortextensions.dll</OnnxRuntimeBinary>
```
skottmckay, line 22: Nothing seems to use OnnxRuntimeBinary or EnginePlatform.

nuget/NativeNuget.nuspec
```diff
@@ -31,15 +31,53 @@
...
+		-->
+		<!-- Windows and Linux via netstandard -->
 		<file src="..\out\Windows\bin\RelWithDebInfo\ortextensions.dll" target="runtimes\win-x64\native" />
+		<!-- update source paths to match the CI as the builds for the various platforms are added
+		<file src="???-win-x86\lib\ortextensions.dll" target="runtimes/win-x86/native"/>
+		<file src="???-win-arm\lib\ortextensions.dll" target="runtimes/win-arm/native"/>
+		<file src="???-win-arm64\lib\ortextensions.dll" target="runtimes/win-arm64/native"/>
+		
+		<file src="???-linux-aarch64\lib\libortextensions.so" target="runtimes/linux-arm64/native"/>		
+		<file src="???-linux-x64\lib\libortextensions.so" target="runtimes/linux-x64/native"/>
```
skottmckay, line 51: @wejoncy can we fill in the various src values with the paths from the CI and check that in so the script updates the checked in file instead of generating it all dynamically?

wejoncy: Yes. Would [this](https://github.com/microsoft/onnxruntime-extensions/pull/408/files#diff-42656c0cafc8b6f749f5d61e541b4d422803d710ac6024293bce190b33ea93b7R91) be good?

skottmckay: Do we need to always add files dynamically vs hardcoding all the ones the CI will include?

Trying to understand what the content of `generate_file_list` equates to in terms of actual entries in the nuspec is virtually impossible. If we check in a file with all the `file` entries it's really easy for a new developer to understand what is going into the nuget package and how that is defined, so I'd really like to do that if possible. 

AFAIK there are 2 scenarios:

On the CI we assemble artifacts and extract them, with the CI controlling the directory names etc. that artifacts are extracted to. if we can use relative paths and the `file` entries are included in the checked in NativeNuget.nuspec a user could technically re-create that directory structure locally if they really wanted as they could see exactly which files are required. 

The other scenario is a user wanting to test some changes they're making locally. Typically they would just be testing one platform, but it's possible they would want to do multiple (e.g. Android and iOS). In this case the src paths are totally dependent on how the user did those individual builds. e.g. did they use build.bat or build_lib.bat for example. Due to that we really need them to provide paths for each individual library they want to add. For that to work we also need something to figure out all the things they're not providing to disable related pieces. e.g. no android AAR means the targets for android need to be excluded. Question is whether it's worth complicating this to allow a developer to update the `files` entries via a script. As it's relatively simple to edit the nuspec if all the possible entries are there to start with I'm not sure it is.

And if a user wanted to test more than a couple of platforms it would be easier for them to run the nuget packaging pipeline on their branch to get a nuget package to test with from that. 

Definitely worth the script being able to update versions and commit ids etc. though.

wejoncy: Let's do only update work in the python script.

tools/android/build_aar.py
```diff
@@ -145,7 +145,7 @@ def path_from_env_var(env_var: str):
     parser.add_argument(
         "--api-level",
         type=int,
-        default=24,
+        default=21,
         help="Android API Level. E.g., 24.",
```
wejoncy, line 149: why do we lower to 21? 
nit: update the help.

nuget/targets/netstandard/Microsoft.ML.OnnxRuntime.Extensions.props
```diff
@@ -1,30 +1,7 @@
...
 
-  <ItemDefinitionGroup Condition="'$(PlatformTarget)' == 'x64' OR ('$(PlatformTarget)' == 'AnyCPU' AND '$(Prefer32Bit)' != 'true')">
-	<Link>
-	  <AdditionalDependencies>$(MSBuildThisFileDirectory)../../runtimes/win-x64/native/onnxruntime.lib;%(AdditionalDependencies)</AdditionalDependencies>
-	</Link>
-  </ItemDefinitionGroup>
-
-  <ItemDefinitionGroup Condition="'$(PlatformTarget)' == 'x86' OR ('$(PlatformTarget)' == 'AnyCPU' AND '$(Prefer32Bit)' == 'true')">
-	<Link>
-	  <AdditionalDependencies>$(MSBuildThisFileDirectory)../../runtimes/win-x86/native/onnxruntime.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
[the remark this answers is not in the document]
sayanshaw24: this is removed in latest

nuget/targets/netstandard/Microsoft.ML.OnnxRuntime.Extensions.props
skottmckay: It's just a custom property but nothing else appears to use it so the same as an unused variable.

nuget/targets/net6.0-ios/Microsoft.ML.OnnxRuntime.Extensions.targets
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Condition="('$(OutputType)'!='Library' OR '$(IsAppExtension)'=='True')">
+    <NativeReference Include="$(MSBuildThisFileDirectory)runtimes\ios\native\onnxruntime_extensions.xcframework">
```
edgchen1, line 4:
```suggestion
    <NativeReference Include="$(MSBuildThisFileDirectory)..\..\runtimes\ios\native\onnxruntime_extensions.xcframework">
```

nuget/targets/xamarinios10/Microsoft.ML.OnnxRuntime.Extensions.targets
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Condition="('$(OutputType)'!='Library' OR '$(IsAppExtension)'=='True')">
+    <NativeReference Condition="'$(Platform)' == 'iPhoneSimulator'" Include="$(MSBuildThisFileDirectory)..\runtimes\ios\native\onnxruntime_extensions.xcframework\ios-arm64_x86_64-simulator\onnxruntime_extensions.framework">
```
edgchen1, line 4:
```suggestion
    <NativeReference Condition="'$(Platform)' == 'iPhoneSimulator'" Include="$(MSBuildThisFileDirectory)..\..\runtimes\ios\native\onnxruntime_extensions.xcframework\ios-arm64_x86_64-simulator\onnxruntime_extensions.framework">
```

nuget/targets/xamarinios10/Microsoft.ML.OnnxRuntime.Extensions.targets
```diff
@@ -0,0 +1,21 @@
...
+      <Kind>Framework</Kind>
+      <IsCxx>True</IsCxx>
+      <SmartLink>True</SmartLink>
+      <ForceLoad>True</ForceLoad>
+      <LinkerFlags>-lc++</LinkerFlags>
+      <WeakFrameworks>CoreML</WeakFrameworks>
+    </NativeReference>
+    <NativeReference Condition="'$(Platform)' == 'iPhone'" Include="$(MSBuildThisFileDirectory)..\runtimes\ios\native\onnxruntime_extensions.xcframework\ios-arm64\onnxruntime_extensions.framework">
```
edgchen1, line 12:
```suggestion
    <NativeReference Condition="'$(Platform)' == 'iPhone'" Include="$(MSBuildThisFileDirectory)..\..\runtimes\ios\native\onnxruntime_extensions.xcframework\ios-arm64\onnxruntime_extensions.framework">
```

tools/android/build_aar.py
skottmckay: Mentioned in the review description. ORT uses 21. If you attempt to use both the ORT and extensions packages the build complains about incompatibility.

skottmckay:
Integrated to #408

PR #414, opened 2023-05-02
Fix Secure Supply Chain Analysis Warning in PR pipeline

sayanshaw24:
We were getting a config analysis warning because the ClipTokenizer C# proj is not referenced in the Nuget.Config file, but it shouldn't need to be : [Pipelines - Run 20230501.4 logs (visualstudio.com)](https://aiinfra.visualstudio.com/ONNX%20Runtime/_build/results?buildId=304623&view=logs&j=096205eb-4599-56da-affa-13246cc49eac&t=e04a6eda-03ec-58f9-2d92-da0ce2c1a83e&l=25)

Removing packageSources from the config ignores the warning and solves the issue for manually triggered pipelines but not PR automated ones - however we could also simply remove the config file if it not needed by the CIs.


snnn (approved):

test/NuGet.config
wenbingl, line 1: Can you also check the file history to see why this file was added?

wenbingl (approved):

test/NuGet.config
sayanshaw24: Yeah, I also had this question - it was added by Scott to install GoogleTestAdapter but there's a package.config to install it so it is not needed, I checked with him and he said it should be fine to delete if CIs don't throw an issue.

.sscignore
```diff
@@ -0,0 +1,3 @@
+{
```
snnn, line 1: Does this file have any impact?

sayanshaw24: not at the minute, there are directions for handling an exception to ignore warnings that are not an actual problem here: https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/secure-supply-chain/project-artemis/central-feed-services-cfs, there is form which needs to be filled out which I have now submitted. Once it is looked at this should work to suppress the warnings.

wenbingl (approved):

PR #415, opened 2023-05-03
Add new required pre/post processing ops to Android and iOS packages.

skottmckay:
Add 
- BpeDecoder for Whisper
- ClipTokenizer for StableDiffusion
- DrawBoundingBoxes for object detection models


wenbingl (approved):

edgchen1 (approved):

tools/ios/package_ops.config
```diff
@@ -1 +1 @@
-com.microsoft.extensions;1;DecodeImage,EncodeImage,BertTokenizer
+com.microsoft.extensions;1;DecodeImage,EncodeImage,BertTokenizer,ClipTokenizer,BpeDecoder,DrawBoundingBoxes
```
wenbingl, line 1: Not interesting in AudioDecoder op?

skottmckay: How big is it?

wenbingl: x64 dll delta is around 250K, compressed size less than 100k.
I think Android and iOS CI pipelines by default include all custom ops, only release pipelines are using selected op config file.

wenbingl: All custom op input/output are described in this Python class. I knew it would be better there are some docs for that, probably we can auto-gen it from the source.

https://github.com/microsoft/onnxruntime-extensions/blob/2fa0b710eac5d39088402e369965cad126575d71/onnxruntime_extensions/_cuops.py#LL393C10-L393C10

Most of custom op unit tests are written in Python, Android/iOS only builds the source code to make sure it be compatible with the compiler and dependencies. Few of them have some C++ tests, not sure how convenient it is to run on Android/iOS.

skottmckay: Have all the open source licenses for things used by the AudioDecoder op both directly and indirectly been reviewed and added to the open source registrations for extensions?

tools/ios/package_ops.config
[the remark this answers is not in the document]
skottmckay: Probably due to replicating the ORT setup where coreml and nnapi support different ops. Could be combined. Will do in next PR.

tools/ios/package_ops.config
skottmckay: Will discuss offline and create separate review for any additional ops. May also want to include more tokenizers.

PR #424, opened 2023-05-05
Improve description of model-name arg

natke:


wenbingl (approved):

natke:
Actually that is not quite right either. Will close and submit a new PR

PR #426, opened 2023-05-05
Add nuget.org publish version option

sayanshaw24:


tools/ci_build/update_nuspec_for_native_nuget.py
```diff
@@ -21,6 +21,11 @@ def update_nuspec(args):
         if package_item.tag == "version" and args.package_version:
             if args.is_release_build:
                 package_item.text = args.package_version
+            elif args.is_for_nuget_publish:
+                # Update prefix and postfix below as per NuGet prelease guidelines and team discussions
+                prefix = "alpha"
```
wenbingl, line 26: can 'alpha.1' be a string parameter?

wenbingl (approved):

tools/ci_build/update_nuspec_for_native_nuget.py
sayanshaw24: yeah - wanted to possibly automate this in the future (if we can decide on better naming criterion like with build version) and keep the prerelease prefix (alpha/beta) and postfix separate. will make a separate PR for this once we decide on that naming criterion.

PR #432, opened 2023-05-09
Minor cmake updates

skottmckay:
Make sure LINUX variable is set as it's used in CMakeLists.txt but is… only defined in recent versions of cmake.

Update ext_ortlib.cmake so it doesn't break when cross-compiling for Android on Windows by defaulting to a build even though it can't be used with Android. Need to address the unit testing gap for Android/iOS separately.


CMakeLists.txt
```diff
@@ -7,6 +7,13 @@ if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "Choose build type: Debug Release RelWithDebInfo." FORCE)
 endif()
 
+# LINUX variable is set from 3.25 on
+if (CMAKE_VERSION VERSION_LESS "3.25")
```
wenbingl, line 11: Are we able to upgrade minimum cmake version to be 3.25 in iOS/Android now?

wenbingl (approved):

CMakeLists.txt
skottmckay: The Android CI is using 3.26.1 and the iOS CI is using 3.25.0. 

What was the reason android or ios blocked a cmake upgrade previously?

skottmckay: Updated to 3.25 as minimum but not clear why the macos builds are failing. Based on other successful CI runs the macos build is using cmake 3.26, so the change shouldn't matter. 

https://aiinfra.visualstudio.com/Lotus/_build/results?buildId=306993&view=logs&j=cdb80bd2-8b48-5122-dfc6-e0f4b7485e64&t=793ef336-d55c-5c67-817c-75d71d7f2bbd&l=314
`--     CMake:                       3.26.3`

Error is this:

> 2023-05-11T06:27:58.7961650Z CMake Error at CMakeLists.txt:151 (message):
2023-05-11T06:27:58.7963680Z   CMAKE_FIND_FRAMEWORK is not NEVER

What's the reason CMAKE_FIND_FRAMEWORK must be NEVER? For mac I would have expected it was fine to use the system framework, and cmake defaults to using FIRST in that case: 

https://github.com/Kitware/CMake/blob/13e46189c7f3b39a26e9ca689bc029b7061d26a7/Modules/Platform/Darwin.cmake#L138-L141

```cmake
# default to searching for frameworks first
if(NOT DEFINED CMAKE_FIND_FRAMEWORK)
  set(CMAKE_FIND_FRAMEWORK FIRST)
endif()
```

However the CMake default hasn't changed any time recently. Same in 3.25.0 as 3.26.4 so how a build that appears to be using CMake 3.26 fails when the minimum cmake version changes to 3.25 is a bit of a mystery.

wenbingl: One consideration is to be equal of less than ORT cmake version requirements due to combined source build.
CMAKE_FIND_FRAMEWORK can be updated if CI pipeline passes.

cmake/ext_ortlib.cmake
```diff
@@ -14,21 +14,23 @@ else()
   if(CMAKE_HOST_APPLE)
     set(ONNXRUNTIME_URL "v${ONNXRUNTIME_VER}/onnxruntime-osx-universal2-${ONNXRUNTIME_VER}.tgz")
   elseif(CMAKE_HOST_WIN32)
```
snnn, line 16: Actually, here you should not use the "HOST" word.

"Host" mean the environment that is used for running the compilers.
"Target" means the environment that will be used for running the built binaries. 

So, if your cross-compile a Win32 binary on a Linux machine, your host environment is Linux, your target environment is Windows. And you want to link your application to Windows libraries instead of Linux libraries. 

That's why in ORT's code usually we use [WIN32](https://cmake.org/cmake/help/latest/variable/WIN32.html) instead of CMAKE_HOST_WIN32. 

See https://github.com/microsoft/onnxruntime/blob/main/docs/cmake_guideline.md#basics-of-cross-compiling for more information about cross-compiling. 

You don't have to support cross-compiling, but it's better to use the right macros.

CMakeLists.txt
snnn: > One consideration is that cmake version here should be equal of less than ORT cmake version requirements due to combined source build.

You are free to choose your requirements. If ORT requires cmake version >=A  and ORT-ext requires cmake version >=B, the combined build will require cmake version >= max(A,B). That's it.

cmake/ext_ortlib.cmake
skottmckay: FWIW this is used to select a pre-built ORT binary for use in C++ unit tests. Currently the setup doesn't support running those when cross-compiling so we wouldn't actually use the ORT library selected in that case. 

We have some separate Android and iOS testing but that uses the published ORT packages directly when building the test app (and it's a test app rather than a unit test). 

I can update to use WIN32/APPLE to be more correct though

CMakeLists.txt
skottmckay: ORT is using 3.25 or higher for everything except ROCm.

Re: build failure when minimum version is 3.25, I did some builds with this extra logging

```cmake
if (CMAKE_FIND_FRAMEWORK)
  message(STATUS "CMAKE_FIND_FRAMEWORK is ${CMAKE_FIND_FRAMEWORK} prior to us calling set. CMAKE_VERSION=${CMAKE_VERSION}")
endif()

set(CMAKE_FIND_FRAMEWORK NEVER CACHE STRING "...")

if(NOT "${CMAKE_FIND_FRAMEWORK}" STREQUAL "NEVER")
  message(FATAL_ERROR "CMAKE_FIND_FRAMEWORK is ${CMAKE_FIND_FRAMEWORK} not NEVER.")
endif()
```

With a build with min version of 3.20, CMAKE_FIND_FRAMEWORK has a default value and the `set` updates it as the only output is:

`-- CMAKE_FIND_FRAMEWORK is FIRST prior to us calling set. CMAKE_VERSION=3.26.3`

With a build with min version of 3.25, it also has a default value but the `set` doesn't update the value. Both builds have an actual cmake version of 3.26.3

```
-- CMAKE_FIND_FRAMEWORK is FIRST prior to us calling set.  CMAKE_VERSION=3.26.3
CMake Error at CMakeLists.txt:155 (message):
  CMAKE_FIND_FRAMEWORK is FIRST not NEVER.
```

Looks like it's due to a [policy](https://cmake.org/cmake/help/latest/policy/CMP0126.htm) that was added in 3.21 which appears to be applied based on the minimum cmake version rather than actual. Builds fine if I add ` cmake_policy(SET CMP0126 OLD)` however that doesn't seem like a good approach going forward given it's using the old version of the policy and that is explicitly deprecated. I tried adding `FORCE` to the `set` didn't help. Setting regular and cache variables does work.

build.bat
```diff
@@ -16,11 +16,12 @@ IF "%VisualStudioVersion:~0,2%" == "16" GOTO :START_BUILD
 set GENERATOR="Visual Studio 17 2022"
 
 :START_BUILD
-set cmake_exe="%VSINSTALLDIR%Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe"
+REM Add path to VS cmake as the fallback.
+set PATH=%PATH%;"%VSINSTALLDIR%Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe"
```
skottmckay, line 20: This and the change to setup.py work in the CIs so that the cmake version that is explicitly installed and added to the path is preferred.

If we want to hardcode defaulting to the VS cmake we can't update the required cmake version.

wenbingl: Using a explicit cmake is intentional since cmake installed by pip are not stable enough which I encountered before.  Also,  this change may bring some uncertainty of which cmake was used to build. it will bring some difficulty on checking some CI build issues.
To override cmake, we may use some environment variable to do that. 
Did you encounter any case that VS/cmake has to be replaced? VS/cmake actually keeps updating and it looks like a LTS version for Windows platform.

setup.py
```diff
@@ -107,11 +107,11 @@ def build_cmake(self, extension):
...
         ]
         cmake_exe = 'cmake'
-        # unlike Linux/macOS, cmake pip package on Windows fails to build some 3rd party dependencies.
-        # so we have to use the cmake installed from Visual Studio.
+
         if os.environ.get(VSINSTALLDIR_NAME):
-            cmake_exe = os.environ[VSINSTALLDIR_NAME] + \
-                        'Common7\\IDE\\CommonExtensions\\Microsoft\\CMake\\CMake\\bin\\cmake.exe'
+            # Add fallback path for VS cmake. Prefer system install if available via existing PATH entry.
+            sys.path.append(os.environ[VSINSTALLDIR_NAME] +
```
edgchen1, line 113: I think `sys.path` is different from environment variable `PATH` - we probably want to update the latter, right?
https://docs.python.org/3/library/sys.html#sys.path

build.bat
skottmckay: Yeah - the win32.wheel build breaks as the VS it's using only has cmake 3.20. The CI itself is has more a recent cmake installed.

wenbingl: I checked VS 2022 installation, it is 3.25.1, 
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin>cmake --version
cmake version 3.25.1-msvc1

is 3.20 due to VS 2019?

If it seeks in general %PATH% for cmake, it actually uses cmake from pip package because of cmake dependency defined in pyproject.toml. And this cmake edition is not good for official build.

skottmckay: > 2023-05-15T04:43:09.6713775Z   "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe" -S C:\a\_work\1\s -B .scb\temp.win-amd64-cpython-37\Release -DCMAKE_LIBRARY_OUTPUT_DIRECTORY=C:\a\_work\1\s\.scb\lib.win-amd64-cpython-37\onnxruntime_extensions -DOCOS_BUILD_PYTHON=ON -DOCOS_EXTENTION_NAME=_extensions_pydll.cp37-win_amd64.pyd -DCMAKE_BUILD_TYPE=Release
2023-05-15T04:43:09.6714176Z   -- Building for: Visual Studio 16 2019
2023-05-15T04:43:09.6714372Z   CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
2023-05-15T04:43:09.6714570Z     CMake 3.25 or higher is required.  You are running version
2023-05-15T04:43:09.6714735Z     3.20.21032501-MSVC_2

It's using VS2019. Did you want to change that so it uses VS 2022?

wenbingl: Yes, please.
In get_vsdevcmd.ps1, I selected the latest VS installation if there is multiple VS versions installed.

PR #438, opened 2023-05-13
Update cgmanifest.json and ThirdPartyNotices.txt

sayanshaw24:
## Note: 
Updated most components as per cmake/externals and others as latest git versions and added necessary corresponding changes (for example, opencv 3p components such as openjpeg and libpng locations changed and old git repo URLs were invalid so updated those). Please let me know if this is fine.


wenbingl:
thanks for the update.
gsl and dr_libs were added recently and seems being missed in these files.

wenbingl (approved):

PR #455, opened 2023-05-24
 Add ADO parameter for nuget version suffix

sayanshaw24:


tools/ci_build/update_nuspec_for_native_nuget.py
```diff
@@ -22,9 +22,7 @@ def update_nuspec(args):
             if args.is_release_build:
                 package_item.text = args.package_version
             elif args.is_for_nuget_publish:
-                # Update version_suffix below if publishing to NuGet
-                version_suffix = "beta" # alpha/beta/rc
-                package_item.text = f"{args.package_version}-{version_suffix}"
+                package_item.text = f"{args.package_version}-{args.nuget_version_suffix}"
             else:
                 import datetime
```
wenbingl, line 27: can this suffix merge with id string below to have something like "alpha-20230524.."? and we can remove is_for_nuget_publish ?

sayanshaw24: i agree, that would be cleaner - but what about if we need to build packages that are not part of an official release or a nightly build (such as the ones for Build where Cassie did not want the commit id)? i think it might be convenient to have the option for the future, say for other conferences or possible future partner requirements.

wenbingl: I just check ORT name pattern is "Microsoft.ML.OnnxRuntime 1.16.0-dev-20230515-0158-af6cb2af87"
It looks great, extensions should align with it, can you make sure the below "else" statement to align with that?

And "dev-20230515-0158-af6cb2af87" can be overwritten by "nuget_version_suffix", that is, if "nuget_version_suffix" exists, "is_for_nuget_publish" is always true so we don't need it.

.pipelines/nuget.yml
```diff
@@ -19,6 +19,15 @@ parameters:
...
   default: false
 
+- name: NugetVersionSuffix
+  displayName: Update nuget version suffix (only needed if publishing to nuget.org)
+  type: string
+  default: alpha
+  values:
+    - alpha
+    - beta
+    - rc
```
wenbingl, line 29: can users add a new value here via ADO paramenter?

wenbingl (commented):
This PR is not needed by 0.8 release, we don't need to cherry-pick it, right?

tools/ci_build/update_nuspec_for_native_nuget.py
sayanshaw24: Yep, for example, it created: Microsoft.ML.OnnxRuntime.Extensions.0.8.0-dev-20230522-2354-f3aa93bd.nupkg, so that is aligned with ORT.

I'll change the logic to remove `is_for_nuget_publish`.

.pipelines/nuget.yml
sayanshaw24: not at the minute, i can add that. was originally thinking that might allow for unwanted/accidental inconsistent naming is that fine?

wenbingl (approved):

PR #457, opened 2023-05-26
Unit test being compatible with ONNXRuntime-GPU package, and some cle…

wenbingl:
…an-ups.


sayanshaw24 (approved):
this is great, thanks for adding this!

PR #462, opened 2023-05-30
Ignore inputs missing from vocab in CLIPTokenizer

sayanshaw24:
Updating https://github.com/microsoft/onnxruntime-extensions/pull/369/ by @matheusgomes28 here due to access issues with merge conflict with updates for compatibility with latest and other minor changes.

Also added tests and validated HF behavior is same as recommended by @wenbingl in the old PR.

Previous note from @matheusgomes28:
> Hi all,
> 
> I came around the issue of unrecognised words in the vocab map when using the ClipTokenizer from OnnxExtensions. For example, if an input contains \n or other missing inputs in the vocab, the call to bbpe_tokenizer_->GetEncoding(boundary + "</w>") would cause a crash as we're dereferencing an end iterator there.
> 
> I added some code to ignore the input if it's not present in the vocab file. Let me know if that looks ok, or if it will cause repercussions elsewhere. Happy to discuss alternatives to fixing this crash.


operators/tokenizer/bpe_tokenizer.hpp
```diff
@@ -54,7 +55,11 @@ inline bool IsUnicodeSpace(char32_t ch) {
 }
 
 inline bool IsEmptyUString(const ustring& str) {
-  return std::all_of(str.begin(), str.end(), [](char32_t ch) { return IsUnicodeSpace(ch); });
+  if (str == ustring("\n")) {
```
sayanshaw24, line 58: updated `IsEmptyUString` behavior because it was returning `true` for input with just newline character ("\n") which caused CLIPTokenizer output to be different from HF output.

wenbingl: In that case, should we remove the '\n' from the table in IsUnicodeSpace?

sayanshaw24:
> Thanks for the update, can you also the post a link to the code snippet how HF tokenizer handle the missing tokens?

Certain errors, such as the newline character, are handled by just skipping the character, which I researched is not explicitly noted by HF but is true as per manual testing. Others are handled with exceptions like this: https://github.com/huggingface/transformers/blob/v4.29.1/src/transformers/models/clip/tokenization_clip.py#L272

For all other unknown inputs, the tokenizer replaces it with an unk_token which defaults to `<|endoftext|>`: https://github.com/huggingface/transformers/blob/v4.29.1/src/transformers/models/clip/tokenization_clip.py#L275

In a future PR, we need to look into what explicit exceptions are written for CLIP, handle those, and then replace all other unknowns with an unk_token to replicate HF. Was initially just going to do that but after testing I realized there are exceptions in HF like for the newline character which is not replaced with an unk_token but skipped.

operators/tokenizer/bpe_tokenizer.hpp
```diff
@@ -269,9 +274,14 @@ class VocabData {
     return special_tokens_.SplitBySpecialTokens(input);
   }
 
-  int GetEncoding(const std::string& key) {
+  // Returns {true, token} if key was found in vocab, and {false, -1} otherwise
+  std::pair<bool, int> GetEncoding(const std::string& key) {
     auto it = vocab_map_.find(key);
-    return it->second;
```
wenbingl, line 274: There is already a unknown token id in this class, GetEncoding can return unknown_id if not found, instead of std::pair.

operators/tokenizer/bpe_tokenizer.hpp
sayanshaw24: No, I believe we should not, because it is technically still Unicode space as per CPython logic. Also, behavior is correct for strings with newline as part of it, it was only wrong for when the input string is _just_ "\n".

sayanshaw24:
> Thanks for the update, can you also the post a link to the code snippet how HF tokenizer handle the missing tokens?

https://github.com/huggingface/transformers/blob/v4.29.1/src/transformers/models/clip/tokenization_clip.py#L275

operators/tokenizer/bpe_tokenizer.hpp
```diff
@@ -54,7 +55,29 @@ inline bool IsUnicodeSpace(char32_t ch) {
 }
 
 inline bool IsEmptyUString(const ustring& str) {
```
wenbingl, line 57: If "\n" was handled by whitespace clean-up, is this change still needed?

operators/tokenizer/bpe_tokenizer.hpp
```diff
@@ -54,7 +55,29 @@ inline bool IsUnicodeSpace(char32_t ch) {
...
+    size_t pos = 0;
+    while ((pos = str.find(search, pos)) != ustring::npos) {
+    str.replace(pos, search.length(), replace);
+        pos += replace.length();
+    }
+    return str;
+}
+
+inline void WhiteSpaceClean(ustring& str) {
+    str = ReplaceString(str, ustring("\n"), ustring(" "));
```
wenbingl, line 79: re.replace("\s+", ' ') should use Python IsUnicodeSpace to detect the WS, instead of just one case of ' '

operators/tokenizer/bpe_tokenizer.hpp
```diff
@@ -54,7 +55,29 @@ inline bool IsUnicodeSpace(char32_t ch) {
...
+    str.erase(std::unique(str.begin(), str.end(), BothSpaces), str.end());
```
wenbingl, line 80: I don't remember str.erase can remove multiple iterators, can you confirm this line can remove multiple occurrences of WS sequences

operators/tokenizer/bpe_tokenizer.hpp
sayanshaw24: I changed this - it used to be for "\n", now it handles the case for just a singular space string, which is needed.

operators/tokenizer/bpe_tokenizer.hpp
sayanshaw24: done! updated BothSpaces to use IsUnicodeSpace which will do that.

test/test_cliptok.py
```diff
@@ -112,6 +112,12 @@ def test_tokenizer(self):
         self._run_tokenizer(["9 8 7 6 5 4 3 2 1 0"])
         self._run_tokenizer(["9 8 7 - 6 5 4 - 3 2 1 0"])
         self._run_tokenizer(["One Microsoft Way, Redmond, WA"])
+        self._run_tokenizer(["I can feel the magic, \n can you?"])
+        self._run_tokenizer(["\n Hey Cortana"])
+        self._run_tokenizer(["lower newer \n"])
+        self._run_tokenizer(["\n"])
+        self._run_tokenizer(["Testing multiple      sequences       of spaces"])
+        self._run_tokenizer(["      in the beginning and the end.      "])
```
wenbingl, line 120: can you add a case of a single whitespace also?

sayanshaw24: done!

wenbingl (approved):

PR #466, opened 2023-06-01
upgrade all dependency versions

wenbingl:


.pipelines/ci.yml
```diff
@@ -21,6 +21,10 @@ jobs:
 
     strategy:
       matrix:
+        py311-1150:
+          python.version: '3.11'
+          torch.version: 'pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu'
```
sayanshaw24, line 26: the `pip3 install` is redundant i believe (and is causing the error in extensions.ci) - same comment for other instance of it.

sayanshaw24 (approved):

PR #469, opened 2023-06-11
Es0m/add static runtime linking

es0m:
- current main of onnxruntime-extensions depends on msvc runtime (vcruntime)
- static linkage is desired to reduce downstream dependency
- OCOS_ENABLE_STATIC_LIB is an existing option which only seems to be used for webassembly checks atm
- added add_compile_options(/MT$<$<CONFIG:Debug>:d>) and CMAKE_MSVC_RUNTIME_LIBRARY for static linkage if OCOS_ENABLE_STATIC_LIB is specified


CMakeLists.txt
```diff
@@ -2,10 +2,14 @@ cmake_minimum_required(VERSION 3.25)
 project(onnxruntime_extensions LANGUAGES C CXX)
 
 # set(CMAKE_VERBOSE_MAKEFILE ON)
-if(NOT CMAKE_BUILD_TYPE)
+if(NOT WIN32 AND NOT CMAKE_BUILD_TYPE)
   message(STATUS "Build type not set - using RelWithDebInfo")
   set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "Choose build type: Debug Release RelWithDebInfo." FORCE)
 endif()
+if (OCOS_ENABLE_STATIC_LIB) 
+    add_compile_options(/MT$<$<CONFIG:Debug>:d>)
```
wenbingl, line 10: what's about other build-flavor, like release, relwithdebinfo?

es0m: for all other build types it'll link with /MT only. I'll build with RelWithDebInfo and fix as needed.

CMakeLists.txt
```diff
@@ -2,10 +2,14 @@ cmake_minimum_required(VERSION 3.25)
 project(onnxruntime_extensions LANGUAGES C CXX)
 
 # set(CMAKE_VERBOSE_MAKEFILE ON)
-if(NOT CMAKE_BUILD_TYPE)
+if(NOT WIN32 AND NOT CMAKE_BUILD_TYPE)
   message(STATUS "Build type not set - using RelWithDebInfo")
   set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "Choose build type: Debug Release RelWithDebInfo." FORCE)
 endif()
+if (OCOS_ENABLE_STATIC_LIB)
```
wenbingl, line 9: if (MSVC)

CMakeLists.txt
```diff
@@ -2,10 +2,14 @@ cmake_minimum_required(VERSION 3.25)
 project(onnxruntime_extensions LANGUAGES C CXX)
 
 # set(CMAKE_VERBOSE_MAKEFILE ON)
-if(NOT CMAKE_BUILD_TYPE)
+if(NOT WIN32 AND NOT CMAKE_BUILD_TYPE)
```
wenbingl, line 5: what do you expected on Win32?

wenbingl (commented):
In Win32, there is few usage of using static runtime link ORT custom op API is using dynamic loading a DLL. Now the latest API can support ort-extensions to be a static library, but I think the changes should be more than the changes in this PR.

CMakeLists.txt
es0m: On WIN32, the config is usually set, and there can be multiple configs.

wenbingl:
Can you add a the failed build case that you are fixing in ort-extensions CI pipeline to make sure the static link build won't be broken by any future change?

https://github.com/microsoft/onnxruntime-extensions/blob/1f0c76cefaa9359b9fc74ff4b9ed93214486205a/.pipelines/ci.yml#L260

@snnn , can you review this PR as well?

CMakeLists.txt
es0m: see e.g. https://github.com/microsoft/onnxruntime-extensions/blob/1f0c76cefaa9359b9fc74ff4b9ed93214486205a/build.bat#LL23C56-L23C56

snnn: This is right. CMake's VS generator support multiple configs. There is no need to set one when running cmake. 
See:
https://cmake.org/cmake/help/latest/prop_gbl/GENERATOR_IS_MULTI_CONFIG.html
 
Only these generators support multiple configs

CMakeLists.txt
```diff
@@ -105,6 +105,11 @@ if(NOT CC_OPTIMIZE)
 endif()
 
 if (MSVC)
+  if (OCOS_ENABLE_STATIC_LIB) 
+    add_compile_options(/MT$<$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>:d>)
+    set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>:Debug>")
```
snnn, line 110: You can set the CMAKE_MSVC_RUNTIME_LIBRARY variable from cmake command line.  So this change is unnecessary.

snnn: OCOS_ENABLE_STATIC_LIB means we want to build this repo's code as a static library. However, it doesn't mean they need to static link to VC Runtime.

wenbingl: @es0m-msft, According to the description, this PR is to enable static linking to VCRuntime, but OCOS_STATIC_LIBRARY flag was created for generating the static library of ort-extensions.lib instead of DLL. Can you try @snnn suggestion to see if it meets your requirements?

es0m: OCOS_STATIC_LIBRARY is different from OCOS_ENABLE_STATIC_LIBRARY?

es0m: @snnn   unfortunately, you can't if you want to respect generator expressions (I tried!)
Would you like a different OPTION for this usecase?

wenbingl: > OCOS_STATIC_LIBRARY is different from protobuf_MSVC_STATIC_RUNTIME?

Sorry for typo, what I meant is protobuf_MSVC_STATIC_RUNTIME. We may create a flag like OCOS_MSVC_STATIC_RUNTIME for this usage if the flag is better.

I just tried in my local dev machine, it works with -DCMAKE_MSVC... in build.bat, except for the issue here: https://github.com/microsoft/onnxruntime-extensions/blob/1f0c76cefaa9359b9fc74ff4b9ed93214486205a/cmake/externals/sentencepieceproject.cmake#L22 which hard-code protobuf_MSVC_STATIC_RUNTIME=OFF, and it needs to be fixed.

snnn: I searched the code, the only occurrence of "OCOS_ENABLE_STATIC_LIB" is:
```cmake
if(NOT OCOS_ENABLE_STATIC_LIB AND CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
```
So, it has nothing to do with VC Runtime. OCOS_ENABLE_STATIC_LIB  has effects only when CMAKE_SYSTEM_NAME equals to "Emscripten", which does not use  vc runtime at all.  So I don't think the variable is used to control which VC Runtime to use. And we should not give the variable a new meaning(which will surprise the person who created this cmake option).

snnn: ONNX Runtime's [build.py](https://github.com/microsoft/onnxruntime/blob/main/tools/ci_build/build.py) has a build option: `--enable_msvc_static_runtime`. It works by setting `CMAKE_MSVC_RUNTIME_LIBRARY` cmake variable to  `MultiThreaded$<$<CONFIG:Debug>:Debug`.  If you build this repo with the main ONNX Runtime repo, you can use that build option. Otherwise you can do the same thing when you invoke cmake. Therefore I think there is no need to introduce a new cmake variable to control this. Every cmake project should respect the `CMAKE_MSVC_RUNTIME_LIBRARY` cmake variable's value.

snnn: And, as `CMAKE_MSVC_RUNTIME_LIBRARY` means "the default value of MSVC_RUNTIME_LIBRARY". In cmake files, we should not set `CMAKE_MSVC_RUNTIME_LIBRARY` as it will change the `default` value instead of the `actual` value: `MSVC_RUNTIME_LIBRARY`.

es0m: https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html#variable:CMAKE_MSVC_RUNTIME_LIBRARY
If you follow the docs, you're supposed to set CMAKE_MSVC_RUNTIME_LIBRARY not MSVC_RUNTIME_LIBRARY.

es0m: regarding your use of OCOS_ENABLE_STATIC_LIB -- I'd suggest writing some documentation on how to use the flag. It is unclear as you can see by it's usage. It's related to WASM only apparently. 
Also, instead of having some command line collection parameters assembled as per build.py, having a clear option to enable the required functionality might be better.

wenbingl:
@es0m, I verified -DCMAKE_MSVC_RUNTIME_LIBRARY="MultiThreaded$<$<CONFIG:Debug>:Debug>" parameter in ort-extensions CI pipeline and fixed the issue in protobuf, it works now:  https://github.com/microsoft/onnxruntime-extensions/pull/472

Is that what you needed?

es0m:
Hi @wenbingl let me check the output on my end. I believe more needs to be set (pthreads_enable_static, etc.)

PR #471, opened 2023-06-12
Update Xcode version and MacOS image.

edgchen1:


wenbingl (approved):

skottmckay (dismissed):

.pipelines/templates/xcode-build-and-test-step.yml
edgchen1, line 1: replaced this and other Xcode tasks with direct xcodebuild calls. it's a little easier to see what's going on this way, and it uses the system's current xcode version which we set with use-xcode-version.yml.

PR #472, opened 2023-06-12
support vc runtime static linkage

wenbingl:


.pipelines/ci.yml
```diff
@@ -271,6 +271,17 @@ jobs:
           ctest -C RelWithDebInfo --output-on-failure
         displayName: Run C++ native tests
 
+  - job: WindowsStaticVC
+    pool:
+      vmImage: 'windows-latest'
+
+    steps:
+      - script: |
+          call .\build.bat -DOCOS_ENABLE_CTEST=ON -DCMAKE_MSVC_RUNTIME_LIBRARY="MultiThreaded$<$<CONFIG:Debug>:Debug>"
```
wenbingl, line 280: the argument string seems not to be easily used, any better idea?

es0m: one could use a cmake option ... ;D

wenbingl: that could be a solution, but I noticed what protobuf did the same thing actually brought us some trouble to add some extra code to switch the flag back and forth, especially when ort-extensions was used as cmake sub-directory for another project.

DCMAKE_MSVC_RUNTIME_LIBRARY is documented feature from cmake and it's kind of a standard that we should follow.

sayanshaw24 (approved):

.pipelines/ci.yml
sayanshaw24: we could perhaps add a magic string to replace "MultiThreaded$<$<CONFIG:Debug>:Debug>" so it is simple for the user when passing the parameter for DCMAKE_MSVC_RUNTIME_LIBRARY to `build.bat` - and then we can parse it in the YAML file.

wenbingl:
Merge this fixing firstly, and any further improvement are welcomed.

Issue #473, opened 2023-06-16
ClipTokenizer ignores pad_token


horixon:
Some clip tokenizer use cases may require a pad other than '0'. A vocab can use id '0' to represent a character/word part/word.

Below is an example using huggyface where eos token is used to pad and is padded as expected with id 49407:
`from tokenizers import Tokenizer

tokenizer_max_length = 77
tokenizer = Tokenizer.from_pretrained("openai/clip-vit-base-patch32")

tokenizer.enable_truncation(tokenizer_max_length)
tokenizer.enable_padding(pad_id=49407, length=tokenizer_max_length)

tokenizer.encode(prompt).ids`
  
It looks like zero is always used and the pad token is not applied. https://github.com/microsoft/onnxruntime-extensions/blob/b399dd6b38f0e5b27625af1b7847f2b0c680a55f/operators/tokenizer/clip_tokenizer.cc#L186

PR #474, opened 2023-06-20
Change default pad token from 0 to 49407

sayanshaw24:
Currently HF uses "<|endoftext|>" (which has id 49407) as default pad token: https://github.com/huggingface/transformers/blob/v4.29.1/src/transformers/models/clip/tokenization_clip.py#L297

Fixes https://github.com/microsoft/onnxruntime-extensions/issues/473.


operators/tokenizer/clip_tokenizer.cc
```diff
@@ -177,14 +177,16 @@ void KernelClipBpeTokenizer::Compute(const ortc::Tensor<std::string>& input,
     }
   }
   int idx = 0;
+  // currently HF uses "<|endoftext|>" (which has id 49407) as default pad token
+  int pad_token = 49407;
```
wenbingl, line 181: can this id value be queried by vocab? This hard code is not guaranteed if the vocab is changed.

sayanshaw24: yep, done!

wenbingl (approved):

PR #477, opened 2023-06-29
Initial BertTokenizer offset mapping implementation

sayanshaw24:
Supports most basic sentences in En, Fr, De, Es, Cn, and Jp, given vocab handles tokens. Special or unknown tokens are not supported at the moment, as model training (the common use-case for offset mapping) does not usually include unknown tokens.

Validation:
- [x] Tested against HuggingFace expected offset mapping values.

### DIFF
--- a/operators/tokenizer/bert_tokenizer.cc
+++ b/operators/tokenizer/bert_tokenizer.cc
@@ -1,6 +1,7 @@
 #include "bert_tokenizer.hpp"
 
 #include <utility>
+#include <iostream>
 
 BertTokenizerVocab::BertTokenizerVocab(std::string_view vocab) : raw_vocab_(vocab) {
   auto tokens = SplitString(raw_vocab_, "\r\n", true);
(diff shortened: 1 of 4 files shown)


test/test_bert_tokenizer.py
```diff
@@ -3,6 +3,7 @@
 import numpy as np
 import transformers
 from onnxruntime_extensions import PyOrtFunction, BertTokenizer, util
+from transformers import BertTokenizerFast
```
wenbingl, line 6: Does BertTokenizer support offset_mapping?

operators/tokenizer/bert_tokenizer.cc
```diff
@@ -70,12 +73,43 @@ std::vector<ustring> WordpieceTokenizer::Tokenize(const ustring& text) {
...
 }
 
-std::vector<ustring> WordpieceTokenizer::Tokenize(const std::vector<ustring>& tokens) {
+std::vector<ustring> WordpieceTokenizer::Tokenize(const std::vector<ustring>& tokens, std::list<OffsetMappingType>& offset_map) {
   std::vector<ustring> result;
   for (const auto& token : tokens) {
     GreedySearch(token, result);
   }
 
+  size_t offset = 0;
```
wenbingl, line 82: Should the following code run only with offset_mapping ouptut being not optional?

wenbingl (approved):

test/test_bert_tokenizer.py
sayanshaw24: not the regular one - only the HuggingFace Fast tokenizers support offset_mapping.

operators/tokenizer/bert_tokenizer.cc
sayanshaw24: Yep, that's a great idea, thanks! Will create a new PR for this and do it for CLIP and Roberta as well.

PR #478, opened 2023-07-03
clip_image_processor

wejoncy:


onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
```diff
@@ -20,6 +20,48 @@ class ModelSource(enum.Enum):
     OTHER = 2
 
 
+def image_processor(extend_config: dict):
```
skottmckay, line 23: I think we're getting to a point where the complexity of this script in trying to handle so many different models is making it far less useful as a reference to someone who wants to create a custom pre/post processing pipeline or tweak an existing one. I'd like to start splitting it up into per-model examples (e.g. add_pre_post_processing_to_mobilenet) to simplify. 

Can we take that approach for CLIPImageProcessor and have it in a completely separate script?

Slightly different as it's a more generic processor and not model specific, so maybe called something like add_HuggingFace_CLIPImageProcessor_to_model.py?

wejoncy: "Transformers" and "Clip processor" was spinning  off from `add_pre_post_processing_to_model.py`. The usage is not changed.

skottmckay: Sorry - I don't think I was clear enough. I don't want the usage of add_pre_post_processing_to_model.py to be unchanged. I want to start breaking that up into simpler scripts so the pre/post processing pipeline is defined as directly as possible. It's much harder for a new user to see what is happening when they have to understand an extra layer of configuration. Main goal is for these to be reference implementations for external users.

With more model specific scripts I would hope it looks more like this example for a super-resolution model with different pre/post processing to the super resolution model we handle here currently:

https://github.com/microsoft/onnxruntime-inference-examples/blob/39d587857554451976fba379245a78184ac7d82d/mobile/examples/super_resolution/MAUI/real_esrgan_ort_e2e.py#L68-L92

That would mean that we leave what is in the existing script alone for now, and create a new script to handle just ClipImageProcessor. 

Does that make sense?

wejoncy: Make sence.

onnxruntime_extensions/tools/add_HuggingFace_CLIPImageProcessor_to_model.py
```diff
@@ -0,0 +1,122 @@
...
+        """,
+    )
+
+    parser.add_argument(
+        "--opset", type=int, required=False, default=16,
+        help="ONNX opset to use. Minimum allowed is 16. Opset 18 is required for Resize with anti-aliasing.",
+    )
+    
+    parser.add_argument(
+        "--extend_config", type=str, required=False, default=None,
```
skottmckay, line 106: Instead of making the user create a json file can we simplify and use equivalent arguments to https://github.com/huggingface/transformers/blob/35eac0df75c692c5b93c12f7eaf3279cab8bd7ce/src/transformers/models/clip/image_processing_clip.py#L56? 

I would expect the usage is typically a one-off for the user, so the overhead of creating a config file may not provide much benefit at the cost of them having to create a correctly formatted json file.

wejoncy: Sure.
Aligned the args with huggingFace ones.

onnxruntime_extensions/tools/add_HuggingFace_CLIPImageProcessor_to_model.py
```diff
@@ -0,0 +1,157 @@
...
+        steps.append(Resize(to_size))
+
+    if args.do_center_crop:
+        to_size = args.crop_size
+        to_size = (to_size, to_size)
+        steps.append(CenterCrop(*to_size))
+
+    if args.do_rescale:
+        steps.append(ImageBytesToFloat(args.rescale_factor))
+    steps.append(ChannelsLastToChannelsFirst())
```
skottmckay, line 38:
```suggestion
        steps.append(ImageBytesToFloat(args.rescale_factor))

    steps.append(ChannelsLastToChannelsFirst())
```

onnxruntime_extensions/tools/add_HuggingFace_CLIPImageProcessor_to_model.py
```diff
@@ -0,0 +1,157 @@
...
+    Used for models like stable-diffusion. should be compatible with 
+    https://github.com/huggingface/transformers/blob/main/src/transformers/models/clip/image_processing_clip.py
+
+    It applies to 'CLIP' image processor, and aligns with the HuggingFace class name.
+
+    A typical usage example is used in Stable diffusion.
+
+    :param model_file: The input model file path.
+    :param output_file: The output file path, where the finalized model saved to.
+    :param onnx_opset: The opset version of onnx model, default(16).
```
skottmckay, line 61: We should default to 18 as that has anti-aliasing

onnxruntime_extensions/tools/add_HuggingFace_CLIPImageProcessor_to_model.py
```diff
@@ -0,0 +1,157 @@
...
+    :param extend_config: The extended config for the model.
+    """
+    args = Dict2Class(kawrgs)
```
skottmckay, line 64: What's the benefit of this vs just passing the args from argparse directly? The usage seems the same either way.

onnxruntime_extensions/tools/pre_post_processing/steps/vision.py
```diff
@@ -522,12 +522,13 @@ class ImageBytesToFloat(Step):
     Convert uint8 or float values in range 0..255 to floating point values in range 0..1
     """
 
-    def __init__(self, name: Optional[str] = None):
+    def __init__(self, rescale_factor: float = 1/255, name: Optional[str] = None):
```
skottmckay, line 525: If we insert a new arg in the middle won't that break existing usage where the name is being specified?

test/test_tools_add_pre_post_processing_to_model.py
```diff
@@ -92,6 +93,66 @@ def new_output():
...
         self.assertTrue(abs(orig_results[orig_idx] - new_results[new_idx]) < 0.01)
 
+    def test_pytorch_mobilenet_using_clip_feature(self):
+        input_model = os.path.join(test_data_dir, "pytorch_mobilenet_v2.onnx")
+        output_model = os.path.join(test_data_dir, "pytorch_mobilenet_v2.updated.onnx")
+        input_image_path = os.path.join(test_data_dir, "wolves.jpg")
+
+        add_clip_feature.clip_image_processor(Path(input_model), Path(output_model), opset=16, do_resize=True, 
+                                              do_center_crop=True, do_normalize=True, do_rescale=True,
+                                              do_convert_rgb = True, size=256, crop_size=224,
```
skottmckay, line 103:
```suggestion
                                              do_convert_rgb=True, size=256, crop_size=224, 
```

onnxruntime_extensions/tools/add_HuggingFace_CLIPImageProcessor_to_model.py
wejoncy: It supports the usage like `clip_image_processor(param=value,.....)`. Otherwise, we have to pass a data object.

onnxruntime_extensions/tools/pre_post_processing/steps/vision.py
wejoncy: We don't have a reference in the repo, so we wouldn't break it.
we may break the case of the other Users' code.

onnxruntime_extensions/tools/add_HuggingFace_CLIPImageProcessor_to_model.py
```diff
@@ -0,0 +1,171 @@
...
+
+        Example usage:
+            object detection:
+            - python -m onnxruntime_extensions.tools.add_HuggingFace_CLIPImageProcessor_to_model model.onnx  
+        """,
+    )
+
+    parser.add_argument(
+        "--opset", type=int, required=False, default=18,
+        help="ONNX opset to use. Minimum allowed is 18. Opset 18 is required for Resize with anti-aliasing.",
```
skottmckay, line 110: The minimum allowed is still 16 isn't it as the pre/post processing Step implementations used that as the base opset? We want people to use 18+ for the anti-aliasing but updating the model should still work with opset 16.

skottmckay (dismissed):

onnxruntime_extensions/tools/add_HuggingFace_CLIPImageProcessor_to_model.py
wejoncy: Yes. Minimum allowed is 16 but we can set the default opset as 18.

skottmckay (approved):

PR #483, opened 2023-07-06
Update OrtExtensionsUsage to also use the ORT Objective-C API.

edgchen1:
Note: This PR depends on the new Objective-C API added in https://github.com/microsoft/onnxruntime/pull/16603.


test/ios/OrtExtensionsUsage/Podfile
```diff
@@ -1,11 +1,13 @@
-platform :ios, '11.0'
+platform :ios, '12.0'
 
 target 'OrtExtensionsUsage' do
   # Comment the next line if you don't want to use dynamic frameworks
   use_frameworks!
 
   # Pods for OrtExtensionsUsage
-  pod 'onnxruntime-c'
+  #pod 'onnxruntime-objc'  # can use this after 1.16 is released
```
edgchen1, line 8: will need to use a pre-release version prior to 1.16, after the corresponding change in ORT to add the custom op registration API gets in.

edgchen1:
some formatting changes - suggest ignoring whitespace changes when viewing diff

test/ios/OrtExtensionsUsage/OrtExtensionsUsage/OrtSwiftClient.swift
```diff
@@ -0,0 +1,65 @@
...
+    throw OrtSwiftClientError.error("Unexpected output shape")
+  }
+
+  let expected_output_element_type = ORTTensorElementDataType.uInt8
+  guard output_type_and_shape.elementType == expected_output_element_type else {
+    throw OrtSwiftClientError.error("Unexpected output element type")
+  }
+
+  let output_data: Data = try output_tensor.tensorData() as Data
+  let expected_pixel_bgr_data: [UInt8] = [128, 64, 32]
```
skottmckay, line 59: nit: A comment connecting the 128/64/32 values to the ordering (BGR and channels last) and the model (which I assume has 32 for all R values, 64 for all G values and 128 for all B values based on the model name) would save the reader from having to figure that slightly indirect relationship out.

skottmckay (approved):

test/ios/OrtExtensionsUsage/Podfile
```diff
@@ -5,7 +5,7 @@ target 'OrtExtensionsUsage' do
   use_frameworks!
 
   # Pods for OrtExtensionsUsage
-  pod 'onnxruntime-c'
+  pod 'onnxruntime-objc'
```
wenbingl, line 8: Is there any version restriction here, something like onnxruntime-objc >= 1.16?

edgchen1: yes, 1.16 is the minimum required version now. it should get the latest like this, but I added `>= 1.16.0`.

skottmckay (approved):
:shipit:

PR #491, opened 2023-07-13
Try to upgrade gradle wrapper to unblock android CI

wenbingl:


cmake/ext_java.cmake
wenbingl, line 1: file re-formatted, review with ignoring white-space.

edgchen1 (approved):

PR #496, opened 2023-07-20
Update whisper model test cases and e2e example

wenbingl:


test/test_audio_signal.py
```diff
@@ -57,7 +59,7 @@ def setUpClass(cls):
         audio = wavefile.readframes(samples)
         audio_as_np_int16 = np.frombuffer(audio, dtype=np.int16)
         audio_as_np_float32 = audio_as_np_int16.astype(np.float32)
-        max_int16 = 2**15
+        max_int16 = 2 ** 15
```
sayanshaw24, line 62: would make a minor difference if any, but have you tried using math.pow instead? might be faster in this case: https://stackoverflow.com/questions/20969773/exponentials-in-python-xy-vs-math-powx-y

sayanshaw24 (approved):
Looks great, this is awesome!

test/test_audio_signal.py
wenbingl: the file was re-formatted by PyCharm. 
BTW, the link you shared actually suggest ** is faster, do I missed something?
_"Using the power operator ** will be faster as it won’t have the overhead of a function call. "_

PR #498, opened 2023-07-25
Skottmckay/refactor azure invoker setup

RandySheriffH:


CMakeLists.txt
```diff
@@ -58,7 +58,7 @@ option(OCOS_ENABLE_OPENCV_CODECS "Enable cv2 and vision operators that require o
 option(OCOS_ENABLE_CV2 "Enable the operators in `operators/cv2`" ON)
 option(OCOS_ENABLE_VISION "Enable the operators in `operators/vision`" ON)
 option(OCOS_ENABLE_AUDIO "Enable the operators for audio processing" ON)
-option(OCOS_ENABLE_AZURE "Enable the operators for azure execution provider" OFF)
+option(OCOS_ENABLE_AZURE "Enable the operators for azure execution provider" ON)
```
RandySheriffH, line 61: For now it is OFF due to a few issues with cmake installed in CI.
The plan is to only build it in linux&windows&android packaging by a pipeline parameter.

CMakeLists.txt
```diff
@@ -379,14 +379,13 @@ if(OCOS_ENABLE_BLINGFIRE)
...
-
   if ($ENV{TEST_AZURE_INVOKERS_AS_CPU_OP} MATCHES "ON")
-    message(STATUS "Azure inovkers will be testable as cpu ops")
+    message(STATUS "Azure invokers will be testable as cpu ops")
     add_compile_definitions(TEST_AZURE_INVOKERS_AS_CPU_OP)
   endif()
 
   # Azure endpoint invokers
-  include(triton)
+  include(triton)  # TODO: Allow curl and triton support to be added independently
```
RandySheriffH, line 388: This could rename to azure.cmake to be more general.

operators/azure/azure_invokers.hpp
```diff
@@ -1,43 +1,44 @@
...
 
-struct AzureTextInvoker : public AzureInvoker {
-  AzureTextInvoker(const OrtApi& api, const OrtKernelInfo& info);
-  void Compute(std::string_view auth, std::string_view input, ortc::Tensor<std::string>& output);
+/// <summary>
+/// Azure Audio to Text
+/// Input: auth_token {string}, ??? (Update when AOAI endpoint is defined)
+/// Output: text {string}
+/// </summary>
+class AzureAudioToText : public CurlInvoker {
```
pranavsharma, line 17: Should we suffix the names with Invoker? Like AzureAudioToTextInvoker or is that too verbose? Rationale is to know instantly that this class is an invoker. For e.g. we've AzureTritonInvoker.

operators/azure/curl_invoker.hpp
```diff
@@ -0,0 +1,98 @@
...
+/// Base class for requests using Curl
+/// </summary>
+class CurlInvoker : public CloudBaseKernel {
+ protected:
+  CurlInvoker(const OrtApi& api, const OrtKernelInfo& info);
+  virtual ~CurlInvoker() = default;
+
+ public:
+  // Compute method that is used to co-ordinate all Curl based Azure requests
+  // TODO: Can this be `const` to enforce it's stateless?
```
pranavsharma, line 84: Yes, all kernels should be stateless. Hence this must be a const method.

operators/azure/cloud_ops_registration.cc
```diff
@@ -0,0 +1,27 @@
...
+// Licensed under the MIT License.
+
+#include "ocos.h"
+#include "azure_invokers.hpp"
+#include "azure_triton_invoker.hpp"
+#include "openai_invokers.hpp"
+
+using namespace ort_extensions;
+
+const std::vector<const OrtCustomOp*>& AzureInvokerLoader() {
```
pranavsharma, line 11: nit: I suppose this can be called CloudInvokerLoader since not all custom ops are Azure.

skottmckay: wasn't clear how best to name this given they're all using 'AzureExecutionProvider' regardless of the endpoint they're calling. because of that I left Azure in the name, but happy to change it

operators/azure/azure_invokers.hpp
skottmckay: I'm fine with adding the suffix to add clarity.

operators/azure/cloud_ops_registration.cc
pranavsharma: Ok, I see. Let's leave it as-is then. It would've been nice to name the EP as CloudExectuionProvider but that it was voted down for marketing reasons.

operators/azure/curl_invoker.hpp
skottmckay: https://github.com/microsoft/onnxruntime-extensions/pull/500

Will update the kernels here once that is checked in as the OrtLiteCustomStruct needs to be changed first.

skottmckay:
We need to give some thought to how inputs are defined for the kernels. I don't believe we can use the input value name directly for things like the form value as that only works if there is one azure operator in the model or there's no overlap between the input names of the azure operators.

e.g. if the input value is called 'prompt' we can add it as the optional prompt input, but if there are 2 nodes there is only one 'prompt' value allowed in the model.

If we want to keep things extremely flexible we could define a syntax to extract the request property name from the tensor name. e.g. 'prompt/0' 'prompt/1'. 

That way we can use the text up to the first '/' as the request property name and ignore the rest. If there's no '/' we use the name as is. 

Or we can define specific order of inputs. That would bring in the complexity of having to know the order and for us to handle missing optional inputs. 

Personally I'd lean towards the first option as it's a relatively simple scheme, users can't get the order wrong, easier for us to implement (no ordering constraints to validate/enforce), and the value name is easily associated with how it is being used.

operators/azure/azure_invokers.hpp
skottmckay: I've added 'Invoker' to the class names. I didn't change the operator names at this point as the 'Invoker' part doesn't seem meaningful enough at that level to warrant the longer name.

RandySheriffH:
> We need to give some thought to how inputs are defined for the kernels. I don't believe we can use the input value name directly for things like the form value as that only works if there is one azure operator in the model or there's no overlap between the input names of the azure operators.
> 
> e.g. if the input value is called 'prompt' we can add it as the optional prompt input, but if there are 2 nodes there is only one 'prompt' value allowed in the model.
> 
> If we want to keep things extremely flexible we could define a syntax to extract the request property name from the tensor name. e.g. 'prompt/0' 'prompt/1'.
> 
> That way we can use the text up to the first '/' as the request property name and ignore the rest. If there's no '/' we use the name as is.
> 
> Or we can define specific order of inputs. That would bring in the complexity of having to know the order and for us to handle missing optional inputs.
> 
> Personally I'd lean towards the first option as it's a relatively simple scheme, users can't get the order wrong, easier for us to implement (no ordering constraints to validate/enforce), and the value name is easily associated with how it is being used.

The 2nd options looks fine - but should we prefer to use the last part of "/" for the purpose?

CMakeLists.txt
skottmckay: Will do at the end so the diffs are easier to see.

skottmckay:
> The 2nd options looks fine - but should we prefer to use the last part of "/" for the purpose?

The second option was specific ordering of inputs where arbitrary names could be used but it sounds like you are voting for the first option of having a defined prefix/suffix. Is that the case?

I'm fine with a prefix or suffix. Suffix is possibly better if we prefer using the node name as a prefix. I assume we're providing tooling to insert these nodes so we could use the node name (generate from operator + node index if it doesn't have one) and combine that with the request input name giving something like OpenAIAudioInvoker0/prompt, OpenAIAudioInvoker0/file etc.

operators/azure/cloud_ops_registration.cc
```diff
@@ -0,0 +1,36 @@
...
+                               CustomAzureStruct("AzureTextToText", AzureTextToTextInvoker),
+                               CustomAzureStruct("OpenAIAudioToText", OpenAIAudioToTextInvoker)
+#ifdef AZURE_INVOKERS_ENABLE_TRITON
+                                   ,
+                               CustomAzureStruct("AzureTritonInvoker", AzureTritonInvoker)
+#endif
+#ifdef TEST_AZURE_INVOKERS_AS_CPU_OP
+                                   ,
+                               CustomCpuStruct("AzureAudioToText", AzureAudioToTextInvoker),
+                               CustomCpuStruct("AzureTextToText", AzureTextToTextInvoker),
```
skottmckay, line 25: Can we just use CPUExecutionProvider for these custom ops? Not sure what AzureExecutionProvider is required for.

RandySheriffH: Yes, we can and we did - the ops will be registered both as CPU and AzureEP ops.

PR #500, opened 2023-07-27
Make kernel Compute method implementations const

skottmckay:
Nodes can be called concurrently and Compute needs to be stateless due to that. 
This matches onnxruntime OpKernel::Compute

Update the kernels to make Compute const.
Some of the tokenizers required moving `byte_list_` and `compute_offset_mapping` to be local variables in the Compute instead of class members. 
Fix a couple of other warnings and some long lines.

### DIFF
--- a/base/string_tensor.cc
+++ b/base/string_tensor.cc
@@ -4,7 +4,7 @@
 #include "string_utils.h"
 #include "ustring.h"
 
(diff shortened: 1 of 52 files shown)


wenbingl (approved):
Nice refinement!

PR #501, opened 2023-07-28
Configure header path for SPM build

YUNQIUGUO:
Add ifdef for swift package manager support.

It requires a prefix for headers searching path.

Similarly as what we do for ORT C headers

### DIFF
--- a/includes/onnxruntime_extensions.h
+++ b/includes/onnxruntime_extensions.h
@@ -2,7 +2,10 @@
 // Licensed under the MIT License.
 
 #pragma once
+#ifdef SPM_BUILD
 #include "onnxruntime/onnxruntime_c_api.h"
+#else
+#include "onnxruntime_c_api.h"
 
 #ifdef _WIN32
 #define ORTX_EXPORT __declspec(dllexport)
--- a/includes/onnxruntime_extensions.h
+++ b/includes/onnxruntime_extensions.h
@@ -8,8 +8,8 @@
 // The reason why we need a prefix is that when Xcode includes the package it copies it to an internally generated path with 
 // the package name as a prefix. 
 // And we don't have control over the include paths when that happens in the user's iOS app. 
-// The Only way we find to make the include path work automatically for now.
-#ifdef SPM_BUILD
+// The only way we found to make the include path work automatically for now.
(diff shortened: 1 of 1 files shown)


skottmckay:
As background info we couldn't find a better way to do this in the ORT code. When xcode includes the package it copies it to an internally generated path with the package name as a prefix. We don't have control over the include paths when that happens in the user's app  so this is the only way we could find to make the include work automatically.

includes/onnxruntime_extensions.h
```diff
@@ -2,7 +2,11 @@
 // Licensed under the MIT License.
 
 #pragma once
+#ifdef SPM_BUILD
```
skottmckay, line 5: Add comments explaining how/where this is used and why

wenbingl, line 5: Is this SPM_BUILD a system pre-defined name or defined by ourselves? Since SPM stands for the sentencepiece model in  its repo and ort-extensions, it would be good if we can find another name.

YUNQIUGUO: I see. 

It's defined in the `cxxSettings` in `Package.swift` by ourselves. info fyi: https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html#cxxsetting

https://github.com/microsoft/onnxruntime-swift-package-manager/blob/c23b993b97b92d39e2e422f1d6f05781d7baee72/Package.swift#L63

https://github.com/microsoft/onnxruntime/blob/21a71d52bd2074b770807b209939ec11e2c64fa7/Package.swift#L44

but I was using the same macro name defined in ORT target. it would be good to have some consistency. @skottmckay Any suggestion on the macro names if we want to change to avoid naming conflicts for sentencepieces in extensions?

edgchen1: would suggest something verbose and unlikely to overlap with user-defined names, like `ORT_SWIFT_PACKAGE_MANAGER_BUILD`

skottmckay: I like Ed's suggestion.

YUNQIUGUO: thanks for the suggestion for naming.

updated as suggested.

wenbingl (approved):

PR #503, opened 2023-08-01
Add Bert tokenizer in the supported model list and code refinement

wenbingl:
…efinement


test/test_whisper.py
```diff
@@ -0,0 +1,88 @@
...
+
+import numpy as np
+import onnxruntime as _ort
+from packaging import version
+from transformers import WhisperProcessor
+from onnxruntime_extensions import OrtPyFunction, util, gen_processing_models
+
+
+@unittest.skipIf(version.parse(_ort.__version__) < version.parse("1.14.0"), "skip for onnxruntime < 1.14.0")
+class TestAutoTokenizer(unittest.TestCase):
```
sayanshaw24, line 14: I see why the Whisper testing code was moved out of `test_autotokenizer.py` (because it is not testing the AutoTokenizer since it is not using AutoTokenizer.from_pretrained) but then this class name should probably be changed from TestAutoTokenizer to something like TestWhisper right?

wenbingl: Good catch!

sayanshaw24 (approved):

PR #505, opened 2023-08-03
Set iOS version for simulator

skottmckay:
New Azure DevOps macOS image added an iOS 17 image. That is not a released version and is incompatible with the iPhone 8 we were using in the simulator.

Hardcode to iOS 16.4 for now to fix. We'll try and figure out a better setup for testing the latest released OS, including automatically choosing a compatible iPhone version.


edgchen1 (approved):

edgchen1 (approved):

.pipelines/templates/run-with-ios-simulator-steps.yml
```diff
@@ -6,7 +6,15 @@ steps:
...
-    ORT_EXTENSIONS_BUILD_SIMULATOR_ID=$(xcrun simctl create iPhoneSimulatorForPipeline com.apple.CoreSimulator.SimDeviceType.iPhone-8)
+    # just so we have a list of options if we need to adjust
+    xcrun simctl list devices
+    xcrun simctl list runtimes
+
+    # hardcode to 16.4 for now.
+    # latest macOS image in Azure DevOps has iOS of 17.0 so that is the default.
+    #   a) iPhone 8 is not supported by that OS so we need update the phone
+    #   b) we don't want to be using a beta iOS version in our main testing
+    ORT_EXTENSIONS_BUILD_SIMULATOR_ID=$(xcrun simctl create iPhoneSimulatorForPipeline com.apple.CoreSimulator.SimDeviceType.iPhone-8 com.apple.CoreSimulator.SimRuntime.iOS-16-4)
```
wenbingl, line 17: Given there is very less iOS related code in extensions and this is the not the final packaging pipeline, should we consider to choose a main-stream iOS version for a more robust test pipeline. iOS-8 was released at 2014 which is 9 years ago.

skottmckay: We're using iOS 16 (latest) with an old device (iPhone 8). 

Ed's going to look at automating picking a valid device + iOS combination for the current image so this doesn't fail again.

PR #512, opened 2023-08-07
Add docs generation script

sayanshaw24:
(updates certain markdown files used in onnxruntime repo gh-pages branch for ort website)


tools/gen_docs.py
wenbingl, line 1: The general idea is to read the local git repo, and output the 'fixed' docs into onnxruntime doc directory, which can be directly used for creating doc PR.

sayanshaw24: Got it, in that case we will need to use GitHub API since we'll need to write to files in the repo and on a specific branch using the script. Need to look into that approach, will change this PR to a draft and mark as ready for review when complete.

PR #513, opened 2023-08-07
More portable Java compiling support in MacOS

wenbingl:
Some java version installed by Homebrew requires this change.


cmake/ext_java.cmake
```diff
@@ -2,6 +2,11 @@ include(FindJava)
 find_package(Java REQUIRED)
 include(UseJava)
 if (NOT ANDROID)
+    set(JAVA_AWT_LIBRARY NotNeeded)
+set(JAVA_JVM_LIBRARY NotNeeded)
```
sayanshaw24, line 6: nit since indentation does not matter outside of scripts in cmake file but alignment?

sayanshaw24 (approved):

sayanshaw24 (approved):

PR #514, opened 2023-08-07
Test against ort-nightly when packaging Azure ops

RandySheriffH:
Test against ort-nightly when packaging Azure ops.


tools/test_cibuildwheel.bat
```diff
@@ -1,7 +1,8 @@
 @echo off
 if "%OCOS_ENABLE_AZURE%"=="1" (
     pushd %1\test
-    python -m pip install onnxruntime
+    python -m pip install coloredlogs flatbuffers numpy packaging protobuf sympy
+    python -m pip install -i https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ORT-Nightly/pypi/simple/ ort-nightly==1.16.0.dev20230806001
```
wenbingl, line 5: I think the nightly version will only keep for some time and then was deleted. Is this what you want?

tools/test_cibuildwheel.bat
```diff
@@ -1,7 +1,8 @@
 @echo off
 if "%OCOS_ENABLE_AZURE%"=="1" (
     pushd %1\test
-    python -m pip install onnxruntime
+    python -m pip install coloredlogs flatbuffers numpy packaging protobuf sympy
```
wenbingl, line 4: Are these ort's dependencies?

tools/test_cibuildwheel.bat
RandySheriffH: Shortly, we will update the version to the RC.
For now, let's just keep the latest one at hand.

tools/test_cibuildwheel.bat
RandySheriffH: that's right.

wenbingl (approved):

PR #518, opened 2023-08-09
Make CIs pass with Azure ops enabled by default.

skottmckay:
Enable Azure ops by default so that we're always building them. 
Add filtering to disable azure ops on platforms/builds where they are not supported.
To simplify, install dependencies required by default instead of requiring an environment variable to be set.
Misc. other cleanups.

TODO: automate testing of the azure ops most likely with a dummy webserver endpoint.


RandySheriffH (approved):

cmake/externals/triton.cmake
```diff
@@ -44,36 +36,44 @@ if (WIN32)
...
+  add_dependencies(getcurl getopenssl)
+
+  set(triton_extra_cmake_args -DVCPKG_TARGET_TRIPLET=${vcpkg_triplet}
                               -DCMAKE_TOOLCHAIN_FILE=${VCPKG_SRC}/scripts/buildsystems/vcpkg.cmake)
-  set(triton_patch_command "")
-  set(triton_dependencies ${VCPKG_DEPENDENCIES})
 
+  # insane... but we added patch.exe from the git 32-bit windows distribution because the CI image has removed it
+  # from the git install that is available, and this seems like the least egregious/inconsistent way to make it work...
+  set(triton_patch_exe ${PROJECT_SOURCE_DIR}/cmake/externals/git.Win32.2.41.03.patch/patch.exe)
```
wenbingl, line 75: should this patch tools in binary format be moved into some sub-folder in /tools?

cmake/externals/triton.cmake
```diff
@@ -102,22 +102,26 @@ else()
...
+# uses CentOS 7.
+#  1) use the full path to the version script file so 'ld' doesn't fail to find it. Looks like ld is running from the
+#     parent directory but not sure why the behavior differs vs. other linux builds
+#       e.g. building locally on Ubuntu is fine without the patch
+#  2) only set the CURL lib path to 'lib64' on a 64-bit CentOS build as 'lib64' is invalid on a 32-bit OS. without
+#     this patch the build of the third-party libraries in the triton client fail as the CURL build is not found.
+#
+# Also fix a Windows issue where the security tools complain about warnings being disabled by increasing the
+# warning level for triton/src/c++/library. Was /W0. Patched to /W3.
+set(triton_patch_command ${triton_patch_exe} --verbose -p1 -i ${PROJECT_SOURCE_DIR}/cmake/externals/triton_cmake.patch)
```
wenbingl, line 123: why does this patch command line not work for this case?
https://github.com/microsoft/onnxruntime-extensions/blob/4842e9d6ae97e907830b2b32502ae275f1b62e68/cmake/externals/sentencepieceproject.cmake#L13

PR #519, opened 2023-08-09
Update domain in triton test models

skottmckay:
Use the new preferred com.microsoft.extensions domain.
Add script to generate models and update domain.
For consistency update the OpenAI op/models/tests to use 'model_name' for the optional model name input. Default model name value can be provided by node attribute, or model author can make it an optional (add initializer with same name) or required input.


test/data/azure/create_triton.py
```diff
@@ -0,0 +1,78 @@
...
+    auth_token = helper.make_tensor_value_info('auth_token', TensorProto.STRING, [-1])
+    X = helper.make_tensor_value_info('X', TensorProto.FLOAT, [-1])
+    Y = helper.make_tensor_value_info('Y', TensorProto.FLOAT, [-1])
+    Z = helper.make_tensor_value_info('Z', TensorProto.FLOAT, [-1])
+    invoker = helper.make_node('AzureTritonInvoker', ['auth_token', 'X', 'Y'], ['Z'],
+                               domain='com.microsoft.extensions', name='triton_invoker',
+                               model_uri='https://endpoint-5095584.westus2.inference.ml.azure.com',
+                               model_name='addf', model_version='1', verbose='1')
+    graph = helper.make_graph([invoker], 'graph', [auth_token, X, Y], [Z])
+    model = helper.make_model(graph,
```
wenbingl, line 14: will this make_model function cause the IR version issue?

wenbingl (approved):

test/data/azure/create_triton.py
skottmckay: Not sure what the issue is. The script is just to re-create the checked in models if/when needed. It's not run as part of any CI..

test/data/azure/create_triton.py
```diff
@@ -0,0 +1,78 @@
+from onnx import helper, TensorProto, save
+# from onnxruntime import *
+
+def createAddf():
+    auth_token = helper.make_tensor_value_info('auth_token', TensorProto.STRING, [-1])
+    X = helper.make_tensor_value_info('X', TensorProto.FLOAT, [-1])
+    Y = helper.make_tensor_value_info('Y', TensorProto.FLOAT, [-1])
+    Z = helper.make_tensor_value_info('Z', TensorProto.FLOAT, [-1])
+    invoker = helper.make_node('AzureTritonInvoker', ['auth_token', 'X', 'Y'], ['Z'],
+                               domain='com.microsoft.extensions', name='triton_invoker',
```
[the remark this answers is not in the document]
skottmckay: I assume we will be adding tools to create/edit models to use the azure custom ops, and a constant defining the domain would be good there.

PR #520, opened 2023-08-10
Fix HF Fast Tokenizer cvt issue for AutoTokenizer imp

sayanshaw24:


onnxruntime_extensions/_hf_cvt.py
```diff
@@ -22,11 +22,22 @@ def __init__(self, tokenizer):
         self.tokenizer = tokenizer
 
     def bpe_tokenizer(self, **kwargs):
-        hf_gpt2_tokenizer = self.tokenizer
+        vocab_file = "./test/data/gpt2.vocab"
```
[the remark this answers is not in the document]
sayanshaw24: that was the original idea but self.tokenizer in this case is usually a GPT2TokenizerFast object (or some other fast tokenizer) and does not have an encoder that I can get the vocab/merges from (previous imp was throwing an error - explained on Teams). I tried getting the path to the vocab or merges at least from init_kwargs but that also did not exist for the current test case, but I can add a check so it tries to get the vocab and merges from self.tokenizer.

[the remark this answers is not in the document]
wenbingl: This doesn't work because of the vocab is not always same. If TokenizerFast doesn't have enough info, we use the slow version to extract the vocab and merges from it.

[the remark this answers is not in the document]
sayanshaw24: Got it, I was thinking the same thing and realized this needed to be fixed. So one other problem is what string to pass into `from_pretrained`, for example in `exec(f"hf_bpe_tokenizer = {slow_tokenizer}.from_pretrained('gpt2')")`? I can extract the name of the slow tokenizer by just stripping "Fast" from the name like we do for cls_name) but the string for `from_pretrained` could be different. Will add an implementation now (like taking the string before "Tokenizer") but  it might not work for all cases, so will have to think of a better fix in the future.

test/test_autotokenizer.py
```diff
@@ -30,15 +30,15 @@ def test_llama_tokenizer(self):
 
     def test_falcon_tokenizer(self):
         # replace the official model name after the model is not gated anymore
-        tokenizer = AutoTokenizer.from_pretrained("hf-internal-testing/llama-tokenizer")
+        tokenizer = AutoTokenizer.from_pretrained("Rocketknight1/falcon-rw-1b")
```
wenbingl, line 33:
```suggestion
        tokenizer = AutoTokenizer.from_pretrained("Rocketknight1/falcon-rw-1b", use_fast=False)
```

Using this way to get a slow Tokenizer, https://github.com/huggingface/transformers/blob/5e5fa0d88c293e6d5be2517b4f45680ba3bb5df2/src/transformers/models/auto/tokenization_auto.py#L594C17-L594C17

onnxruntime_extensions/_hf_cvt.py
```diff
@@ -22,11 +22,20 @@ def __init__(self, tokenizer):
         self.tokenizer = tokenizer
 
     def bpe_tokenizer(self, **kwargs):
-        hf_gpt2_tokenizer = self.tokenizer
+        hf_bpe_tokenizer = self.tokenizer
+
+        if ("Fast" in str(self.tokenizer)):
```
wenbingl, line 27: May just report an error for users to ask them use the slow version.

test/test_autotokenizer.py
sayanshaw24: perfect, missed this, thanks!

onnxruntime_extensions/_hf_cvt.py
sayanshaw24: Sounds good!

wenbingl (approved):

PR #521, opened 2023-08-11
Add PreTrainedTokenizer support

sayanshaw24:
add PreTrainedTokenizer support to HF cvt

### DIFF
--- a/onnxruntime_extensions/_hf_cvt.py
+++ b/onnxruntime_extensions/_hf_cvt.py
@@ -146,6 +146,17 @@ _PROCESSOR_DICT = {
                                      "SentencepieceDecoder",    HFTokenizerConverter.spm_decoder,
                                      default_inputs={'add_bos': [True]})
 }
+
+_PROCESSOR_TYPE_DICT = {
+    'bert':   "BertTokenizer",
+    'gpt2':   "GPT2Tokenizer",
+    'clip':   "ClipTokenizer",
+    'roberta':"RobertaTokenizer",
+    't5':     "T5Tokenizer",
+    'llama':  "LlamaTokenizer",
+    'bpe':    "GPT2Tokenizer",
+    'falcon': "GPT2Tokenizer"
+}
 # @formatter:on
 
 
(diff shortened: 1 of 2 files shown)


wenbingl:
Any test cases?

sayanshaw24:
> Any test cases?

Yeah, looking for some right now, will update soon.

onnxruntime_extensions/_hf_cvt.py
```diff
@@ -24,16 +25,31 @@ def __init__(self, tokenizer):
...
+            attrs = {}
+            attrs['vocab'] = f'{tokenizer_json["model"]["vocab"]}'
+            sorted_merges = {}
+            i = 0
+            for merge in tokenizer_json["model"]["merges"]:
+                sorted_merges[i] = (merge[:merge.index(" ")], merge[merge.index(" ")+1:])
+                i += 1
+            attrs['merges'] = '\n'.join("{} {}".format(
+                *sorted_merges[n_]) for n_ in range(len(sorted_merges)))
+            attrs.update(**kwargs)
```
sayanshaw24, line 41: parsing of vocab and merges need to be fixed - `PreTrainedTokenizerFast` object has vocab but not merges, so they need to be parsed in from somewhere - I have this implementation right now, but it's broken.

wenbingl (commented):
I read through the source code of PreTrainedTokenizerFast at https://github.com/huggingface/transformers/blob/b42010bb1d3cbf262d27e0a328661885be46dfdb/src/transformers/tokenization_utils_fast.py#L410. And it looks like it's only a thin wrapper of Its rust tokenizer, which seems not be feasible to get enough information to convert it into ONNX format.

onnxruntime_extensions/_hf_cvt.py
wenbingl: it doesn't work. Are you always assuming all vocab and merges are same?

wenbingl:
> > I read through the source code of PreTrainedTokenizerFast at https://github.com/huggingface/transformers/blob/b42010bb1d3cbf262d27e0a328661885be46dfdb/src/transformers/tokenization_utils_fast.py#L410. And it looks like it's only a thin wrapper of Its rust tokenizer, which seems not be feasible to get enough information to convert it into ONNX format.
> 
> What about the slow PreTrainedTokenizer? It is definitely true that most tokenizers built on top of that will likely have a common base tokenizer (like GPT2Tokenizer for Falcon) but adding it would make our code more extensive - implementation is almost done, if we can just fix the parsing of vocab/merges it will work.

the slower versions is an abstract class, and the pretty straightforward way to add more its derived class in the table, and drop this change.

sayanshaw24:
> > > I read through the source code of PreTrainedTokenizerFast at https://github.com/huggingface/transformers/blob/b42010bb1d3cbf262d27e0a328661885be46dfdb/src/transformers/tokenization_utils_fast.py#L410. And it looks like it's only a thin wrapper of Its rust tokenizer, which seems not be feasible to get enough information to convert it into ONNX format.
> > 
> > 
> > What about the slow PreTrainedTokenizer? It is definitely true that most tokenizers built on top of that will likely have a common base tokenizer (like GPT2Tokenizer for Falcon) but adding it would make our code more extensive - implementation is almost done, if we can just fix the parsing of vocab/merges it will work.
> 
> the slower versions is an abstract class, and the pretty straightforward way is to add more its derived class in the table, and drop this change.

sounds good!

PR #522, opened 2023-08-13
Android fixes/improvements

skottmckay:
- Ensure we log an error message before throwing on Android
  - message in exception will be lost due to how the shared libraries are built (both onnxruntime and extensions use static libc++ so there are no shared exception types between them)
- support static or dynamic build of curl/openssl on android
  - TBD which we want to use. static is smaller but build fails on x86 currently so might need a mix
- add infra for anything deriving from BaseKernel to log messages using the ORT logger
  - ensures messages from custom kernels end up in the same place as messages from ORT

### DIFF
--- a/includes/exceptions.h
+++ b/includes/exceptions.h
@@ -23,7 +23,6 @@
 
 namespace OrtW {
 
-#if !defined(OCOS_SHARED_LIBRARY)
 // All C++ methods that can fail will throw an exception of this type
 struct Exception : std::exception {
   Exception(std::string message, OrtErrorCode code) : message_{std::move(message)}, code_{code} {}
@@ -35,7 +34,6 @@ struct Exception : std::exception {
   std::string message_;
   OrtErrorCode code_;
 };
-#endif 
 
 // helper that outputs an error message in a platform aware manner
 // Usages:
--- a/includes/exceptions.h
+++ b/includes/exceptions.h
@@ -42,6 +42,9 @@ struct Exception : std::exception {
 inline void LogError(const ORTCHAR_T* file, int line, const char* msg) {
 #if defined(__ANDROID__)
   __android_log_print(ANDROID_LOG_ERROR, "onnxruntime-extensions", "Error in %s line %d: %s", file, line, msg);
+#elif defined(_WIN32)
+  // need to use wcerr as ORTCHAR_T is wchar_t on Windows
+  std::wcerr << "Error in " << file << " line " << line << ": " << msg << std::endl;
 #else
   std::cerr << "Error in " << file << " line " << line << ": " << msg << std::endl;
 #endif
@@ -75,7 +78,7 @@ inline void LogError(const ORTCHAR_T* file, int line, const char* msg) {
     std::string msg(msg_in);                                               \
     OrtW::LogError(ORT_FILE, __LINE__, msg.c_str());                       \
     throw std::runtime_error((std::to_string(code) + ": " + msg).c_str()); \
-  } while(false)
+  } while (false)
 #else
 #define ORTX_CXX_API_THROW(msg, code) \
   throw std::runtime_error((std::to_string(code) + ": " + msg).c_str())
--- a/includes/exceptions.h
+++ b/includes/exceptions.h
@@ -35,7 +35,7 @@ struct Exception : std::exception {
   std::string message_;
   OrtErrorCode code_;
 };
-#endif
+#endif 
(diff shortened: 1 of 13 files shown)


operators/azure/logger.hpp
```diff
@@ -0,0 +1,83 @@
...
+  const OrtApi& api_;
+  int api_version_{0};  // api version in use
+  // OrtLogger if available. Unused otherwise
+  const void* ort_logger_{nullptr};
+};
+
+// specialization for when ORT logger is available (compiled against and running with ORT version > 1.15
+// return 'bool'. the actual type doesn't matter - just has to be different to the default of 'void'
+template <>
+Logger<has_Logger_LogMessage<OrtApi>::type>::Logger(const OrtApi& api,
```
edgchen1, line 51: would the specialization need to be for `std::true_type`? i.e., to ensure that `has_Logger_LogMessage<OrtApi>::value` is actually true.

alternatively, `if constexpr (has_Logger_LogMessage<OrtApi>::value) { ... }` might be a simpler approach that can avoid the template specializations.

skottmckay: OrtLogger doesn't exist until 1.15 so I need this to potentially be thrown away due to SFINAE, so constexpr isn't an option (AFAIK).

skottmckay:
> Previously, onnx python windows packages were built with static vc runtime. Then we found it could lead to mysterious hard to debug crashes: [onnx/onnx#3493](https://github.com/onnx/onnx/issues/3493) . It all because they didn't catch all exceptions at binary boundaries.

For now unfortunately there's no other option. The custom op interface for Compute returns void - so there's no way to indicate failure if you do not throw. There are some v2 custom op interface functions recently added to allow Compute to return a status but we'd need to convert all the custom ops.

operators/azure/logger.hpp
```diff
@@ -0,0 +1,95 @@
...
+// Disable GCC 'ignoring attributes on template argument' warning due to Logger_LogMessage using
+// `__attribute__((warn_unused_result))`.
+// The template here is about whether Logger_LogMessage exists or not, so the attribute is irrelevant.
+#ifdef __GNUC__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wignored-attributes"
+#endif
+
+template <class, class = void>
+struct has_Logger_LogMessage : std::false_type {};
```
edgchen1, line 21: maybe move `has_Logger_LogMessage` and `LoggerImpl` into an internal `detail` namespace?

operators/azure/azure_triton_invoker.cc
```diff
@@ -98,7 +98,7 @@ int8_t* CreateNonStrTensor(const std::string& data_type,
 
 #define CHECK_TRITON_ERR(ret, msg)                                                    \
   if (!ret.IsOk()) {                                                                  \
-    return ORTX_CXX_API_THROW("Triton err: " + ret.Message(), ORT_RUNTIME_EXCEPTION); \
+    ORTX_CXX_API_THROW("Triton err: " + ret.Message(), ORT_RUNTIME_EXCEPTION); \
```
edgchen1, line 101: nit: spacing of trailing '\'. also noticed that the `msg` parameter is not used

includes/exceptions.h
```diff
@@ -29,21 +35,23 @@ struct Exception : std::exception {
...
+// Usages:
+//  - logging exception message when they may not propagate up
+//  - logging failure when using the ORT logger
+inline void LogError(const ORTCHAR_T* file, int line, const char* msg) {
 #if defined(__ANDROID__)
-  __android_log_print(ANDROID_LOG_ERROR, "onnxruntime-extensions", "Exception in %s line %d: %s", file, line, msg);
+  __android_log_print(ANDROID_LOG_ERROR, "onnxruntime-extensions", "Error in %s line %d: %s", file, line, msg);
 #else
-  std::cerr << "Exception in " << file << " line " << line << ": " << msg << std::endl;
+  std::cerr << "Error in " << file << " line " << line << ": " << msg << std::endl;
```
edgchen1, line 46: may need to convert from `const wchar_t*` to `const char*` first if ORTCHAR_T is wchar_t

edgchen1: huh, didn't know that `std::wostream` can work with `char*` too

edgchen1 (approved):

edgchen1 (approved):

PR #525, opened 2023-08-15
Only enable azureop build in CI builds

wenbingl:


wenbingl:
checked some log for the message and it works.

RandySheriffH (approved):

PR #531, opened 2023-08-17
Add more HF tokenizer supports in gen_processing_models

wenbingl:


onnxruntime_extensions/_hf_cvt.py
```diff
@@ -126,25 +121,38 @@ def spm_decoder(self, **kwargs):
...
+                                         'BpeDecoder',      HFTokenizerConverter.bpe_decoder, None),
+    "CodeGenTokenizer":     TokenOpParam('GPT2Tokenizer',   HFTokenizerConverter.bpe_tokenizer,
+                                         'BpeDecoder',      HFTokenizerConverter.bpe_decoder, None),
+    "CLIPTokenizer":        TokenOpParam('CLIPTokenizer',   HFTokenizerConverter.clip_tokenizer,
+                                         'BpeDecoder',      HFTokenizerConverter.bpe_decoder, None),
+    "RobertaTokenizer":     TokenOpParam('RobertaTokenizer',        HFTokenizerConverter.roberta_tokenizer,
+                                         'BpeDecoder',              HFTokenizerConverter.bpe_decoder, None),
+    "BartTokenizer":        TokenOpParam('RobertaTokenizer',        HFTokenizerConverter.roberta_tokenizer,
+                                         'BpeDecoder',              HFTokenizerConverter.bpe_decoder, None),
+    "LayoutLMv3Tokenizer":  TokenOpParam('RobertaTokenizer',        HFTokenizerConverter.roberta_tokenizer,
```
sayanshaw24, line 142: where did you find these tokenizers/did you look through the HF list and find ones using Roberta? trying to determine if there is a better way to know which ones are built on top ones we already have support for (like Roberta) so we can add more like this.

sayanshaw24 (approved):

onnxruntime_extensions/_hf_cvt.py
wenbingl: check the huggingface source file which link was put in the comment, which can show you what the exact tokenizer type of one class is.

PR #532, opened 2023-08-18
Make Azure ops work on Android

skottmckay:
- Use in-memory certs for curl on Android
  - could not get curl+openssl to be able to use the system ones 
    - tried patching openssl to use md5 hash and a bunch of other things but nothing worked with the system certs
  
- Use static build for curl and openssl
  - smaller binary size

- Pass in logger to curl callback so errors can be logged properly
  - make logger macro slightly more generic by making `logger` an argument


operators/azure/curl_invoker.cc
```diff
@@ -3,41 +3,103 @@
...
+#include <openssl/err.h>
+#include <openssl/ssl.h>
+#endif
+
 namespace ort_extensions {
+namespace {
+// need to do in memory cert on Android pending finding a way to use the system certs.
+#if defined(USE_IN_MEMORY_CURL_CERTS)
+// in-memory certs
+CURLcode sslctx_function(CURL* /*curl*/, void* sslctx, void* /*parm*/) {
```
skottmckay, line 22: Copied from https://curl.se/libcurl/c/cacertinmem.html

wenbingl (dismissed):

operators/azure/curl_invoker.cc
```diff
@@ -3,41 +3,103 @@
...
+#include "curl_certs/cacert.pem.inc"
+
+  // TODO: Doing this on every requests seems excessive. See if we can cache anything.
+  BIO* cbio = BIO_new_mem_buf(curl_pem, sizeof(curl_pem));
+  X509_STORE* cts = SSL_CTX_get_cert_store(static_cast<SSL_CTX*>(sslctx));
+
+  CURLcode rv = CURLE_ABORTED_BY_CALLBACK;
+
+  if (!cts || !cbio) {
+    return rv;
```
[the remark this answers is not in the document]
skottmckay: Updated to use gsl::finally and to cache the store across requests.

operators/azure/curl_invoker.cc
```diff
@@ -3,41 +3,137 @@
...
+    success = true;
+
+    return cts;
+  }();
+
+  // slightly obtuse cleanup on exit. 
+  // a lambda can't capture the static `store`, so call GetCachedStore() to get the pointer.
+  // we can't explicitly set `store` to nullptr due to this setup, but it's static cleanup and nothing else should be
+  // attempting to use the `store` when cleanup happens as it's only used in the SSL_CTX of a request. 
+  static auto free_store = gsl::finally([] {
```
edgchen1, line 81: I wonder if it's less obtuse to have a static `std::unique_ptr<X509_STORE*>` with a custom deleter?

IMO, `gsl::final_action` seems easier to reason about within local scopes.

test/data/azure/get_certs_for_model.py
```diff
@@ -0,0 +1,98 @@
...
+                out.write(line)
+                in_cert = "-----END CERTIFICATE-----" not in line
+
+        assert num_certs > 0
+        assert not in_cert  # mismatched begin/end if not false
+        print(f"Processed {num_certs} certificates")
+
+        # rewind and return as UTF-8 string
+        out.seek(0)
+        certs = out.read()
```
edgchen1, line 52: do we need to write a temporary file? could we write into something like [`StringIO`](https://docs.python.org/3/library/io.html#io.StringIO) instead?

test/data/azure/get_certs_for_model.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python3
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# Curl + openssl has issues reading the system certificates on Android.
+# Pending a better solution we create an in-memory certificate store from certificates included in the model.
+#
+# The certificates must be added to the first Azure operator in the model in an attribute called 'x509_certificates'.
```
edgchen1, line 8: seems like picking the first one could be tricky. e.g., if there are multiple Azure ops with no dependencies on each other, is one guaranteed to be first?

but pending a better solution, it seems fine for now. if we keep this approach later, I think it may be better to specify the global cert configuration at a higher level (e.g., in an environment variable) rather than as an op attribute.

skottmckay: if we need to support multiple sets of certs, we would need to make them an explicit input to the node as well as having a per-node cert store. not sure if that is ever likely though. 

otherwise I don't think the order matters. one azure op needs the certs, and cert store init happens during kernel creation (which is during session state finalization and prior to any execution). 

but ideally we can find a solution where we can use the system certs so they're not adding to model size.

edgchen1: oh, I see that the op just uses a non-empty cert attribute to initialize the store.

test/data/azure/get_certs_for_model.py
```diff
@@ -0,0 +1,94 @@
...
+    return certs
+
+
+def get_certs_from_url(url: str):
+    """
+    Read the contents of a URL that returns a PEM file, and return the certificates as a UTF-8 string
+    for inclusion as a node attribute of an Azure custom operator.
+    e.g. https://curl.se/ca/cacert.pem
+    :param url: URL that returns the PEM file contents
+    :return: UTF-8 string containing the certificates
```
edgchen1, line 60: nit: I guess the return type is actually `bytes`?

PR #538, opened 2023-08-21
Update cgmanifest for azure ops

RandySheriffH:
Update cgmanifest for azure ops.


pranavsharma (approved):

wenbingl (approved):

cgmanifest.json
```diff
@@ -158,7 +158,67 @@
...
+        },
+        "comments": "triton client to talk to AML triton service"
+      }
+    },
+    {
+      "component": {
+        "type": "git",
+        "git": {
+          "commitHash": "f54b0e47a08782a6131cc3d60f94d038fa6e0a51",
+          "repositoryUrl": "https://github.com/Tencent/rapidjson"
```
skottmckay, line 187: When we get packages using vcpkg does it need to specify this commit hash so we're consistent?

cgmanifest.json
```diff
@@ -158,7 +158,67 @@
...
+    },
+    {
+      "component": {
+        "type": "git",
+        "git": {
+          "commitHash": "04f42ceca40f73e2978b50e93806c2a18c1281fc",
+          "repositoryUrl": "https://github.com/madler/zlib"
+        },
+        "comments": "a general purpose data compression library"
+      }
```
skottmckay, line 210: We already have an entry for zlib from opencv. It would be helpful for the comment to explain how/when this zlib version is used.

RandySheriffH: Done - commented like " ... it is a dependency introduced by triton"

cgmanifest.json
RandySheriffH: it is the commit hash from rapidjson 1.1.0, which is consistent with the one we referred in cmake.

skottmckay: How does vcpkg select that version and does anything prevent it randomly changing in the future?

RandySheriffH: we are with a fixed version of vcpkg, would package versions from there randomly changing?

wenbingl (dismissed):

wenbingl (approved):

cgmanifest.json
skottmckay: I would have expected that to be the vcpkg software rather than a fixed version for every single package vcpkg supports.

There are versioning tags for selecting specific versions of packages (at least in the manifest mode).

https://learn.microsoft.com/en-us/vcpkg/reference/vcpkg-json#dependency-version-gt
https://learn.microsoft.com/en-us/vcpkg/users/versioning#baselines

PR #541, opened 2023-08-23
Remove the deprecating std::codecvt_utf8 from code base.

wenbingl:
std::codecvt is deprecating in C++ standard, and the perf is extreme slow in the tokenizer use case.


sayanshaw24 (approved):
this is a great find, and fix, thanks for looking into this!

PR #544, opened 2023-08-24
Test against rc

RandySheriffH:
Test Azure ops against 1.16 RC.


wenbingl:
put it as CI variable to you can run the pipeline with a specific build without updating the code?

RandySheriffH:
> put it as CI variable to you can run the pipeline with a specific build without updating the code?

We are deprecate this whatsoever for 1.17 ...

PR #545, opened 2023-08-24
Update pyfunc.cc for the prefast warning.

wenbingl:
PyCustomOpDef::AddOppyop/pyfunc.cc | Don't use std::move on constant variables. (es.56).

### DIFF
--- a/pyop/pyfunc.cc
+++ b/pyop/pyfunc.cc
@@ -401,7 +401,7 @@ void PyCustomOpDef::AddOp(const PyCustomOpDef* cod) {
   auto val = std::make_pair(op_domain, std::vector<PyCustomOpFactory>());
   auto [it_domain_op, success] = PyOp_container().insert(val);
   assert(success || !it_domain_op->second.empty());
-  it_domain_op->second.emplace_back(PyCustomOpFactory(cod, op_domain, op));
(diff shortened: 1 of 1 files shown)


pyop/pyfunc.cc
```diff
@@ -399,7 +399,7 @@ void PyCustomOpDef::AddOp(const PyCustomOpDef* cod) {
 
   // No need to protect against concurrent access, GIL is doing that.
   auto val = std::make_pair(op_domain, std::vector<PyCustomOpFactory>());
-  const auto [it_domain_op, success] = PyOp_container().insert(val);
+  auto [it_domain_op, success] = PyOp_container().insert(val);
   assert(success || !it_domain_op->second.empty());
   it_domain_op->second.emplace_back(PyCustomOpFactory(cod, op_domain, op));
```
snnn, line 404: At line 404 you probably do not need to have the "PyCustomOpFactory" string.

```
  it_domain_op->second.emplace_back(cod, op_domain, op);
```

snnn (dismissed):

pyop/pyfunc.cc
wenbingl: good to know, thanks.

snnn (approved):

PR #547, opened 2023-08-26
Enable running C++ unit tests on Android

edgchen1:
Enable running C++ unit tests on Android emulator.

- Update build script to copy over test files and run tests on an Android device.
- Set up a CI job to run tests on the emulator.
- Fix a test failure encountered when running on an actual Android device. Update string_lower to call ToLower on each UTF code point instead of each UTF-8 byte.


skottmckay (dismissed):
:shipit:

.pipelines/ci.yml
```diff
@@ -485,6 +485,41 @@ jobs:
...
+      - script: brew install ninja
+        displayName: Install ninja
+
+      - template: templates/run-with-android-emulator-steps.yml
+        parameters:
+          steps:
+
+          - bash: |
+              python ./tools/build.py \
+                --config Debug \
```
wenbingl, line 516: Do you think this Debug build is stable on simulator, Or RelWithDebInfo is more robust?

wenbingl (dismissed):

.pipelines/ci.yml
edgchen1: I guess RelWithDebInfo would be closer to the released code, will change to that.

wenbingl (approved):

Issue #330, opened 2022-12-14
Why Onnxruntime extensions are not installing on Arm Based Processor


faridelya:
**Error when installing onnxruntime-extensions**

 - I also tried different versions but everytime got same error
 ```
error: could not find a version that satisfies the requirement onnxruntime-extensions (from versions: none) 
error: no matching distribution found for onnxruntime-extensions
```
**Server Spec:**
```
Architecture
64-bit (Arm)  # Arm Based processor used in aws ec2 instance
CPU:
Core 16  and 1 thread per core
```
**Enviroment:**
```
python version  3.9

pip install onnxruntime-extensions # but got above error

#===============================================================

#Again deleted enviroment and created new enviroment with

python version  3.7
   
tried to install  pip install onnxruntime-extensions  # but same error arise
```

Any help would be appreciated please don not ignore  
why this is not installing ? or ths is not compatable with ARM base processor?

wenbingl:
Are you able to build it from source. https://github.com/microsoft/onnxruntime-extensions/blob/main/docs/development.md#python-package? The code itself is OK with ARM platforms, and there is no ARM package available now?

BTW, what platform are you requesting from ARM support?

faridelya:
@wenbingl thanks for kind reply.
i did install from official github repo for ARM processor on Ubuntu .
**The hard step were to install cmake before installing Onnxruntime-entensions**  
then it work for me  ----  but 

For docker it is tricky to install cmake but i didn't make it, if you can help please [click here](https://stackoverflow.com/questions/74810869/how-install-cmake-in-dockerfile-for-onnxruntime-extensions)

Chizkiyahu:
> Are you able to build it from source. https://github.com/microsoft/onnxruntime-extensions/blob/main/docs/development.md#python-package? The code itself is OK with ARM platforms, and there is no ARM package available now?
> 
> BTW, what platform are you requesting from ARM support?

Linux arrch64 and docker run on mac m1 
like `manylinux2014_aarch64` in [onnx pypi](https://pypi.org/project/onnxruntime/#files)

I can build a Github action that builds the wheel 
but I don't know how to update  [azure pipline](https://github.com/microsoft/onnxruntime-extensions/blob/396044310e41df9cc6f890661cd1a18e2954eb50/.pipelines/wheels_linux.yml#L10C14-L10C14) code

PR #551, opened 2023-08-30
build wheel linux aarch64

Chizkiyahu:
GitHub action to build wheels for Linux aarch64
because building from the source can take a very long time 
https://github.com/microsoft/onnxruntime-extensions/issues/330


Chizkiyahu:

Chizkiyahu:
this makes sense to convert this action to the Azure pipeline like the other build
but I don't know how to do it

wenbingl:
> this makes sense to convert this action to the Azure pipeline like the other build but I don't know how to do it

Thanks for the contribution, I can merge this PR to add this workflow in GitHub Action for a temporary solution.
In the long run we will add the ARM support in the release packages which will be generated in ADO as same as others.

Chizkiyahu:
> > this makes sense to convert this action to the Azure pipeline like the other build but I don't know how to do it
> 
> Thanks for the contribution, I can merge this PR to add this workflow in GitHub Action as a temporary solution for who don't build the package by themself.
> 
> In the long run we will add the ARM support in the release packages which will be generated in ADO as same as others.

Thanks for the response
Because it takes a long time to run on 2 CPU core
I think to reduce the number of times the workflow is trigger 
What is the best in push to main or when a new tag is created or both?

another quantion how files will publish to pypi?

wenbingl:
> > > this makes sense to convert this action to the Azure pipeline like the other build but I don't know how to do it
> > 
> > 
> > Thanks for the contribution, I can merge this PR to add this workflow in GitHub Action as a temporary solution for who don't build the package by themself.
> > In the long run we will add the ARM support in the release packages which will be generated in ADO as same as others.
> 
> Thanks for the response Because it takes a long time to run on 2 CPU core I think to reduce the number of times the workflow is trigger What is the best in push to main or when a new tag is created or both?
> 
> another quantion how files will publish to pypi?

The pipeline script for building Linux pipeline is here:  https://github.com/microsoft/onnxruntime-extensions/blob/main/.pipelines/wheels_linux.yml, we will add ARM supports in the next release if the code base tests well on your workflow. 

And I also changed the trigger on the main branch merged, may adjust it if there are too many PRs in the future.

PR #555, opened 2023-08-31
Refactor BBPE based Tokenizers

wenbingl:
1. Optimize BPE search date structure for efficiency.
2. Dedup the code with minor difference among BBPE classes.


operators/tokenizer/bpe_kernels.cc
```diff
@@ -0,0 +1,359 @@
...
+          break;
+        }
+
+        res.push_back(p.first);
+
+        if (compute_offset_mapping) {
+          if (ModelName() == BpeModelConf::kModel_CLIP) {
+            offset_mapping.emplace_back(std::make_pair(offset, ort_extensions::narrow<size_t>(offset + p.second)));
+            offset += p.second;
+          } else {
```
sayanshaw24, line 228: should we add a check for ModelName() == BpeModelConf::kModel_Roberta here because the offset map alg might not apply to others (in case we add more BPE-based tokenizers in the future)?

or we could just add a comment to say that it is for Roberta - "If adding new BPE tokenizers, check for a different offset mapping algorithm".

sayanshaw24 (approved):

operators/tokenizer/bpe_kernels.cc
```diff
@@ -0,0 +1,359 @@
...
+
+  if (ModelName() == BpeModelConf::kModel_CLIP) {
+    // Merges consecutive '\s+' for CLIP
+    /*
+      text = re.sub(r"\s+", " ", text)
+      text = text.strip()
+    */
+    ustring& str = input;
+    str.erase(std::unique(str.begin(), str.end(),
+                          [](char32_t lhs, char32_t rhs) { return IsUnicodeSpace(lhs) && IsUnicodeSpace(rhs); }),
```
edgchen1, line 124: this is an interesting use of the equality predicate. however, std::unique requires an equivalence relation.

from here: https://en.cppreference.com/w/cpp/algorithm/unique
"Elements are compared using the given binary predicate p. The behavior is undefined if it is not an equivalence relation."

https://en.wikipedia.org/wiki/Equivalence_relation

operators/tokenizer/bpe_utils.hpp
```diff
@@ -0,0 +1,231 @@
...
+      if (it->second != p_id) {
+        ORTX_CXX_API_THROW("Duplicate special tokens.", ORT_INVALID_ARGUMENT);
+      }
+    } else {
+      token_map_[p_str] = p_id;
+      token_list_.push_back(SpecialTokenInfo(std::move(p_str), p_id));
+    }
+  }
+
+  std::list<std::pair<ustring, int>> SplitBySpecialTokens(ustring input) const {
```
edgchen1, line 28: looks like the `std::list` only gets appended to. consider `std::vector` instead as it has less dynamic allocation overhead.

edgchen1: general: usually `vector` is a better choice if you don't need to frequently insert or remove from locations other than the end.

operators/tokenizer/bpe_kernels.h
```diff
@@ -0,0 +1,85 @@
...
+};
+
+class BpeModel;
+
+struct KernelBpeTokenizer : BaseKernel {
+  KernelBpeTokenizer(const OrtApi& api, const OrtKernelInfo& info, const BpeModelConf& conf);
+
+  void Compute(const ortc::Tensor<std::string>& input,
+               ortc::Tensor<int64_t>& tokenize_output,
+               std::optional<ortc::Tensor<int64_t>*> attention_mask,
```
edgchen1, line 33: `optional<Tensor>` or `Tensor*` both imply "optional Tensor", but what does `optional<Tensor*>` mean?

wenbingl: std::optional is used to indict this is an optional output in the ONNX operator in LiteCustomOp design. but I don't know why std::optional<...> & is not used. @RandySheriffH

operators/tokenizer/bpe_kernels.cc
wenbingl: Nice catch! in the case, I have to write a loop for eliminate the spaces.

operators/tokenizer/bpe_kernels.cc
wenbingl: I assume this behavior applies all models which don't clean up white-space like CLIP, not only Roberta.

operators/tokenizer/bpe_kernels.cc
wenbingl: both are UnicodeSpace is an equivalence relation, isn't it?  It satisfies reflexivity, symmetry and transitivity

operators/tokenizer/bpe_utils.hpp
wenbingl: One concern about std::vector related to this case is when element is large how efficient std::vector can move them into the new buffer on growing vector buffer.

operators/tokenizer/bpe_kernels.cc
edgchen1: I think reflexivity is not satisfied. for example, if x is not a unicode space, then `IsUnicodeSpace(x) && IsUnicodeSpace(x) == false`

operators/tokenizer/bpe_kernels.h
edgchen1: put another way, can the `Tensor*` contained within the `optional<Tensor*>` be null? if so, why bother with the `optional`? if not, why is it a pointer?

yeah, maybe we can pass a `Tensor&` instead. `optional<std::reference_wrapper<Tensor>>` could be a possibility.

wenbingl: LiteCustomOp needs a type like std::optional to know whether this ONNX operator output is optional. The current 'protocol' is std::optional<Tensor*> so the function prototype here has to align with it. But I agree std::optional<Tensor*> can be improved in LiteCustomOp in the future.

sayanshaw24 (approved):

test/test_trie_tokenizer.py
```diff
@@ -144,11 +144,11 @@ def test_trie_tokenizer(self):
 
     def test_ort_trie_tokenizer(self):
         vocab_data = util.read_file(self.vocab_file, 'rb')
-        tokr = OrtPyFunction.from_customop("TrieTokenizer", vocab=vocab_data)
+        tokr = OrtPyFunction.from_customop("TrieTokenizer", vocab=vocab_data, cpu_only=True)
         tokens = tokr(["I love you"])
         self.assertEqual(list(tokens[0]), [74, 31337, 22799])
 
-        detok = OrtPyFunction.from_customop("TrieDetokenizer", vocab=vocab_data)
+        detok = OrtPyFunction.from_customop("TrieDetokenizer", vocab=vocab_data, cpu_only=True)
```
sayanshaw24, line 151: nit: TrieDecoder instead of TrieDetokenizer?

wenbingl: decoder is too common which is also used for LLM, I intentionally uses a different name.

PR #556, opened 2023-09-07
Add support for Fairseq models (like XLMRobertaTokenizer)

sayanshaw24:
### Add support for Fairseq models (like XLMRobertaTokenizer), built on SPM.

**Usage**: pass in value `true` for `add_bos`, `add_eos`, and new optional input boolean `fairseq` when creating SentencePieceTokenizer.

Validation:
- [x] Tested against HuggingFace XLMRobertaTokenizer (uses Fairseq, SPM)

### DIFF
--- a/onnxruntime_extensions/_cuops.py
+++ b/onnxruntime_extensions/_cuops.py
@@ -315,7 +315,8 @@ class SentencepieceTokenizer(CustomOp):
             cls.io_def('alpha', onnx_proto.TensorProto.FLOAT, [None]),
             cls.io_def('add_bos', onnx_proto.TensorProto.BOOL, [None]),
             cls.io_def('add_eos', onnx_proto.TensorProto.BOOL, [None]),
-            cls.io_def('reverse', onnx_proto.TensorProto.BOOL, [None])
+            cls.io_def('reverse', onnx_proto.TensorProto.BOOL, [None]),
+            cls.io_def('xlm_roberta', onnx_proto.TensorProto.BOOL, [None])
         ]
 
     # beyond Python 3.7, the order of the dict is guaranteed to be insertion order
@@ -326,7 +327,8 @@ class SentencepieceTokenizer(CustomOp):
             'alpha': [0],
             'add_bos': [False],
             'add_eos': [False],
-            'reverse': [False]
+            'reverse': [False],
+            'xlm_roberta': [False]
         }
 
     @classmethod
--- a/onnxruntime_extensions/_cuops.py
+++ b/onnxruntime_extensions/_cuops.py
@@ -316,7 +316,7 @@ class SentencepieceTokenizer(CustomOp):
             cls.io_def('add_bos', onnx_proto.TensorProto.BOOL, [None]),
             cls.io_def('add_eos', onnx_proto.TensorProto.BOOL, [None]),
             cls.io_def('reverse', onnx_proto.TensorProto.BOOL, [None]),
-            cls.io_def('xlm_roberta', onnx_proto.TensorProto.BOOL, [None])
+            cls.io_def('fairseq', onnx_proto.TensorProto.BOOL, [None])
         ]
 
     # beyond Python 3.7, the order of the dict is guaranteed to be insertion order
@@ -328,7 +328,7 @@ class SentencepieceTokenizer(CustomOp):
             'add_bos': [False],
             'add_eos': [False],
             'reverse': [False],
-            'xlm_roberta': [False]
+            'fairseq': [False]
         }
 
     @classmethod
--- a/operators/tokenizer/sentencepiece_tokenizer.cc
+++ b/operators/tokenizer/sentencepiece_tokenizer.cc
@@ -63,7 +63,7 @@ void KernelSentencepieceTokenizer::Compute(const ortc::Tensor<std::string>& inpu
         content.push_back(tokenizer_.eos_id());
       }
 
-      if (fairseq.has_value() && fairseq.value()) {
+      if (fairseq.has_value() && (*fairseq)) {
         // HF Fairseq Example (XLMRobertaTokenizer) : https://huggingface.co/transformers/v4.6.0/_modules/transformers/models/xlm_roberta/tokenization_xlm_roberta.html#XLMRobertaTokenizer
         //
         // Original fairseq vocab and spm vocab must be "aligned":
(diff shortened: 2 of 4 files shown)


test/test_sentencepiece_ops.py
```diff
@@ -437,6 +476,33 @@ def test_string_sentencepiece_tokenizer_bin(self):
...
+            nbest_size=np.array(
+                [0], dtype=np.int64),
+            alpha=np.array([0], dtype=np.float32),
+            add_bos=np.array([False], dtype=np.bool_),
+            add_eos=np.array([False], dtype=np.bool_),
+            reverse=np.array([False], dtype=np.bool_),
+            xlm_roberta=np.array([True], dtype=np.bool_))
+        del inputs['model']
+        cc_txout = cc_sess.run(None, inputs)
+        assert_almost_equal(ids[0], cc_txout[0])
```
wenbingl, line 505: why is it 'almost' equal?

operators/tokenizer/sentencepiece_tokenizer.cc
```diff
@@ -53,6 +54,16 @@ void KernelSentencepieceTokenizer::Compute(const ortc::Tensor<std::string>& inpu
...
         content.push_back(tokenizer_.bos_id());
       }
+    } else if (xlm_roberta.has_value() && xlm_roberta) {
+      // As per HF, the first "real" token "," has position 4 in the XLMRobertaTokenizer vocab
+      // and position 3 in the SPM vocab, so we add a padding value of 1 to all IDs.
+      //std::transform(content.begin(), content.end(), content.begin(), [](int& i) { return i + 1; });
+      std::for_each(inloop.begin(), inloop.end(), [](int& n) { n++; });
+
+      // Add HF BOS (0) and EOS (2) tokens for the XLMRobertaTokenizer
+      content.push_back(xlm_bos);
```
wenbingl, line 64: what's the difference of xml_bos and tokenizer_.bos_id()?

operators/tokenizer/sentencepiece_tokenizer.cc
```diff
@@ -53,6 +54,16 @@ void KernelSentencepieceTokenizer::Compute(const ortc::Tensor<std::string>& inpu
       if (add_bos) {
         content.push_back(tokenizer_.bos_id());
       }
+    } else if (xlm_roberta.has_value() && xlm_roberta) {
+      // As per HF, the first "real" token "," has position 4 in the XLMRobertaTokenizer vocab
```
wenbingl, line 58: It's good to paste the XLM-Roberta tokenizer code snippet here to help understanding.

wenbingl:
don't forget to update this file
https://github.com/microsoft/onnxruntime-extensions/blob/696e7502f16de207bad1f13d236af9455c6e26d6/onnxruntime_extensions/_cuops.py#L318C16-L318C16

test/test_sentencepiece_ops.py
```diff
@@ -437,6 +476,33 @@ def test_string_sentencepiece_tokenizer_bin(self):
         for i in range(0, 2):
             assert_almost_equal(exp[i], py_txout[i])
             assert_almost_equal(exp[i], cc_txout[i])
+            
+    def test_xlm_roberta_tokenizer(self):
```
edgchen1, line 480: in general, can we also add C++ unit tests to exercise new functionality? the reason I ask is that the Python tests don't run on certain platforms like Android and iOS.

sayanshaw24: interesting - we should likely discuss this and revamp the entire testing process in another PR then i think? because we don't have C++ unit testing for any of the new functionality we have added recently (not to mention everything else, which we know works on Android/iOS because they're in prod but don't have unit testing for otherwise)

this file for example would need to include all the tokenizers, i only see wordpiece for now: https://github.com/microsoft/onnxruntime-extensions/blob/696e7502f16de207bad1f13d236af9455c6e26d6/test/static_test/test_tokenizer.cc

edgchen1: if it is more involved to set up, another PR is fine. but it would be good to have more tests actually running on those platforms.

I'm currently working on enabling the C++ unit tests on iOS so hopefully we will have that soon.

wenbingl: it requires a lot efforts and we also need know the value of this work.
Do we encounter any case that C++ code tests well with Python while it doesn't work on Android?
I know there is a lot of  if-else for mobile in build script, but there is every few #ifdef ANDROID in C++ code.

edgchen1: > it requires a lot efforts and we also need know the value of this work.

The value is that we can be more confident about the code that we release on these platforms if there is more test coverage on the same platform.

> Do we encounter any case that C++ code tests well with Python while it doesn't work on Android?

Not aware of a Python test or equivalent failing on Android, but we don't run the Python tests on Android.
I will note that I encountered a C++ test failure that showed up on an Android device in the process of enabling those tests on Android.
The more testing we can do on Android and iOS the better.

FYI @skottmckay

operators/tokenizer/sentencepiece_tokenizer.cc
sayanshaw24: done!

operators/tokenizer/sentencepiece_tokenizer.cc
sayanshaw24: xlm is 0 and spm is 1 - updated as per discussion yesterday

wenbingl (dismissed):

onnxruntime_extensions/_cuops.py
```diff
@@ -315,7 +315,8 @@ def get_inputs(cls):
             cls.io_def('alpha', onnx_proto.TensorProto.FLOAT, [None]),
             cls.io_def('add_bos', onnx_proto.TensorProto.BOOL, [None]),
             cls.io_def('add_eos', onnx_proto.TensorProto.BOOL, [None]),
-            cls.io_def('reverse', onnx_proto.TensorProto.BOOL, [None])
+            cls.io_def('reverse', onnx_proto.TensorProto.BOOL, [None]),
+            cls.io_def('xlm_roberta', onnx_proto.TensorProto.BOOL, [None])
```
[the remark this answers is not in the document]
sayanshaw24: typo - fixed!

test/test_sentencepiece_ops.py
sayanshaw24: let's discuss in a separate thread since this is not specific to this PR - resolving this.

wenbingl (approved):

PR #558, opened 2023-09-12
Enable the status returnable APIs from ORT 1.16 C ABI

wenbingl:
1. Support the custom ORT C++ lib package.
2. Migrate to the ComputeV2 and CreateKernelV2 functions which supports return error status.
3. Update the math domain operators with new function signature.


test/shared_test/test_exceptions.cc
```diff
@@ -11,7 +11,9 @@
 // throw in ctor which will be called during model load
 struct ExceptionalKernel1 : BaseKernel {
   ExceptionalKernel1(const OrtApi& api, const OrtKernelInfo& info) : BaseKernel(api, info) {
+    OCOS_API_IMPL_BEGIN
```
wenbingl, line 14: With kernel move to new APIs, these tests needed to be updated as well

sayanshaw24 (approved):
looks good! we likely need to add some comments/docs (especially for `includes/onnxruntime_customop.hpp`) in the future.

PR #562, opened 2023-09-14
Add more hf models into converter APIs

wenbingl:


test/test_autotokenizer.py
```diff
@@ -69,6 +69,45 @@ def test_clip_tokenizer(self):
         actual_ids = ort_tok([text])[0]
         np.testing.assert_array_equal(ids, actual_ids)
 
+    def test_xmlroberta_tokenizer(self):
```
sayanshaw24, line 72: nit: typo - should be xlmroberta not xml

sayanshaw24 (approved):

test/test_autotokenizer.py
wenbingl: let's fix it the next PR.

PR #566, opened 2023-09-26
Add token indices output to sentencepiece

samwebster:


operators/tokenizer/sentencepiece_tokenizer.cc
```diff
@@ -32,35 +33,47 @@ void KernelSentencepieceTokenizer::Compute(const ortc::Tensor<std::string>& inpu
...
                                            ortc::Tensor<int32_t>& output,
                                            ortc::Tensor<int64_t>& output1,
+                                           ortc::Tensor<int32_t>& output2,
                                            std::optional<bool> fairseq) const {
   // Update with the new API
   auto& str_input = input.Data();
   // computation
 
-  std::vector<int64_t> indices;
+  std::vector<int64_t> instance_indices;
```
sayanshaw24, line 42: looks great! `instance_indices` should have the offset values as I discussed with Justin. However, we would like to keep `indices` as they were implemented previously as well, in case there are other partner requirements for it - if not, we can deprecate it in the future. could you add back the `indices` output (i.e., we have both `indices` and `instance_indices`)?

i can then approve and we can check this in to unblock shipping for you.

samwebster: Thanks. The old indices are still there just renamed to `instance_indices` to make it clear how they differ from `token_indices`

sayanshaw24: right, sorry missed that. yeah i think the name change should be fine, could you also update it here (and add the output value for `token_indices`): https://github.com/microsoft/onnxruntime-extensions/blob/main/onnxruntime_extensions/_cuops.py#L338

samwebster: Done

sayanshaw24: cheers, thanks! will approve after CI checks pass.

samwebster:
Could somebody please make the necessary fixes as I can't see the build logs? I've tried running your pipeline in our subscription but it doesn't work

edgchen1:
there is a segmentation fault in a Python test. the attached logs are from a Linux build but there are also failures on Windows and MacOS. maybe see if you can reproduce this locally.

[build_log.txt](https://github.com/microsoft/onnxruntime-extensions/files/12741948/build_log.txt)

joburkho:
Does this change modify the required outputs of the existing SentencepieceTokenizer operator?

I ask because Office has already shipped its Fluency model to production that uses the existing SentencepieceTokenizer operator, so if this breaks models that use the existing SentencepieceTokenizer operator, please create another operator for this work or ensure any new outputs are optional, not required.
________________________________

samwebster:
Additional output made optional and tests fixed

sayanshaw24 (dismissed):

operators/tokenizer/sentencepiece_tokenizer.cc
```diff
@@ -32,35 +33,49 @@ void KernelSentencepieceTokenizer::Compute(const ortc::Tensor<std::string>& inpu
                                            bool add_rev,
                                            ortc::Tensor<int32_t>& output,
                                            ortc::Tensor<int64_t>& output1,
+                                           std::optional<ortc::Tensor<int32_t>*> output2,
```
skottmckay, line 36: Operator inputs/outputs are based on their order so I believe this needs to be added at the end of current arguments otherwise it will break any existing usage.

e.g. any existing model where the `fairseq` output is being consumed will expect the 3rd output to be a scalar bool and this change would break that model

samwebster: Thanks. Done

sayanshaw24 (approved):

PR #567, opened 2023-10-03
Close gap in vocab for AutoTokenizer support for GPT4Tokenizer

sayanshaw24:
For WSSI work as requested by Ran.

### DIFF
--- a/onnxruntime_extensions/_hf_cvt.py
+++ b/onnxruntime_extensions/_hf_cvt.py
@@ -9,6 +9,8 @@ _hf_cvt.py: HuggingFace Tokenizer/Processor Converter
 
 import json
 import onnx
+import uuid
+import numpy as np
 from numpy import array as nparray
 from functools import partial
 from collections import namedtuple, OrderedDict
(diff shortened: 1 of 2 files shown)


RandySheriffH (approved):

onnxruntime_extensions/_hf_cvt.py
```diff
@@ -23,12 +25,31 @@ def __init__(self, tokenizer):
...
     def bpe_tokenizer(self, **kwargs):
         hf_gpt2_tokenizer = self.tokenizer
+        attrs = None
 
         if type(self.tokenizer).__name__.endswith('Fast'):
             raise ValueError('Please use the slow version of the tokenizer (ex: GPT2Tokenizer).')
-
-        attrs = {'vocab': json.dumps(
-            hf_gpt2_tokenizer.encoder, separators=(',', ':'))}
+        elif(self.tokenizer.name_or_path.endswith('gpt-4')):
```
natke, line 32: Can we add a comment here?

test/test_autotokenizer.py
```diff
@@ -68,6 +68,17 @@ def test_clip_tokenizer(self):
             pre_kwargs={"WITH_DEFAULT_INPUTS": True})[0])
         actual_ids = ort_tok([text])[0]
         np.testing.assert_array_equal(ids, actual_ids)
+        
+    def test_gpt2_tokenizer(self):
```
natke, line 72: Does the GPT-4 tokenizer get tested by this code?

onnxruntime_extensions/_hf_cvt.py
natke: to explain why we need the if clause

sayanshaw24: yep, i added one in L33!

test/test_autotokenizer.py
sayanshaw24: yep - to be more precise though, the gpt4 vocab gets tested - the actual tokenization uses the same alg as the gpt2 tok which is why loading it from HF uses the HF `GPT2Tokenizer` class as well.

PR #570, opened 2023-10-09
Ensure noexcep_operators and ocos_operators get built always

baijumeswani:
When ONNX Runtime is built with the flags `--use_extensions` within a minimal build (reduced ops), some of the operators are not built within the `onnxruntime-extensions`.

For example, building onnx runtime with the flag `--cmake_extra_defines=OCOS_ENABLE_TF_STRING=ON` only expects the ocos string operators to be built and included. These operators are currently built inside the `noexcep_operators` and as a result `ocos_operators` library does not get built (as it has no valid sources).

However, onnxruntime takes a dependency on `ocos_operators` library (and now also on `noexcep_operators` library after https://github.com/microsoft/onnxruntime/pull/17850). Since `ocos_operators` library does not get built, it results in ONNX Runtime build failure as well.

This PR introduces placeholder source files that always get included in the target sources (regardless of whether there are other src files or not). Doing this, guarantees that both the libraries get built.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -493,6 +493,11 @@ endif()
 # ### make all compile options.
 add_compile_options("$<$<C_COMPILER_ID:MSVC>:/utf-8>")
 add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/utf-8>")
+
+# Library will not be built if the target src for the lib does not contain any valid *.cc files.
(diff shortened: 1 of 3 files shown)


sayanshaw24 (approved):
Looks good, thanks!

skottmckay (approved):

PR #573, opened 2023-10-13
Add ability to create a 'dummy' nuget package using build with no ops

skottmckay:
Add ability to create a 'dummy' nuget package from a build with no ops. This is required on iOS due to the ORT C# package including the ability to register custom ops. On iOS that creates a compile time dependency for the RegisterCustomOps symbol to exist as static libraries are involved. If the user can't (Xamarin) or does not want to use extensions that results in usage of ORT being broken.

For simplicity, consistency and to minimize changes to the build infrastructure we build a nuget package for all platforms so everything is 1:1 with the real package except the build uses an empty config to select the ops to include. Technically we only need something with iOS builds.

We should only need to publish this once unless the RegisterCustomOps API changes.

See https://github.com/microsoft/onnxruntime/issues/16463


.pipelines/templates/build-dummy-package-for-nuget.yml
```diff
@@ -0,0 +1,33 @@
...
+parameters:
+- name: DoCompliance
+  displayName: Run Compliance Tasks?
+  type: boolean
+  default: true
+
+- name: DoEsrp
+  displayName: Run code sign tasks? Must be true if you are doing an OnnxRuntime extensions release.
+  type: boolean
+  default: true
```
sayanshaw24, line 12: do we have to run compliance or code sign for the dummy package?

.pipelines/templates/build-dummy-package-for-nuget.yml
```diff
@@ -0,0 +1,33 @@
...
+  default: none
+
+stages:
+- template: build-package-for-nuget.yml
+  parameters:
+    IsDummyPackage: true
+    DoCompliance: ${{ parameters.DoCompliance }}
+    DoEsrp: ${{ parameters.DoEsrp }}
+    IsReleaseBuild: ${{ parameters.IsReleaseBuild }}
+    NugetVersionSuffix: ${{ parameters.NugetVersionSuffix }}
```
sayanshaw24, line 31: perhaps we can add a flag to the version suffix for the dummy package for clarity

wenbingl:
> Add ability to create a 'dummy' nuget package from a build with no ops. This is required on iOS due to the ORT C# package including the ability to register custom ops. On iOS that creates a compile time dependency for the RegisterCustomOps symbol to exist as static libraries are involved. If the user can't (Xamarin) or does not want to use extensions that results in usage of ORT being broken.
> 
> For simplicity, consistency and to minimize changes to the build infrastructure we build a nuget package for all platforms so everything is 1:1 with the real package except the build uses an empty config to select the ops to include. Technically we only need something with iOS builds.
> 
> We should only need to publish this once unless the RegisterCustomOps API changes.
> 
> See [microsoft/onnxruntime#16463](https://github.com/microsoft/onnxruntime/issues/16463)

I think the mobile package is already minimized enough and cannot see too much value to create a dummy package for that. Moving forward, the extensions package might not be "optional" for users and if the user extremely looks for binary size and they should build from source.

tools/android/build_aar.py
```diff
@@ -229,7 +232,7 @@ def main():
         api_level=args.api_level,
         sdk_path=args.sdk_path,
         ndk_path=args.ndk_path,
-        other_args=unknown_args,
+        build_py_args=args.build_py_args
```
edgchen1, line 235: oops, looks like I messed this up when I added `build_py_args` to this script. thanks for fixing it.
do we still need to allow unknown args (i.e. use `parse_known_args`) now?

.pipelines/templates/build-package-for-nuget.yml
```diff
@@ -215,7 +214,8 @@ stages:
         versionSpec: 6.2.1
 
     - task: PowerShell@2
-      displayName: 'Updating NuSpec'
+      displayName: 'Updating NativeNuget.NuSpec'
+      condition: eq(${{ parameters.IsDummyPackage }}, false)
```
edgchen1, line 218: consider using [conditional step insertion](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#conditionally-run-a-step).

e.g.,
```yaml
steps:
# ...
- ${{ if eq(parameters.IsDummyPackage, false) }}:
  - task: # non dummy package step...
- ${{ else }}:
  - task: # dummy package step...
```

I think the condition is more obvious here instead of in the task's `condition` value. also, the disabled step won't be displayed in the UI.

.pipelines/nuget.yml
```diff
@@ -14,6 +14,12 @@ parameters:
   type: boolean
   default: false
 
+- name: BuildTemplate
```
edgchen1, line 17: would it be simpler to just have an `IsDummyPackage` parameter at this level? the build-package-for-nuget.yml template can be called with different parameters based on it.

a yaml template file input is very flexible but requires the caller to know what kind of template file to specify.

tools/android/build_aar.py
skottmckay: There's no usage I could see where they're needed now that the build.py args are handled separately. Will remove.

.pipelines/templates/build-dummy-package-for-nuget.yml
skottmckay: What would the flag be? The dummy package has a different name so it's differentiated from the 'real' package via that already.

.pipelines/templates/build-dummy-package-for-nuget.yml
skottmckay: I assume so as we need to publish it. Technically all compliance checks are done for the real package so don't need to be repeated, but it's easier to not have to explain why we skip them here if someone asks. It's a one-off build and publish so any cost from doing so is minimal.

.pipelines/nuget.yml
```diff
@@ -28,12 +33,24 @@ pr:
...
-    OrtNugetPackageId: 'Microsoft.ML.OnnxRuntime.Extensions'
+- ${{ if eq(parameters.IsDummyPackage, false) }}:
+  - template: templates/build-package-for-nuget.yml
+    parameters:
+      DoCompliance: ${{ parameters.DoCompliance }}
+      DoEsrp: ${{ parameters.DoEsrp }}
+      IsReleaseBuild: ${{ parameters.IsReleaseBuild }}
+      NugetVersionSuffix: ${{ parameters.NugetVersionSuffix }}
+
+- ${{ if eq(parameters.IsDummyPackage, true) }}:
```
edgchen1, line 46: nit: would `else` work?
```suggestion
- ${{ else }}:
```

.pipelines/templates/build-package-for-macosx.yml
```diff
@@ -20,11 +20,11 @@ stages:
   - template: mac-shared-lib-build.yml
     parameters:
       MacosArch: 'arm64'
-      AdditionalBuildFlags: "${{ parameters.AdditionalBuildFlags }} CMAKE_OSX_ARCHITECTURES=arm64"
+      AdditionalBuildFlags: "${{ parameters.AdditionalBuildFlags }} --cmake_extra_defines CMAKE_OSX_ARCHITECTURES=arm64"
```
edgchen1, line 23: good change. relying on `--cmake_extra_defines` preceding this is a bit confusing.
FYI I added [`--one_cmake_extra_define`](https://github.com/microsoft/onnxruntime-extensions/blob/68b9d1dc47663a9017c55d136c804417c8efec7d/tools/build.py#L131-L137) which takes one argument for situations like this.

edgchen1 (dismissed):

.pipelines/templates/build-package-for-nuget.yml
```diff
@@ -214,20 +213,41 @@ stages:
...
+            python $(Build.SourcesDirectory)\tools\ci_build\update_nuspec_for_native_nuget.py `
+              --package_version $OrtExtVersion `
+              --commit_id $(Build.SourceVersion) `
+              --is_release_build ${{ parameters.IsReleaseBuild }} `
+              --nuget_version_suffix ${{ parameters.NugetVersionSuffix }}
+
+            cat $(Build.SourcesDirectory)\nuget\NativeNuget.nuspec
+          workingDirectory: '$(Build.SourcesDirectory)'
+
+    - ${{ if eq(parameters.IsDummyPackage, true) }}:
```
edgchen1, line 232:
```suggestion
    - ${{ else }}:
```

.pipelines/templates/build-package-for-nuget.yml
```diff
@@ -241,12 +261,21 @@ stages:
...
-        packDestination: $(Build.ArtifactStagingDirectory)
+    - ${{ if eq(parameters.IsDummyPackage, false) }}:
+      - task: NuGetCommand@2
+        displayName: Packing Microsoft.ML.OnnxRuntime.Extensions NuGet-package
+        inputs:
+          command: 'pack'
+          packagesToPack: '$(Build.BinariesDirectory)\nuget\NativeNuget.nuspec'
+          packDestination: $(Build.ArtifactStagingDirectory)
+
+    - ${{ if eq(parameters.IsDummyPackage, true) }}:
```
edgchen1, line 272:
```suggestion
    - ${{ else }}:
```

edgchen1 (approved):

edgchen1 (approved):

PR #574, opened 2023-10-14
Fix newline and apostrophe handling for BPE

sayanshaw24:
fixes newline and apostrophe handling for issues discovered by long text testing.

**Update**: CLIPTokenizer tokenizes apostrophes separately (['you', ''', 're']) but CLIPTokenizerFast tokenizes together (['you', ''re']), **unless ftfy is installed**: https://github.com/huggingface/transformers/issues/22166.

### DIFF
--- a/operators/tokenizer/bpe_kernels.cc
+++ b/operators/tokenizer/bpe_kernels.cc
@@ -150,6 +150,7 @@ std::vector<int64_t> KernelBpeTokenizer::Tokenize(ustring& input,
     }
     // remove newlines as CLIP ignores them (treats them as whitespace which is then cleaned)
     str.erase(std::remove(str.begin(), str.end(), '\n'), str.end());
+    str.erase(std::remove(str.begin(), str.end(), '\r'), str.end());
     input = str;
   }
 
--- a/operators/tokenizer/bpe_kernels.cc
+++ b/operators/tokenizer/bpe_kernels.cc
@@ -198,8 +198,7 @@ std::vector<int64_t> KernelBpeTokenizer::Tokenize(ustring& input,
     }
 
     while (static_cast<int64_t>(res.size()) < max_length) {
-      // Skip splitting words with apostrophes for CLIP (e.g., you're, i'm, etc.)
-      auto [b, tok] = regcmp.GetNextToken(ModelName() == BpeModelConf::kModel_CLIP);
+      auto [b, tok] = regcmp.GetNextToken();
       
       if (!b) break;
 
--- a/operators/tokenizer/bpe_kernels.cc
+++ b/operators/tokenizer/bpe_kernels.cc
@@ -148,6 +148,8 @@ std::vector<int64_t> KernelBpeTokenizer::Tokenize(ustring& input,
     if (IsUnicodeSpace(str.back())) {
       str.pop_back();
     }
+    // remove newlines as CLIP ignores them
+    str.erase(std::remove(str.begin(), str.end(), '\n'), str.end());
     input = str;
   }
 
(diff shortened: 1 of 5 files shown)


operators/tokenizer/bpe_utils.hpp
```diff
@@ -100,28 +100,30 @@ class TokenWithRegularExp {
...
-      }
-
-      if (m_text.size() > 2) {
-        if (((m_text[1] == U'r') && (m_text[2] == U'e')) ||
-            ((m_text[1] == U'v') && (m_text[2] == U'e')) ||
-            ((m_text[1] == U'l') && (m_text[2] == U'l'))) {
-          std::u32string_view res = m_text.substr(0, 3);
-          m_text = m_text.substr(3);
+  std::u32string_view TryMatch(bool skipApostrophes) {
+    if (!skipApostrophes) {
```
wenbingl, line 104: can you copy the Python implementation here for better understanding?

operators/tokenizer/bpe_kernels.cc
```diff
@@ -148,6 +148,8 @@ std::vector<int64_t> KernelBpeTokenizer::Tokenize(ustring& input,
     if (IsUnicodeSpace(str.back())) {
       str.pop_back();
     }
+    // remove newlines as CLIP ignores them
+    str.erase(std::remove(str.begin(), str.end(), '\n'), str.end());
```
wenbingl, line 152: they are removed or are treated as a WS?

operators/tokenizer/bpe_utils.hpp
sayanshaw24: sounds good - issue is Python implementation has same regex as us but from manual testing it works differently: for example given `r"""<\|startoftext\|>|<\|endoftext\|>|'s|'t|'re|'ve|'m|'ll|'d|[\p{L}]+|[\p{N}]|[^\s\p{L}\p{N}]+"""`, the `'s|'t|'re|'ve|'m|'ll|'d` part splits text like "_you're_" to ["_you_", "_'re_"] given our old implementation, but Python imp gives ["_you_", "_'_", "_re_"].

`skipApostrophes` is intended to bring parity with the regex compiling for Python. I can update the comment, that is not as clear as this explanation.

operators/tokenizer/bpe_kernels.cc
sayanshaw24: same thing in CLIP as there is WS clean in CLIPTokenizer: https://github.com/huggingface/transformers/blob/v4.34.0/src/transformers/models/clip/tokenization_clip.py#L94

test/test_autotokenizer.py
```diff
@@ -3,6 +3,7 @@
 import unittest
 
 import numpy as np
+import ftfy
```
wenbingl, line 6: why do you need import it here? I think it was used by HF tokenizers.

test/test_cliptok.py
```diff
@@ -105,6 +106,11 @@ def test_tokenizer(self):
         self._run_tokenizer(["Testing multiple      sequences       of spaces"])
         self._run_tokenizer(["      in the beginning and the end.      "])
         self._run_tokenizer([" "])
+        
+        # Currently, HF CLIPTokenizer and CLIPTokenizerFast implementations are not consistent with apostrophe handling
+        # until ftfy is installed so we test aposrophe handling for both slow and fast Python implementations.
+        self._run_tokenizer(["Testing words with apostrophes such as you're, i'm, don't, etc."], use_slow=True)
```
wenbingl, line 112: generally, we don't test other components buggy behavior, just make sure ours align with the expected behavior.

test/test_autotokenizer.py
sayanshaw24: yeah but they did not check in that PR so HF natively does not import it so it fails without that import locally - might still pass on the ci but it will cause issues for locally testing in the future so i think it is better to have it in the code.

test/test_cliptok.py
sayanshaw24: done!

test/test_autotokenizer.py
wenbingl: not quite follow here, If you want other be aware that ftfy should be installed to to avoid the test failure. You may check ftfy installation here and raise a warning it is missed.

wenbingl (approved):

PR #576, opened 2023-10-18
Refactor String and Audio operators with status-return prototype.

wenbingl:


operators/text/string_equal.cc
```diff
@@ -0,0 +1,158 @@
...
+      ++p1;
+      ++p3;
+      if (parent->shape2_[last] != 1) {
+        ++p2;
+      }
+      dim = static_cast<int>(last);
+      while (dim > 0 && index1[dim] >= parent->shape1_[dim]) {
+        index1[dim] = 0;
+        if (parent->shape2_[dim] != 1) {
+          p2 -= parent->cum_shape2_[dim] * parent->shape2_[dim];
```
sayanshaw24, line 82: is this implementation based on some existing framework? do you have a paper or docs link i can look at for better understanding (same for the audio operators)?

sayanshaw24 (approved):

operators/text/string_equal.cc
wenbingl: I don't think it's based on any existing framework. There might be some link in the source files which lead to TF equivalent which may help understanding.

operators/text/op_ragged_tensor.cc
```diff
@@ -85,35 +71,31 @@ void KernelRaggedTensoroDense::Compute(const ortc::Tensor<int64_t>& input0,
...
-void KernelStringRaggedTensoroDense::Compute(const ortc::Tensor<int64_t>& input0,
-                                             const ortc::Tensor<std::string>& input1,
-                                             const ortc::Tensor<int64_t>& input2,
-                                             const ortc::Tensor<std::string>& input3,
-                                             ortc::Tensor<std::string>& output) const {
-  // const OrtValue* inputs[4];
-  // OrtTensorDimensions dims[4];
-
+OrtStatusPtr StringRaggedTensorToDense(const ortc::Tensor<int64_t>& input0,
+                                                     const ortc::Tensor<std::string>& input1,
```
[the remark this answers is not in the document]
wenbingl: will do it in the next PR.

operators/text/string_equal.cc
sayanshaw24: i see, perfect, thanks!

PR #578, opened 2023-10-19
Remove ftfy imports and add warning if not installed

sayanshaw24:
ftfy is required for parity between CLIPTokenizer and CLIPTokenizerFast.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,7 +1,6 @@
 # include requirements.txt so pip has context to avoid installing incompatible dependencies
 -r requirements.txt
 pytest
-ftfy
 # multiple versions of onnxruntime are supported, but only one can be installed at a time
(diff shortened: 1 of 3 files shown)


requirements-dev.txt
```diff
@@ -1,7 +1,6 @@
 # include requirements.txt so pip has context to avoid installing incompatible dependencies
 -r requirements.txt
 pytest
-ftfy
```
wenbingl, line 4: without ftfy, slow HF tokenizer can still the correct result in CI pipeline?

wenbingl (approved):

requirements-dev.txt
sayanshaw24: yes, but locally it will throw issues if it's not installed - hence the warning so when we're developing we don't have issues.

PR #581, opened 2023-10-23
Add CUDA build support and some code refinements

wenbingl:


operators/math/cuda/negpos.cu
wenbingl, line 1: @souptc , @RandySheriffH , it looks like we at least need 4 files to write a simplest cuda kernel due to the limitation of nvcc, is there any good idea for this?

includes/onnxruntime_customop.hpp
```diff
@@ -18,6 +18,41 @@
...
+};
+
+
+namespace Ort {
+
+namespace Custom {
+
+struct CudaContext : public CustomOpContext {
+  cudaStream_t cuda_stream = {};
+  struct {}   *cudnn_handle = {};
```
wenbingl, line 46: it looks like it is not a sustainable way to keep add handles here, since every cuda lib has it's own handle.

wenbingl, line 46: using struct {} * instead of  cudnn_handle_t to avoid installing cudnn mandatorily.

CMakeLists.txt
```diff
@@ -273,6 +275,13 @@ macro(standardize_output_folder bin_target)
     PDB_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin")
 endmacro()
 
+if(OCOS_USE_CUDA)
+  include(ext_cuda)
```
sayanshaw24, line 279: can we check if the system arch has GPU here and throw an error? so if a user sets OCOS_USE_CUDA to ON and tries building (accidentally) on a machine without a GPU they won't get failures they don't realize the origin for (similar to when a user tries to use ort gpu functionality without having onnxruntime-gpu installed).

CMakeLists.txt
```diff
@@ -324,6 +333,11 @@ if(OCOS_ENABLE_MATH)
   endif()
 
   file(GLOB TARGET_SRC_MATH "operators/math/*.cc" "operators/math/*.h*")
+  if(OCOS_USED_CUDA)
+    file(GLOB TARGET_SRC_MATH_CUDA "operators/math/cuda/*.*")
```
sayanshaw24, line 337: do we need to add onnxruntime-gpu to requirements.txt or does ort take care of that if we use CUDA kernels?

wejoncy:
Hi @wenbingl  Is it possible to support a general Cuda Op? This op is only responsible for converting tensor to DLPACK and collect output tensor.
The benefits are that we don't have to write a specific kernel for a specific op., and we could free the flexibility to user to support any kinds of customized operations just like what torch.extentions does.

wenbingl:
> Hi @wenbingl Is it possible to support a general Cuda Op? This op is only responsible for converting tensor to DLPACK and collect output tensor. The benefits are that we don't have to write a specific kernel for a specific op., and we could free the flexibility to user to support any kinds of customized operations just like what torch.extentions does.

The current goal is to implement a kernel just like we did in onnxruntime. Are you suggesting how we can support the CUDA custom op in Python, which hasn't been implemented yet.

wejoncy:
> > Hi @wenbingl Is it possible to support a general Cuda Op? This op is only responsible for converting tensor to DLPACK and collect output tensor. The benefits are that we don't have to write a specific kernel for a specific op., and we could free the flexibility to user to support any kinds of customized operations just like what torch.extentions does.
> 
> The current goal is to implement a kernel just like we did in onnxruntime. Are you suggesting how we can support the CUDA custom op in Python, which hasn't been implemented yet?

Yes, It seems quite useful if we have such a mechanism to support user customized cuda kernel but not touch python inside the graph as what ort-training did (python-op).
With this, users can easily add a custom kernel support by registering a kernel function pointer to ORT, and ort can just call the function pointer `(signautre is (DLTensor* tensors, int tensor_lengths, const char* function_name))`. This will allow us easily to support all torch extension Ops, TVM ops or any user implemented ops without needing to merge these kernels to ORT/ORT-ext.

wenbingl:
> > > Hi @wenbingl Is it possible to support a general Cuda Op? This op is only responsible for converting tensor to DLPACK and collect output tensor. The benefits are that we don't have to write a specific kernel for a specific op., and we could free the flexibility to user to support any kinds of customized operations just like what torch.extentions does.
> > 
> > 
> > The current goal is to implement a kernel just like we did in onnxruntime. Are you suggesting how we can support the CUDA custom op in Python, which hasn't been implemented yet?
> 
> Yes, It seems quite useful if we have such a mechanism to support user customized cuda kernel but no touch python inside the graph as what ort-training did (python-op). With this, users can easily add a custom kernel support by registering a kernel function pointer to ORT, and ort can just call the function pointer `(signautre is (DLTensor* tensors, int tensor_lengths, const char* function_name))`. This will allow us easily to support all torch extension Ops, TVM ops or any user implemented ops without needing to merge these kernels to ORT/ORT-ext.

It's a good idea and it's another 'extensions' for the current custom op mechanism. Do you have any use case or a good torch extensions/TVM ops that can be integrated?

wejoncy:
> It's a good idea and it's another 'extensions' for the current custom op mechanism. Do you have any use case or a good torch extensions/TVM ops that can be integrated?
Yes, (PagedAttention)[https://github.com/vllm-project/vllm] , (Flashattention)[https://github.com/Dao-AILab/flash-attention]
highly optimized w4a16/w8a8 kernels for Matmul in TVM/TensorRT-LLM. 
I know it's possible to port those cuda code into ORT, but it's really hard to maintain.

Especially, all these kernels can comunicate with ORT by DLPACK Tensor.

CMakeLists.txt
wenbingl: that's Python support issue.
Moving forward, we don't specify any dependency on ort or ort-gpu since we don't know which one the user's preference is.

CMakeLists.txt
wenbingl: On building, it doesn't require CUDA, instead of CUDA runtime libraries.

operators/math/cuda/negpos.cu
RandySheriffH: Yes, let's keep it this way for bypass NVCC incompatibilities for now.

RandySheriffH (approved):

PR #583, opened 2023-10-26
Make `token_type_ids` optional output for HFBertTokenizer op

YUNQIUGUO:
As title.


operators/tokenizer/bert_tokenizer.cc
```diff
@@ -407,18 +407,23 @@ void KernelHfBertTokenizer::Compute(const ortc::Tensor<std::string>& input,
 
   std::list<OffsetMappingType> offset_map;
 
-  // Only compute offset mapping if optional output for it exists.
+  // Only compute offset mapping for optional output if it exists.
```
sayanshaw24, line 410: nit: i think this comment is confusing - so we're not computing offset mapping _for_ an optional output right - we're computing it if the optional output asking for it exists. perhaps if the original comment is unclear as well, we can change it to something like "Offset mapping is an optional output - we only compute it if the user demands it, i.e. `compute_offset_mapping` is `true`"

test/shared_test/test_ortops_hfberttokenizer.cc
```diff
@@ -141,3 +141,104 @@ TEST(hf_bert_tokenizer_opertor, test_default) {
...
+      {TestValue{"text", {"Where does John's friend live?", context2}, input_dims}},
+      {TestValue{"text", {"Which state does John's friend live?", context2}, input_dims}}};
+
+  std::vector<std::vector<TestValue>> outputs = {
+      {TestValue{
+           "input_ids",
+           std::vector<int64_t>{101, 2040, 2003, 2198, 1005, 1055, 2905, 1029, 102, 2198, 2003, 1037, 2184, 2095, 2214,
+                                2879, 1012, 2002, 2003, 1996, 2365, 1997, 2728, 3044, 1012, 3870, 4482, 2003, 2728,
+                                1005, 1055, 2564, 1012, 2016, 12011, 2012, 15384, 8256, 1012, 9665, 3044, 2003, 3870,
+                                1005, 1055, 2684, 1012, 2016, 2913, 2012, 15384, 4482, 102},
```
sayanshaw24, line 180: could we potentially first use a Python script to get the input ids as per the HF tokenizer and write them into a temp file, read them from that file in this C++ file to compare with the output for the tests, and finally delete the temp files?

i think that will be a lot cleaner and will also make sure we don't miss any changed functionality in the future if HF introduces any changes which result in different input ids (e.g. from vocab/merges handling changes) and help catch such failures.

@wenbingl thoughts?

operators/tokenizer/bert_tokenizer.cc
YUNQIUGUO: ah okay got it now. will update the comment here. thanks!

test/shared_test/test_ortops_hfberttokenizer.cc
wenbingl: that's ideal case, but they run these unit test on mobile platforms, which don't support that.

operators/tokenizer/bert_tokenizer.cc
```diff
@@ -396,7 +396,7 @@ KernelHfBertTokenizer::KernelHfBertTokenizer(const OrtApi& api, const OrtKernelI
 void KernelHfBertTokenizer::Compute(const ortc::Tensor<std::string>& input,
```
wenbingl, line 396: a side question, why use KernelHfBertTokenizer, instead of KernelBertTokenizer?

onnxruntime_extensions/tools/pre_post_processing/steps/nlp.py
```diff
@@ -180,7 +180,7 @@ def _vocab_to_dict(vocab_or_file: Union[Dict[str, int], Path, str]):
 
 
 class BertTokenizer(Step):
-    def __init__(self, tokenizer_param: TokenizerParam, name: Optional[str] = None):
+    def __init__(self, tokenizer_param: TokenizerParam, last_output_optional: bool = False, name: Optional[str] = None):
```
skottmckay, line 183: nit: `need_token_type_ids_output` would be clearer.

operators/tokenizer/bert_tokenizer.cc
```diff
@@ -396,7 +396,7 @@ KernelHfBertTokenizer::KernelHfBertTokenizer(const OrtApi& api, const OrtKernelI
...
                                     ortc::Tensor<int64_t>& output,
                                     ortc::Tensor<int64_t>& output1,
-                                    ortc::Tensor<int64_t>& output2,
+                                    std::optional<ortc::Tensor<int64_t>*> output2,
```
skottmckay, line 399: nit: can we give the outputs meaningful names? `output`, `output1` and `output2` requires the reader to unnecessarily go through the implementation to figure out what each one is.

operators/tokenizer/bert_tokenizer.cc
```diff
@@ -407,18 +407,23 @@ void KernelHfBertTokenizer::Compute(const ortc::Tensor<std::string>& input,
...
   bool compute_offset_mapping = false;
   if (offset_mapping.has_value()) {
     compute_offset_mapping = true;
   }
 
+  // Only allocate token_type_ids if optional output exists.
+  bool generate_token_type_ids = false;
+  if (output2.has_value()) {
+    generate_token_type_ids = true;
+  }
```
skottmckay, line 420:
```suggestion
  bool generate_token_type_ids = output2.has_value();
```

test/shared_test/test_ortops_hfberttokenizer.cc
```diff
@@ -141,3 +141,104 @@ TEST(hf_bert_tokenizer_opertor, test_default) {
     TestInference(*ort_env, model_path.c_str(), inputs[i], outputs[i]);
   }
 }
+
+TEST(hf_bert_tokenizer_operator, test_optional_output) {
```
skottmckay, line 145: nit: instead of cut-and-pasting so much code it would be nicer to refactor the test logic into a helper function that had a bool for producing/validating the token_type_ids output.

operators/tokenizer/bert_tokenizer.cc
YUNQIUGUO: using HFBertTokenizer here as the model I am adding ppp to are from hugging face distilbert-based models.

operators/tokenizer/bert_tokenizer.cc
YUNQIUGUO: added comments for the outputs here instead. hopefully that indicates the meaning more clearer. as we already have those named variables in the code below.

skottmckay (approved):
:shipit:

PR #590, opened 2023-11-02
Add BPE cache to improve efficiency of BPE Tokenizers

sayanshaw24:
HF also implements a cache for BPE: https://github.com/huggingface/transformers/blob/6f316016877197014193b9463b2fd39fa8f0c8e4/src/transformers/models/gpt2/tokenization_gpt2.py#L216C6-L216C6

We use a LRU cache algorithm for the same in C++.

### DIFF
--- a/operators/tokenizer/bpe_kernels.cc
+++ b/operators/tokenizer/bpe_kernels.cc
@@ -132,6 +132,7 @@ std::vector<int64_t> KernelBpeTokenizer::Tokenize(ustring& input,
                                                   std::list<OffsetMappingType>& offset_map) const {
   std::vector<int64_t> res;
   std::list<std::pair<uint32_t, uint32_t>> byte_list;
+  LRUCache bpe_cache = LRUCache(500);
 
   bool clean_up_spaces = false;
   if (ModelName() == BpeModelConf::kModel_CLIP) {
@@ -205,6 +206,7 @@ std::vector<int64_t> KernelBpeTokenizer::Tokenize(ustring& input,
       std::string utf8_token = std::string(ustring(tok));
 
       size_t space_dif = 0;
+
       if (compute_offset_mapping) {
         // Handle special case for offset mapping
         if (utf8_token.at(0) == ' ') {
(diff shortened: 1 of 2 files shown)


operators/tokenizer/bpe_utils.hpp
```diff
@@ -230,3 +230,70 @@ class TokenWithRegularExp {
  private:
   std::u32string_view m_text;
 };
+
+class LRUCache {
+  // Store keys of cache
+  std::list<std::string> dq;
+
+  // Store references of keys in cache for efficiency
+  std::unordered_map<std::string, std::list<std::string>::iterator> references;
```
wenbingl, line 239: why is the map unordered?

operators/tokenizer/bpe_utils.hpp
```diff
@@ -230,3 +230,70 @@ class TokenWithRegularExp {
...
+  
+  // Maximum capacity of cache
+  int capacity;
+
+ public:
+  // Declare the size
+  LRUCache(int n) { capacity = n; }
+
+  // Add tok to the LRU cache
+  void add(std::string tok, std::list<std::pair<uint32_t, uint32_t>> output) {
```
wenbingl, line 252: const std::string&

operators/tokenizer/bpe_utils.hpp
```diff
@@ -230,3 +230,70 @@ class TokenWithRegularExp {
...
+      input_ids_and_offsets.insert(out);
+    }
+
+    // update keys and references
+    dq.push_front(tok);
+    references[tok] = dq.begin();
+  }
+
+  // Check if token is already tokenized
+  bool already_tokenized(std::string tok) {
```
wenbingl, line 284: the  method naming style.

operators/tokenizer/bpe_kernels.cc
```diff
@@ -133,6 +133,14 @@ std::vector<int64_t> KernelBpeTokenizer::Tokenize(ustring& input,
   std::vector<int64_t> res;
   std::list<std::pair<uint32_t, uint32_t>> byte_list;
 
+  // HF implements a cache for BPE:
+  // https://github.com/huggingface/transformers/blob/6f316016877197014193b9463b2fd39fa8f0c8e4/src/transformers/models/gpt2/tokenization_gpt2.py#L216C6-L216C6
+
+  // We use a LRU cache algorithm for the same in C++ in order to save compute.
+
+  // Current cache capacity is set to a relatively small 500 in order to support mobile platforms.
+  LRUCache bpe_cache = LRUCache(500);
```
wenbingl, line 142: can it be used for next Compute call?

wenbingl (commented):
Need to measure the perf more since there are some extra efforts around bpe method now.

operators/tokenizer/bpe_utils.hpp
sayanshaw24: because that one is just for the iterators for the keys (address of keys). `dq` is what we are using as a queue for the Queue+Map LRU implementation.

operators/tokenizer/bpe_utils.hpp
sayanshaw24: done!

operators/tokenizer/bpe_utils.hpp
sayanshaw24: fixed!

operators/tokenizer/bpe_kernels.cc
sayanshaw24: no - if we move it outside to use it for the next Compute call it throws errors because the type of BPE tokenizer (GPT2/CLIP/Roberta) is different on different calls and the tests fail because ids/offsets are wrong.

sayanshaw24:
> Need to measure the perf more since there are some extra efforts around bpe method now.

Added an average tokenization time in microseconds using `std::chrono` that is currently printed out in `Compute` after completing all calls to `Tokenize`. In the future, we can refactor this to make it cleaner and potentially even add an optional output for it.

operators/tokenizer/bpe_kernels.cc
```diff
@@ -130,6 +135,17 @@ std::vector<int64_t> KernelBpeTokenizer::Tokenize(ustring& input,
...
+
+  auto start = std::chrono::high_resolution_clock::now();
+
+  // HF implements a cache for BPE:
+  // https://github.com/huggingface/transformers/blob/6f316016877197014193b9463b2fd39fa8f0c8e4/src/transformers/models/gpt2/tokenization_gpt2.py#L216C6-L216C6
+
+  // We use a LRU cache algorithm for the same in C++ in order to save compute.
+
+  // Current cache capacity is set to a relatively small 500 in order to support mobile platforms.
+  LRUCache bpe_cache = LRUCache(500);
```
wenbingl, line 147: Add a flag to enable/disable cache feature

operators/tokenizer/bpe_kernels.cc
```diff
@@ -26,6 +27,10 @@ const char BpeModelConf::kModel_GPT2[] = "GPT2";
 const char BpeModelConf::kModel_Roberta[] = "Roberta";
 const char BpeModelConf::kModel_CLIP[] = "CLIP";
 
+// We specifically measure performance for tokenization as our BPE implementation includes a number of optimizations
+long long total_tokenization_time = 0.0;
```
wenbingl, line 31: these time measurement code should be removed.

operators/tokenizer/bpe_utils.hpp
```diff
@@ -230,3 +230,70 @@ class TokenWithRegularExp {
  private:
   std::u32string_view m_text;
 };
+
+class LRUCache {
+  // Store keys of cache
+  std::list<std::string> dq;
```
wenbingl, line 236: given it's a limited capacity cache, the container can be a vector here.

operators/tokenizer/bpe_utils.hpp
```diff
@@ -230,3 +230,70 @@ class TokenWithRegularExp {
  private:
   std::u32string_view m_text;
 };
+
+class LRUCache {
```
wenbingl, line 234: Can you find some high-quality C++ LRUCahe in GitHub for a reference?

PR #591, opened 2023-11-02
Support 'added_token' attribute for BPE tokenizer and some code refactoring.

wenbingl:
1. added_token attribute, which is similar to HF's tokenizer
2. BPE code refactoring for status return style.
3. some minor issues fixing.

### DIFF
--- a/operators/tokenizer/bpe_tokenizer.hpp
+++ b/operators/tokenizer/bpe_tokenizer.hpp
@@ -90,15 +90,19 @@ class BpeModel {
     return nullptr;
   }
 
-  OrtStatusPtr LoadAddedTokens(const char* added_tokens) {
-    std::istringstream istrea(added_tokens);
+  OrtStatusPtr LoadAddedTokens(const char* added_tokens, int64_t id_begin) {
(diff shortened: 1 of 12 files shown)


operators/tokenizer/bpe_tokenizer.hpp
```diff
@@ -80,8 +86,62 @@ class BpeModel {
...
+      if (i > id2token_map_.size()) {
+        id2token_map_.resize(i + 1);
+      }
       id2token_map_[i] = t;
     }
+
+    return nullptr;
+  }
+
+  OrtStatusPtr LoadAddedTokens(const char* added_tokens) {
```
sayanshaw24, line 101: is there some HF reference for this?

sayanshaw24 (approved):

operators/tokenizer/bpe_tokenizer.hpp
wenbingl: yes, check here, https://github.com/huggingface/transformers/blob/cc3e4781854a52cf090ffde28d884a527dab6708/src/transformers/tokenization_utils.py#L52,

Issue #584, opened 2023-10-30
Likely a typo


axelfurlan:
https://github.com/microsoft/onnxruntime-extensions/blob/646962072d397b87e1b48b87d3a92b8c3d3fa1ad/onnxruntime_extensions/tools/add_pre_post_processing_to_model.py#L289

Shouldn't this be `new_model` instead of `new_mode`?

skottmckay:
It actually shouldn't, but the name should be clearer and there could be a comment clarifying what we're doing.

We want to run the shape inferencing to validate the model (could add `strict=True`) but do not want to save that model as it will be unnecessarily large with all the shape information (which obviously can be inferred).

axelfurlan:
Fixed in https://github.com/microsoft/onnxruntime-extensions/pull/595

axelfurlan:
Fixed in https://github.com/microsoft/onnxruntime-extensions/pull/595

PR #595, opened 2023-11-07
Fixed a bug (typo) in yolo_detection

dizcza:
I'm only learning onnxruntime extension but this seems to me like a typo.

### DIFF
--- a/onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
+++ b/onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
@@ -286,7 +286,7 @@ Because we need to execute the model to determine the output shape in order to a
     pipeline.add_post_processing(post_processing_steps)
 
     new_model = pipeline.run(model)
-    new_mode = onnx.shape_inference.infer_shapes(new_model)
+    new_model = onnx.shape_inference.infer_shapes(new_model)
     onnx.save_model(new_model, str(output_file.resolve()))
 
 


dizcza:
This, however, doesn't fix the issue I'm struggling with:

```python
from pathlib import Path
from onnxruntime_extensions.tools.add_pre_post_processing_to_model import yolo_detection

model_path = Path("yolov8n-pose.onnx")
new_model_path = model_path.with_name("yolov8n-pose_nms.onnx")

yolo_detection(model_path, new_model_path)

# ----------------

import numpy as np
import onnxruntime as ort
from onnxruntime_extensions import get_library_path

so = ort.SessionOptions()
so.register_custom_ops_library(get_library_path())

path = "yolov8n-pose_nms.onnx"

session = ort.InferenceSession(path, so, providers=['CPUExecutionProvider'])

outname = [i.name for i in session.get_outputs()]  # ['image_out']
inname = [i.name for i in session.get_inputs()]    # ['image']

im = np.ones((640, 640, 3), dtype=np.uint8)

inp = {inname[0]: im}
outputs = session.run(outname, inp)
```

```
Traceback (most recent call last):
  File "/home/dizcza/Projects/Airtouch/airpy/onnxruntime_demo.py", line 22, in <module>
    outputs = session.run(outname, inp)
  File "/home/dizcza/miniconda3/envs/mediapipe/lib/python3.9/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 217, in run
    return self._sess.run(output_names, input_feed, run_options)
onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Invalid rank for input: image Got: 3 Expected: 1 Please fix either the inputs or the model.
```

dizcza:
If I flatten the image with `im = im.reshape(-1)` as suggested [here](https://github.com/microsoft/onnxruntime/issues/16557#issuecomment-1643725056), I get another error:

```
2023-11-07 17:30:09.644952955 [E:onnxruntime:, sequential_executor.cc:514 ExecuteKernel] Non-zero status code returned while running DecodeImage node. Name:'pre_process_0' Status Message: 2: [DecodeImage] Invalid input. Failed to decode image.
Traceback (most recent call last):
  File "/home/dizcza/Projects/Airtouch/airpy/onnxruntime_demo.py", line 22, in <module>
    outputs = session.run(outname, inp)
  File "/home/dizcza/miniconda3/envs/mediapipe/lib/python3.9/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 217, in run
    return self._sess.run(output_names, input_feed, run_options)
onnxruntime.capi.onnxruntime_pybind11_state.RuntimeException: [ONNXRuntimeError] : 6 : RUNTIME_EXCEPTION : Non-zero status code returned while running DecodeImage node. Name:'pre_process_0' Status Message: 2: [DecodeImage] Invalid input. Failed to decode image.
```

dizcza:
All right, apart from a strange ubyte input complaint, I think I cannot simply apply the same post-processing steps to the YoloPose model (not YoloDetect).

Long story short, how to merge NMS as a post-processing step to the YoloPose v8 model?

skottmckay:
You can see the individual pre/post processing steps used for an object detection yolo model [here](https://github.com/microsoft/onnxruntime-extensions/blob/31d2f64786f624be44355241f8740a7b32e212a2/onnxruntime_extensions/tools/add_pre_post_processing_to_model.py#L255-L284) 

Whilst most of those steps are applicable, there are some differences for pose models that require a slightly different pipeline and some changes to existing steps to support the keypoint data.

For reference:
- Object detection post-processing
  - https://github.com/ultralytics/ultralytics/blob/e7bd159a44cf7426c0f33ed9b413ef4439505a03/ultralytics/models/yolo/detect/predict.py#L23
- Pose post-processing: 
  - https://github.com/ultralytics/ultralytics/blob/e7bd159a44cf7426c0f33ed9b413ef4439505a03/ultralytics/models/yolo/pose/predict.py#L31

Differences:

For object detection there are 80 classes and no mask data (guessing 'mask' originally comes from a segmentation model usage but in this case that's where the keypoint data is).
For pose there's 1 class and 51 items of mask data (17 keypoints x 3 values for each). 

Detection:
image

Pose:
image
 
To handle the differences SplitOutBoxAndScore and SelectBestBoundingBoxesByNMS need to be augmented to pass through the mask data. would probably require making some things configuration based. e.g. inferring whether there is an object confidence score or not (used in older yolo detection models) gets messy if you also want to infer the number of classes/mask size, and it's simpler to explicitly specify all those values when configuring the post-processing step. 

There are some additional changes required to extract the key points from the mask data. That requires plugging in their shape. For this particular model it is [17, 3] as there are 17 keypoints (nose, eyes etc.) with x, y and a confidence score. Other models could have more/less keypoints and not have thhe confidence score. The x and y values from the keypoints need to be scaled as well (similar to what ScaleBoundingBoxes does). 

https://github.com/ultralytics/ultralytics/blob/e7bd159a44cf7426c0f33ed9b413ef4439505a03/ultralytics/models/yolo/pose/predict.py#L48-L49

Those additional changes would need to be implemented as new post-processing steps, but maybe you don't need that to be done in the model at this point.

dizcza:
@skottmckay thanks for the detailed response.

> Those additional changes would need to be implemented as new post-processing steps, but maybe you don't need that to be done in the model at this point.

But that's precisely what I want: I'm running YoloPose v8 on Android and OpenCV backend gives me 1.5 FPS - way too slow. To speed things up, I need to
1. quantize the model
2. merge NMS to the graph (currently I'm doing NMS as it's done [here](https://github.com/mallumoSK/yolov8/blob/feed33efc6f95c139f89d05c8c7edd4cebbcfea0/yolo/YoloPose.cpp#L32))

My naive idea was that I could simply borrow your script for yolo detection - I was wrong: it seems like a lot of work.

dizcza:
Anyway, yolo pose post processing is a feature request.

And this PR just fixes a negligible issue.

wenbingl (approved):

skottmckay:
> Anyway, yolo pose post processing is a feature request.
> 
> And this PR just fixes a negligible issue.

@dizcza The change actually creates a different issue: https://github.com/microsoft/onnxruntime-extensions/issues/584#issuecomment-1800741416

The naming was bad and there could have been comments explaining things better. 

Can you describe your intended usage to help prioritize adding support for the pose model to the post processing? Is this for a production scenario and if so is there a timeline?

dizcza:
> @dizcza The change actually creates a different issue: [#584 (comment)](https://github.com/microsoft/onnxruntime-extensions/issues/584#issuecomment-1800741416)
> 
> The naming was bad and there could have been comments explaining things better.

Oh, so this PR should have been closed then.

> Can you describe your intended usage to help prioritize adding support for the pose model to the post processing? Is this for a production scenario and if so is there a timeline?

I'm using the yolov8 pose to monitor people interacting with the camera, mostly kids. I need to parse their poses, extract hand-relative orientations, and do some logic. The app runs on Windows, Linux, and Android. The Android part is the most challenging since a straightforward approach brings only 1.5 FPS - that's why I started looking into your repository specifically for (1) quantization and (2) NMS included in the model to get rid of manually parsing a large tensor and then calling `cv::dnn::NMSBoxes`. And NMS is the only post-processing step I'm interested in.

This is for a production scenario.

I think I can do quantization without any (significant) modification of your code.

skottmckay:
We'll look at adding the post-processing changes to simplify usage. Possibly that will also cover most if not all of the new segmentation yolo model post-processing as well as a bonus. 

FWIW in the next ORT release we're look at adding the one missing operator for the NNAPI EP to have full coverage of the model. Using the NNAPI EP _could_ give better performance, but NNAPI performance is highly unpredictable from device to device. Also adding the 2 missing operators for CoreML to have complete coverage of the model and CoreML tends to give a far more consistent performance gain.  

IIRC generally the internals of the model are quantized and the model input/output would still be fp32, so hopefully the pre/post processing can be used as-is.

wejoncy:
> This, however, doesn't fix the issue I'm struggling with:
> 
> ```python
> from pathlib import Path
> from onnxruntime_extensions.tools.add_pre_post_processing_to_model import yolo_detection
> 
> model_path = Path("yolov8n-pose.onnx")
> new_model_path = model_path.with_name("yolov8n-pose_nms.onnx")
> 
> yolo_detection(model_path, new_model_path)
> 
> # ----------------
> 
> import numpy as np
> import onnxruntime as ort
> from onnxruntime_extensions import get_library_path

> Traceback (most recent call last):
>   File "/home/dizcza/Projects/Airtouch/airpy/onnxruntime_demo.py", line 22, in <module>
>     outputs = session.run(outname, inp)
>   File "/home/dizcza/miniconda3/envs/mediapipe/lib/python3.9/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 217, in run
>     return self._sess.run(output_names, input_feed, run_options)
> onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Invalid rank for input: image Got: 3 Expected: 1 Please fix either the inputs or the model.
> ```

Hi, this is as we add a pre-process step in the model so that we can only accept the raw jpg/png image as input.
`open('xx.jpg').read()`

dizcza:
> Hi, this is as we add a pre-process step in the model so that we can only accept the raw jpg/png image as input. `open('xx.jpg').read()`

Thanks, I thought that "raw bytes" means numpy `ubyte` data type. But I see how convenient it gets in [Android native C++](https://github.com/microsoft/onnxruntime-inference-examples/blob/afc83c003c2b262929d1667d8934021bfde8f394/mobile/examples/object_detection/android/app/src/main/java/ai/onnxruntime/example/objectdetection/ObjectDetector.kt#L22-L33).

BTW, an ONNX Android with Native C++ integration example is missing, I'm really struggling collecting bits and pieces for such a use-case. And, of course, I want to reuse already working C++ ONNX backend code in Android. But this is a question for ONNX guys, I'll ask one somewhere else.

skottmckay:
What would be different about Android C++ vs any other C++ usage? I believe the AAR has the C++ headers and the native shared library (libonnxruntime.so) is separate from the jni layer (libonnxruntime4j_jni.so).

dizcza:
> What would be different about Android C++ vs any other C++ usage? I believe the AAR has the C++ headers and the native shared library (libonnxruntime.so) is separate from the jni layer (libonnxruntime4j_jni.so).

I'm talking about
1. a complete example of how to setup an Android project with ONNX native C++ bindings. Such things are always tricky unless you do it every day. I found only one working unofficial example https://github.com/Rohithkvsp/OnnxRuntimeAndorid
2. limitations of ONNX Java bindings. See this issue: https://github.com/microsoft/onnxruntime/issues/18369

skottmckay:
Typically we'd expect an Android app developer to not be writing C++ directly given ORT has Java bindings. Is there a significant performance advantage to not using the ORT Java bindings?

dizcza:
> Typically we'd expect an Android app developer to not be writing C++ directly given ORT has Java bindings.

Why? Because Java devs are afraid of C++? :) Won't you say such a thing to Android devs (including myself), who have lots of OpenCV code in C++, and all of a sudden they are forced to rewrite their C++ code to Java if OpenCV stops providing an interface to pass all their objects as pointers from Java to C++? We'd rather not support Android platforms at all then...

> Is there a significant performance advantage to not using the ORT Java bindings?

Nope. The only reason I'd like to use ONNX in C++ on Android is to reuse the working code I have on C++. Luckily, there is not much code of ONNX in C++ and I think I can port it to Java with some effort.

skottmckay:
Craigacp is the expert on the Java bindings so let's see what he says about exposing the native handler. 

Worst case is it not possible to load the model as bytes in Java and pass that down to the C++ level to use ORT from then on? e.g. https://github.com/microsoft/onnxruntime-inference-examples/blob/afc83c003c2b262929d1667d8934021bfde8f394/mobile/examples/object_detection/android/app/src/main/java/ai/onnxruntime/example/objectdetection/MainActivity.kt#L98

dizcza:
> Worst case is it not possible to load the model as bytes in Java and pass that down to the C++ level to use ORT from then on? e.g. https://github.com/microsoft/onnxruntime-inference-examples/blob/afc83c003c2b262929d1667d8934021bfde8f394/mobile/examples/object_detection/android/app/src/main/java/ai/onnxruntime/example/objectdetection/MainActivity.kt#L98

I cannot do this because neither OrtEnv nor SessionOptions have the getNativeHandle() function public. Citing myself, 
> Even if they had, why not provide the same functionality, getNativeHandle(), for OrtSession in the first place?

dizcza:
Sorry, I got your point. You suggest creating an environment and session options in C++... yeah, that could and should work.

dizcza:
With such an approach, you're really narrowing down the usage. I don't see any reason why the getNativeHandle function should not be public for all OrtX Java classes.

skottmckay:
I'm not a Java expert so not sure if there's a reason to keep the function internal. I'm guessing the approach was to keep as much internal as possible to minimize the surface area of misuse (misuse == github issues == time to investigate) and maximize the flexibility of the implementation (anything not exposed can be changed at will). 

if this is a valid use case where exposing the handle simplifies things for a developer who wants to directly use C++ we can definitely consider exposing it. however if the only reason it's required is to read a raw resource the cost/benefit of exposing it may fall on the side or more potential cost than benefit if you can read the model in java and pass it down to the C++ layer and everything else is done there.

skottmckay:
@dizcza This PR should provide the necessary pieces to run NMS in the model and return bounding boxes and keypoints scaled to the original image.

https://github.com/microsoft/onnxruntime-extensions/pull/605

I have manually tested. Need to add unit tests.

I _think_ you should be able to take vision.py and general.py from the PR and drop them into an existing installed onnxruntime-extensions python package under the tools/pre_post_processing/steps directory, replacing the old versions. After that tutorials/yolov8_pose_e2e.py from the PR would hopefully work as-is.

Alternatively clone the repo, checkout the PR branch, and run `pip install -e .` from the top level directory of the repo for the python package to point to the repo.

We don't have a custom operator to draw the keypoints, so default is to output essentially the same data as the original model for the selected bounding boxes with all the co-ordinates scaled to match the original image. Details in the comments in the tutorial. Only difference is the value for the selected class is included in the output, which is redundant when there's only one class, so there's 57 elements instead of the 56 in the exported onnx model. 

Pre-processing supports providing bytes from a jpg or png. You can adjust the pre/post processing pipelines as needed. They're defined directly in the e2e tutorial for simplicity.

dizcza:
@skottmckay wow, that's quite a lot of work!

I tested it, and it works as expected. I have two questions then:

1. Would it work out of the box with the C++ Onnxruntime extensions, as described on the official docs https://onnxruntime.ai/docs/extensions/#c?
2. How to feed `uint8` frames of shape `(3, W, H)` from a videocapture as the input to this exported model?

dizcza:
> How to feed uint8 frames of shape (3, W, H) from a videocapture as the input to this exported model?

I think in this case I don't need the first step, `ConvertImageToBGR`, in the pre-processing steps. But the following steps refer to this node and substituting them for `image`, as the model input name, doesn't help.

skottmckay:
Basically this puts the pre/post processing inside the model, so the C++ code (or GPU/NPU with NNAPI/CoreML) can handle it.

You only need extensions at runtime for custom operators like DecodeImage/EncodeImage/DrawBoundingBoxes. 

If you have already decoded to {C, H, W} you don't need DecodeImage. Assuming you handle any actions using the bounding boxes/key points you don't need any other custom ops. If that is the case, extensions is providing the tools to put the pre/post processing in the model but doing it using ONNX operators, so no runtime dependency on the onnxruntime extensions packages.

Assuming the channel order matches the model (most of RGB but some are BGR) you can 
- change the model input name and shape to whatever you want - line 47
  - the first pre-processing step automatically uses the input name 
  - shape can use names if you don't know the H and W values until runtime
    - e.g. `inputs = [create_named_value("rgb_chw_bytes", onnx.TensorProto.UINT8, [3, "dynamic_H", "dynamic_W"])]`
    - but if you do know the actual shapes it's better to use them
- remove the ConvertImageToBGR
- add `layout=CHW` to the Resize
- LetterBox needs and update to support CHW though. 
  - should be simple enough to add that. I'll see if I can add it to this PR
- obviously remove the ChannelsLastToChannelsFirst step

For the sake of testing out the pre/post processing you could temporarily changing the input to be H, W, C.

skottmckay:
I have hacked up a version that fills most of the gaps and seems to work with CHW byte input. 
- automatically connect first pre-processing step to graph input names
  - all the other steps auto-connect but I overlooked the very first one doesn't
- adds layout awareness to LetterBox and ScaleNMSBoundingBoxesAndKeyPoints
  - both need to know where to get the H and W values from, and LetterBox needs to create the correctly ordered pads
- add Identity step so you can pass in decoded bytes with shape
  -  needs a Step to be able to connect up with ScaleNMSBoundingBoxesAndKeyPoints to determine original image shape
  - you can use this as the first pre-processing step. no perf cost - just allows connecting the shape up with the post-processing

I'll clean it up a bit more next week but hopefully works as-is for now.

dizcza:
All right, I tested your latest commit and was able to feed a decoded RGB image of shape (H, W, C) (I modified the code to this shape format).

When I tested it on C++, however, it always returns an empty tensor of shape `(0, 57)`... but at least it ran without errors. I'll look into this next week.

dizcza:
I find your constants strange:

```
        SelectBestBoundingBoxesByNMS(iou_threshold=0.7, score_threshold=0.25, max_detections=100, has_mask_data=True),
```

I'm using
```
        SelectBestBoundingBoxesByNMS(iou_threshold=0.5, score_threshold=0.70, max_detections=100, has_mask_data=True),
```

Also, `max_detections` is unclear: is it before or after the NMS suppression - is it input to NMS or its output? It should be clarified in the docs.

skottmckay:
Updated to have max_boxes_per_class and max_detections parameters. The latter runs as the last step (if specified) to reduce the total to the best k boxes.

dizcza:
Thanks @skottmckay I think I can use your script as it is now. Compiling NMS into the graph hasn't yielded a significant performance increase, though I haven't tried testing the model on a video with lots of people - maybe in this case I'll see such a performance gain. Of course, the most important thing in including pre- and post-processing steps into a graph is to reduce the code (and headache) to do it manually either in C++ or Android.

The only thing left for me is quantization with ONNX, which I will probably postpone as the CPU desktop performance is already great, and Android inference will not get sped up by a factor of more than 2, which means FPS 1.5 --> 3.0 at most, which is still beyond our accepted range, which forces me to consider Android ML Kit that provides FPS 15.0 with float weights.

skottmckay:
Did you try the XNNPACK EP? It may have some additional ARM fp32 kernels that speed things up. If ML Kit is also using CPU I'd be very surprised that it would be 5 to 10x faster than the equivalent model using ORT. 

There's a PR to fill the 2 missing operators for the model to run on CoreML (Softmax and Split) and we'll try and add the one missing operator for NNAPI this weeek (Split). NNAPI is a bit hit and miss though as performance is very device specific.

dizcza:
> Did you try the XNNPACK EP?

I've just tried XNNPACK and it hasn't yielded any performance gain at all. First of all, the docs at https://onnxruntime.ai/docs/execution-providers/Xnnpack-ExecutionProvider.html#usage are outdated as there is no such public method (`AppendExecutionProvider`) for Android platform. I found the `addXnnpack` function and used it like so:

```kotlin
    private fun createOrtSession(environment: OrtEnvironment): OrtSession {
        val so = SessionOptions()
        val po = mapOf<String, String>()
        so.addXnnpack(po)
        so.setOptimizationLevel(SessionOptions.OptLevel.ALL_OPT)
        return environment.createSession(readModel(), so)
    }
```

>  If ML Kit is also using CPU I'd be very surprised that it would be 5 to 10x faster than the equivalent model using ORT.

I meant ML Kit Pose Android implementation - it's shipped with a completely different model (MediaPipe Poses) that, on second thought, could be already quantized for the end user: just plug and play. It's limited though to one person in an image and that's quite a limitation for our app.

skottmckay:
The documentation is for usage with the C/C++ API and is correct. I didn't realize the Android usage had diverged. I'll get the documentation updated to mention that and see if we can make the Android API a little more equivalent with a more generic `AppendExecutionProvider` function.

If it's [this](https://developers.google.com/ml-kit/vision/pose-detection/android) ML Kit model the input is also significantly smaller than the YOLOv8 model. 640x640 vs 480x360 is nearly 2.5x the amount of data to process. Throw in single detection vs. multiple and the performance difference starts to make more sense.

dizcza:
> If it's [this](https://developers.google.com/ml-kit/vision/pose-detection/android) ML Kit model the input is also significantly smaller than the YOLOv8 model. 640x640 vs 480x360 is nearly 2.5x the amount of data to process. Throw in single detection vs. multiple and the performance difference starts to make more sense.

Yes, I understand: a model with a smaller input layer always wins overall...

What I don't understand is why XNNPACK does not speed up the original Yolo V8 Pose floating point model; I expect the conv and matmul operations to run faster with XNNPACK. I saw a profiling API method for such debugging, I'm not sure if I'm into this though.

skottmckay:
ORT has some [arm64 MLAS kernels](https://github.com/microsoft/onnxruntime/tree/main/onnxruntime/core/mlas/lib/aarch64) covering core things like MatMul/GEMM/Conv. The XNNPACK EP has additional ones, but if the model doesn't require them (i.e. the MLAS kernels cover the majority of compute intensive operators) the performance wouldn't differ much. Both MLAS and XNNPACK are CPU based.

PR #600, opened 2023-11-11
Add Windows GPU CI

mszhanyi:
1. Download the cuda runtime binaries in each run
2. Build with the latest onnxruntime release (1.16.2)
3. Use cuda11.8, sm version 8.6


.pipelines/templates/set_winenv.yml
```diff
@@ -0,0 +1,27 @@
...
+    type: string
+    default: '11.8'
+  - name: SecondaryCUDAVersion
+    type: string
+    default: '12.2'
+
+steps:
+  - ${{ if eq(parameters.DownloadCUDA, 'true') }}:
+      - powershell: |
+          azcopy.exe cp --recursive "https://lotusscus.blob.core.windows.net/models/cuda_sdk/v${{ parameters.PrimaryCUDAVersion }}" $(Agent.TempDirectory)
```
wenbingl, line 17: can cudnn be found the same azure blob as well?

wenbingl (approved):

.pipelines/ci.yml
```diff
@@ -473,6 +473,70 @@ stages:
...
+          nvcc --version
+          where nvcc
+        displayName: check cuda version
+
+      - task: DownloadGitHubRelease@0
+        inputs:
+          connection: 'GitHub - Release'
+          userRepository: 'microsoft/onnxruntime'
+          defaultVersionType: 'specificTag'
+          version: 'v1.16.1'
```
wenbingl, line 501: should it be replaced by ort.version variable?

.pipelines/ci.yml
```diff
@@ -473,6 +473,70 @@ stages:
...
+          itemPattern: '*-win-x64-gpu-1.16.1*'
```
wenbingl, line 502: same as above.

.pipelines/templates/set_winenv.yml
snnn: Yes.

.pipelines/ci.yml
```diff
@@ -473,6 +473,70 @@ stages:
...
       displayName: Run python test
 
+- stage: WindowsCUDABuilds
+  dependsOn: []
+  jobs:
+  - job: WindowsC
+    pool:
+      name: 'onnxruntime-extensions-Win2022-GPU-A10'
+    variables:
+      ort.version: '1.16.1'
```
snnn, line 483: I suggest making a rule: all variable names must only use upper case letters and underscore . Because all such variables will be converted to environment variables. If you have a pipeline variable `ort.version`, it will become an env var `ORT_VERSION` that can be accessed from any build step. However, the two names are different. The differences could cause confusions.  see https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#environment-variables for more details.

.pipelines/ci.yml
```diff
@@ -473,6 +473,70 @@ stages:
...
+
+          @echo %vsdevcmd% will be used as the VC compiler
+          @echo ##vso[task.setvariable variable=vsdevcmd]%vsdevcmd%
+        displayName: 'locate vsdevcmd via vswhere'
+
+      - script: |
+          call $(vsdevcmd)
+          set CUDA_PATH=$(Agent.TempDirectory)\v11.8
+          set CudaToolkitDir=%CUDA_PATH%
+          xcopy "$(Agent.TempDirectory)\v11.8\extras\visual_studio_integration\MSBuildExtensions\" "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Microsoft\VC\v170\BuildCustomizations" /y
```
snnn, line 531: This is not needed. We'd better not making changes to the system.

.pipelines/ci.yml
```diff
@@ -473,6 +473,70 @@ stages:
...
+          call .\build.bat -DOCOS_ENABLE_CTEST=ON -DOCOS_USE_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES=86 -DOCOS_ONNXRUNTIME_VERSION="$(ort.version)" -DONNXRUNTIME_PKG_DIR=.\onnxruntime-win-x64-gpu-$(ort.version)
```
snnn, line 532: See: https://github.com/microsoft/onnxruntime/blob/main/tools/ci_build/build.py#L2513 
That is how ORT setup cuda home.  If you could follow the same approach, you do not need to install the MSBuildExtensions.

.pipelines/ci.yml
```diff
@@ -473,6 +473,70 @@ stages:
...
+          EnvSetupScript: 'set_env_cuda.bat'
+          DownloadCUDA: true
+
+      - script: |
+          nvidia-smi
+          nvcc --version
+          where nvcc
+        displayName: check cuda version
+
+      - task: DownloadGitHubRelease@0
```
snnn, line 496: Can we get it from Nuget instead? Then we can use a prerelease version when needed. We have a https://aiinfra.visualstudio.com/PublicPackages/_artifacts/feed/ORT-Nightly feed.  It would be great if we can test every ORT release with this pipeline before the release is published.

snnn: Like, first the release manager publish a test package to the ADO feed. Then we manually run this pipeline and specify the test package's version number when we queue the build.

wenbingl: Same thought, but since they are unstable and the fixing could take some time. I'd like to create an CI.optional pipeline to put this job, as well as Androids.

snnn: We can start with using a stable release from nuget.org first. Then we may add another pipeline to build ort-extension with a nightly ORT release.

.pipelines/ci.yml
```diff
@@ -473,6 +473,68 @@ stages:
...
+
+      - task: ExtractFiles@1
+        inputs:
+          archiveFilePatterns: '**/*.zip'
+          destinationFolder: '$(Build.SourcesDirectory)'
+          cleanDestinationFolder: false
+          overwriteExistingFiles: true
+        displayName: Unpack ONNXRuntime package.
+
+      - script: |
```
snnn, line 514: Is this step really needed? CMake can locate Visual Studio by itself. CMake doesn't need this. 
If this step is needed, vsdevcmd.bat is not the right bat file to invoke.  It needs to be  something like:
```yaml
- ${{ if eq(parameters.BuildArch, 'x64') }}:
  - script: |
      @echo off
      set vswherepath="%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe"
      for /f "usebackq delims=" %%i in (`%vswherepath% -latest -property installationPath`) do (
        if exist "%%i\VC\Auxiliary\Build\vcvars64.bat" (
          set vcvarsall="%%i\VC\Auxiliary\Build\vcvars64.bat"
        )
      )
      @echo %vcvarsall% will be used as the VC compiler
      @echo ##vso[task.setvariable variable=vcvarsall]%vcvarsall%
    displayName: 'locate vcvarsall via vswhere'
- ${{ if eq(parameters.BuildArch, 'x86') }}:
  - script: |
      @echo off
      set vswherepath="%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe"
      for /f "usebackq delims=" %%i in (`%vswherepath% -latest -property installationPath`) do (
        if exist "%%i\VC\Auxiliary\Build\vcvars32.bat" (
          set vcvarsall="%%i\VC\Auxiliary\Build\vcvars32.bat"
        )
      )
      @echo %vcvarsall% will be used as the VC compiler
      @echo ##vso[task.setvariable variable=vcvarsall]%vcvarsall%
    displayName: 'locate vcvarsall via vswhere'
```

mszhanyi: No. I've removed this step.

snnn (approved):

PR #603, opened 2023-11-13
Make first cuda op build with cuda context

RandySheriffH:
Make first cuda op build with a working cuda context.


includes/custom_op_lite.h
```diff
@@ -418,6 +418,21 @@ struct Variadic : public TensorBase {
   TensorPtrs tensors_;
 };
 
+#ifdef USE_CUDA
+
+struct CudaContext {
+  void Init(const OrtW::CustomOpApi& api, const OrtKernelContext& ctx) {
+    const auto& ort_api = api.GetOrtApi();
+    ort_api.KernelContext_GetResource(&ctx, 1, 10000, &cuda_stream);
```
wenbingl, line 426: what's 10000?

includes/custom_op_lite.h
```diff
@@ -418,6 +418,21 @@ struct Variadic : public TensorBase {
...
+    if (!cuda_stream) {
+      ORTX_CXX_API_THROW("Failed to fetch cuda stream from context", ORT_RUNTIME_EXCEPTION);
```
wenbingl, line 428: Given now API is fallible, returning a status is a better now.

CMakeLists.txt
```diff
@@ -816,7 +816,6 @@ endif()
 if(OCOS_USE_CUDA)
   target_include_directories(ortcustomops PUBLIC ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES})
   target_link_libraries(extensions_shared PUBLIC cudart cublas cufft)
-  set_target_properties(ocos_operators PROPERTIES CUDA_SEPARABLE_COMPILATION ON)
```
wenbingl, line 819: Do you have some insight about which way is better?

RandySheriffH: The CUDA_SEPARABLE_COMPILATION cause link issues.
Removing it  helper the build pass.

includes/custom_op_lite.h
RandySheriffH: The ::Init() is to be called during “CreateTuple(...)”, which return a **tuple** to a compute function.
We cannot return a status upwards to notify the failture.

wenbingl (dismissed):

includes/custom_op_lite.h
RandySheriffH: Added an enum to specify the resource type.

wenbingl (approved):

PR #605, opened 2023-11-17
Add support for YOLO v8 Pose post-processing.

skottmckay:
The output has additional values in the 'mask' data for the keypoints.

- Update the post processing steps to support extracting and scaling the keypoints.
- Simplify the existing step to split out the boxes and scores by using a basic Split operator if there is no confidence score for a bounding box to apply to the class scores.
  - Confidence score for a bounding box is YOLO versions prior to 8.
- Update existing tests
- Fix issue with NMS output - was picking highest score for box and not the score for the selected class


onnxruntime_extensions/tools/pre_post_processing/docs/pdoc/index.md
```diff
@@ -0,0 +1,371 @@
+Module pdoc
```
wejoncy, line 1: Seems that we don't need this file.

tutorials/yolov8_pose_e2e.py
```diff
@@ -0,0 +1,261 @@
...
+    success = model.export(format="onnx")  # export the model to ONNX format
+    assert success, "Failed to export yolov8n-pose.pt to onnx"
+    import shutil
+    shutil.move(pt_model.with_suffix('.onnx'), onnx_model_name)
+
+
+def add_pre_post_processing_to_yolo(input_model_file: Path, output_model_file: Path,
+                                    output_image: bool = False,
+                                    decode_input: bool = True,
+                                    input_shape: Optional[List[int | str]] = None):
```
wejoncy, line 29: the expression `int | str` requires python3.10, right?

skottmckay: Is there an alternative way that works with earlier versions other than `Any`? 

As this is model editing that can happen offline in any environment, requiring python 3.10 shouldn't be a blocker although it might be inconvenient.

onnxruntime_extensions/tools/pre_post_processing/docs/pdoc/index.md
skottmckay: I cleared out the docs directory and generated from scratch. This file does seem pointless, but it's easier to leave the generated output as is so there are less diffs the next time someone updates the docs.

tutorials/yolov8_pose_e2e.py
wejoncy: Could we use `Any` for python below 3.10 and `int | str` for 3.10+?
```
if sys.version_info.minor<10:
       shape_type_constrain = Any
else:
       shape_type_constrain = int | str
```
Then
```
input_shape: Optional[List[shape_type_constrain ]] = None):
```

skottmckay: I'll go with this recommendation of using `Union` to keep things simple:

https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html

```py
# On Python 3.10+, use the | operator when something could be one of a few types
x: list[int | str] = [3, 5, "test", "fun"]  # Python 3.10+
# On earlier versions, use Union
x: list[Union[int, str]] = [3, 5, "test", "fun"]
```

wejoncy (approved):

ccl-private:
Hi, @skottmckay thank you for your contribution. I have trained a custom key-points model, which causes 'mask' data to be not 56. so I modified two places. First assert classes_masks_out == custom_num; Second, post_processing_steps Split(num_outputs=3, axis=1, splits=[4, 1, custom_num]).
But when running it, it reported an error.  Here is the output log:
Updating model with jpg/png image bytes as input.
Adding pre/post processing to the model...
Updating model ...
Pre/post proceessing added.
Updated model saved.
Running the model to validate output using /slow_disk/ccl/data/goji_pose/data/2024-10-14 095626.jpg.
Traceback (most recent call last):
  File "/slow_disk/ccl/codes/yolo_custom/onnx_core_goji.py", line 373, in <module>
    _run_inference(updated_model_path, args.input, args.output_image is not None, args.test_image, rgb_layout)
  File "/slow_disk/ccl/codes/yolo_custom/onnx_core_goji.py", line 240, in _run_inference
    session = ort.InferenceSession(str(onnx_model_path), providers=providers, sess_options=session_options)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/slow_disk/ccl/codes/yolo_custom/.venv/lib/python3.11/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 419, in __init__
    self._create_inference_session(providers, provider_options, disabled_optimizers)
  File "/slow_disk/ccl/codes/yolo_custom/.venv/lib/python3.11/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 491, in _create_inference_session
    sess.initialize_session(providers, provider_options, disabled_optimizers)
onnxruntime.capi.onnxruntime_pybind11_state.Fail: [ONNXRuntimeError] : 1 : FAIL : Node (post_process_57) Op (Reshape) [ShapeInferenceError] Invalid position of 0.

skottmckay:
The splits should add up to custom_num. e.g. in the original where there are 56 elements, 4 are the box coords, 1 is the score, and 3 x 17 for the locations. 4 + 1 + 51 == 56.

So your splits should be [4, 1, (custom_num - 5)]

PR #606, opened 2023-11-21
Add compatibility with v17 API

### LINKED ISSUE #572: Custom ops fail when used with recent main line builds of ORT
Due to a change in the logic of when to call the V2 custom op API (https://github.com/microsoft/onnxruntime/commit/37dcefb5b7ed620b823f6638c178eac8b6e0ed9a#diff-eb59007a97532913933bd6d031e8f441215420a2e73657094e85c17a52976d17), onnxruntime-extensions custom ops break with the latest ORT. I think changes are required around https://github.com/microsoft/onnxruntime-extensions/blob/main/includes/onnxruntime_customop.hpp#L283C6-L293C6

samwebster:
This follows the existing pattern but I wonder if a better fix would be for ops to report their supported version as 16? The code would then work as-is with v18 etc. whereas the current pattern requires updates each time the API is extended. Fixes #572

### DIFF
--- a/includes/custom_op_lite.h
+++ b/includes/custom_op_lite.h
@@ -845,6 +845,12 @@ struct OrtLiteCustomOp : public OrtCustomOp {
     OrtCustomOp::CreateKernelV2 = nullptr;
     OrtCustomOp::KernelComputeV2 = nullptr;
 #endif
+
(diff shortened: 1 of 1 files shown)


samwebster:
Please also consider alternative fix #607

PR #607, opened 2023-11-21
Add some API forwards compatibility

samwebster:
Alternative fix for #572


samwebster:
Please also consider alternative fix https://github.com/microsoft/onnxruntime-extensions/pull/606

wenbingl (approved):
I prefer this one, I cannot think of the default value of the member in OrtCustomOp is not zero right now.

PR #609, opened 2023-11-22
Make ort extensions build successfully on Linux GPU 

mszhanyi:
How to test locally:
export CUDA_PATH=/usr/local/cuda-11.8
sh ./build.sh -DOCOS_ENABLE_CTEST=ON -DOCOS_ONNXRUNTIME_VERSION={ORT_VERSION} -DOCOS_USE_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES=86 -DONNXRUNTIME_PKG_DIR={ORT_RELEASE_DIR} 

Note:
1. Since std::filesystem is used in unit test, gcc version should be bigger than 9.0
2. CI pipeline will be enabled in next PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -814,6 +814,7 @@ if(_BUILD_SHARED_LIBRARY)
 endif()
 
 if(OCOS_USE_CUDA)
+  target_include_directories(ocos_operators PUBLIC ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES})
   target_include_directories(ortcustomops PUBLIC ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES})
   target_link_libraries(extensions_shared PUBLIC cudart cublas cufft)
 endif()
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -814,6 +814,7 @@ if(_BUILD_SHARED_LIBRARY)
 endif()
 
 if(OCOS_USE_CUDA)
+  target_include_directories(ocos_operators PUBLIC ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES})
```
wenbingl, line 817: Interesting, do you know why this missing line doesn't cause any trouble in Windows platform?

wenbingl (approved):

CMakeLists.txt
edgchen1: I'm also curious. would expect the usual cmake public header propagation to handle this given the target_link_libraries above:
https://github.com/microsoft/onnxruntime-extensions/blob/e10b4118b6f5584cfa94d1429f5b386dd91f7014/CMakeLists.txt#L775

does the ordering of these lines with the target_link_libraries call matter?

mszhanyi: I don't know the reason now. Maybe it's caused by different generators.

wenbingl: @edgchen1 @mszhanyi , I guess the reason could be CUDA windows installation already put this path into MSCV default path. but I didn't verify that.

PR #612, opened 2023-11-29
Add Fairseq decoder support

sayanshaw24:
Adds support for decoding tokenized text from Fairseq model tokenizers such as XLMRobertaTokenizer.

Validation
- [x] Tested against HF XLMRobertaTokenizer
- [x] Passes ORT Extensions tests


operators/tokenizer/sentencepiece_decoder.hpp
```diff
@@ -41,6 +42,27 @@ struct KernelSentencepieceDecoder {
     std::transform(p_ids, p_ids + ids.NumberOfElement(),
                    std::back_inserter(tids),
                    [](auto _id) { return static_cast<int>(_id); });
+    if (fairseq.has_value() && (*fairseq)) {
```
wenbingl, line 45: nit: '(' around *fairseq seems unnecessary.

wenbingl (approved):

operators/tokenizer/sentencepiece_decoder.hpp
sayanshaw24: absolutely - likely replaced something inside the () with *fairseq and forgot to remove it, will fix it in the next PR.

PR #614, opened 2023-11-30
Rashuai/bert cuda

RandySheriffH:
Add FastGelu as contrib op.


wejoncy:
Is it possible to get the GPU device-id in custom op?

RandySheriffH:
> Is it possible to get the GPU device-id in custom op?

No, custom op has no way to know the device yet.

CMakeLists.txt
```diff
@@ -64,6 +64,7 @@ option(OCOS_ENABLE_CV2 "Enable the operators in `operators/cv2`" ON)
 option(OCOS_ENABLE_VISION "Enable the operators in `operators/vision`" ON)
 option(OCOS_ENABLE_AUDIO "Enable the operators for audio processing" ON)
 option(OCOS_ENABLE_AZURE "Enable the operators for azure execution provider" OFF)
+option(OCOS_ENABLE_CONTRIB "Enable the contrib operators" ON)
```
wenbingl, line 67: This flag has dependency on CUDA, not sure how we can handle it correctly, and 'CONTRIB' is too general.
Maybe we just remove this flag, and only let USE_CUDA to enable it?

RandySheriffH: Dropped.

wenbingl (approved):

PR #615, opened 2023-12-01
add an experimental CUDA python unit test pipeline

wenbingl:


mszhanyi:
Error in WindowsPyCUDA is fixed.

wenbingl:
> Error in WindowsPyCUDA is fixed.

Can you have a look at the failure again? Since it was testing with Nightly build, it should be put in ci-optional pipeline.

And Can you enhance your existing CUDA pipeline with these Python unit tests as well? @mszhanyi

PR #622, opened 2023-12-15
Add macos platform suppport to onnxruntime-extensions-c pod

YUNQIUGUO:
As title.

- Currently it consumes a pre-release version of onnxruntime-objc pod (as the required new macos support is not yet released.)
- Validated in a macOS app with updated onnxruntime-extensions-c pod
- Pipeline changes for extensions.ios_packaging, successful build: https://aiinfra.visualstudio.com/Lotus/_build/results?buildId=390131&view=results

Motivation:

In order to allow macos support in SPM with one single Package.swift, extensions pod should include macos library support as the best solution though it has not been asked yet.

Note: validated with the dev pod archive as binary target for SPM (with platform macOS and iOS), it works as expected.


.pipelines/ios_packaging.yml
YUNQIUGUO, line 1: TODO: actually can consider updating this packaging name now.

test/ios/OrtExtensionsUsage/OrtExtensionsMacOSUsageTests/OrtExtensionsMacOSUsageTests.swift
YUNQIUGUO, line 1: not adding UITests for macos as observed some UI-tester runner crashing intermittently issue. The unit tests should cover the main testing purpose.

cmake/externals/opencv.cmake
```diff
@@ -37,7 +37,7 @@ set(BUILD_TBB                   OFF CACHE INTERNAL "")
 set(BUILD_TIFF                  OFF CACHE INTERNAL "")
 set(BUILD_WEBP                  OFF CACHE INTERNAL "")
 set(BUILD_WITH_STATIC_CRT       OFF CACHE INTERNAL "")
-if(IOS)
+if(APPLE)
```
YUNQIUGUO, line 40: enabling openCV to build zlib from source for macOS as well otherwise will result in undefined symbols issue in the output binary.

but `APPLE` covers more general apple platforms (including watchos, tvos, etc). not sure if there's a narrower definition could be used here.

tools/ios/assemble_pod_package.py
```diff
@@ -83,14 +83,19 @@ def assemble_pod_package(
     pod_description = "Pod containing pre and post processing custom ops for onnxruntime."
     with open(framework_info_file, mode="r") as f:
         framework_info = json.load(f)
-        ios_deployment_target = framework_info["apple_deployment_target"]
+        ios_deployment_target = framework_info["iphonesimulator"]["apple_deployment_target"]
+        # NOTE: hard coded the cmake osx sysroot variable for macosx key here.
```
YUNQIUGUO, line 87: as noted, not sure why sysroot is not setting as `macosx` here.

edgchen1: this will need to be updated - the path is too specific

YUNQIUGUO: yeah I agree. not sure why the cmake_osx_sysroot is not equal to macosx here. will figure out this.

tools/build.py
```diff
@@ -500,28 +500,30 @@ def _generate_build_tree(
             # if neither code signing development identity nor team id are provided, don't code sign
             cmake_args += ["-DCMAKE_XCODE_ATTRIBUTE_CODE_SIGNING_ALLOWED=NO"]
 
-        if args.ios:
+        if args.ios or args.apple_sysroot == "macosx":
```
edgchen1, line 503: I don't quite understand what this condition means now. what if `args.apple_sysroot` is unset for a MacOS build as it was before?

tools/ios/build_xcframework.py
```diff
@@ -17,9 +19,10 @@
 
 from utils import get_logger, run  # noqa
 
-_supported_platform_archs = {
+_all_supported_platform_archs = {
```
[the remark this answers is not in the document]
YUNQIUGUO: yeah.. actually initially I was thinking about keeping two sets of supported targeted platform, one is "default" (for the case when `--platform_arch` is not set.) and one is for "all supported"

but it turns out that maybe keeping one set is fine. will update the names back if necessary.

tools/build.py
YUNQIUGUO: would adding an additional build.py script argument makes more sense here? 

initially I was trying to avoid creating a new argument for `macosx` so by inferring it from the apple_sysroot value. but it does seem to be an issue for the general MacOS build when apple_sysroot is not set.

tools/ios/assemble_pod_package.py
```diff
@@ -157,7 +162,7 @@ def parse_args():
     if args.xcframework_output_dir is not None:
         args.xcframework_dir = args.xcframework_output_dir / "onnxruntime_extensions.xcframework"
         args.public_headers_dir = args.xcframework_output_dir / "Headers"
-        args.framework_info_file = args.xcframework_output_dir / "framework_info.json"
+        args.framework_info_file = args.xcframework_output_dir / "xcframework_info.json"
```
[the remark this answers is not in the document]
YUNQIUGUO: will do.

tools/build.py
YUNQIUGUO: sorry actually maybe I could just use `arg.build_apple_framework` for these conditional apple cmake definitions.

edgchen1: there's some iOS specific logic here too. maybe `args.build_apple_framework or args.ios`, similar to ORT's build.py?

https://github.com/microsoft/onnxruntime/blob/50cbcf95877b60795f32c4538611f9a119bb0291/tools/ci_build/build.py#L1293

tools/build.py
```diff
@@ -644,7 +646,7 @@ def adb_shell(*args, **kwargs):
 
 
 def _run_ios_tests(args, config: str, config_build_dir: Path):
-    if platform.machine() == "x86_64" and args.apple_arch != "x86_64":
+    if platform.machine() == "x86_64" and args.apple_arch != "x86_64" and args.apple_sysroot != "macosx":
```
edgchen1, line 646: this function only gets called when `args.ios` is true - does it make sense for `args.apple_sysroot` to ever be "macosx" here?

YUNQIUGUO: thanks for the catch up. fixed.

tools/ios/assemble_pod_package.py
```diff
@@ -81,16 +81,21 @@ def assemble_pod_package(
...
     pod_description = "Pod containing pre and post processing custom ops for onnxruntime."
-    with open(framework_info_file, mode="r") as f:
-        framework_info = json.load(f)
-        ios_deployment_target = framework_info["apple_deployment_target"]
+    with open(xcframework_info_file, mode="r") as f:
+        xcframework_info = json.load(f)
+        ios_deployment_target = xcframework_info["iphonesimulator"]["apple_deployment_target"]
+        macos_deployment_target = ""
+        for key in xcframework_info.keys():
+            if "MacOSX" in key:
```
edgchen1, line 89: hm, so why is it not "macosx"? seems odd that we can look up "iphonesimulator" fine.

is `CMAKE_OSX_SYSROOT` getting defined to "macosx"?

YUNQIUGUO: from the building log, it looks like it's setting its value from apple.sysroot:
2023-12-16T01:17:53.6341750Z   /Users/runner/hostedtoolcache/Python/3.9.18/x64/bin/cmake /Users/runner/work/1/s -DPython_EXECUTABLE=/Users/runner/hostedtoolcache/Python/3.9.18/x64/bin/python -DOCOS_ENABLE_SELECTED_OPLIST=OFF -DCMAKE_OSX_ARCHITECTURES=x86_64 -DCMAKE_OSX_DEPLOYMENT_TARGET=11.0 -DCMAKE_XCODE_ATTRIBUTE_CODE_SIGNING_ALLOWED=NO -DOCOS_BUILD_APPLE_FRAMEWORK=ON **-DCMAKE_OSX_SYSROOT=macosx** -DOCOS_ENABLE_SELECTED_OPLIST=ON -G 'Unix Makefiles' -DCMAKE_BUILD_TYPE=Release

not sure why in the resulting framework_info.json the cmake_OSx_ROOT is not equal to "macosx"..

tools/ios/assemble_pod_package.py
```diff
@@ -81,16 +81,22 @@ def assemble_pod_package(
     pod_name = "onnxruntime-extensions-c"
     pod_summary = "ONNX Runtime Extensions C/C++ Pod"
     pod_description = "Pod containing pre and post processing custom ops for onnxruntime."
-    with open(framework_info_file, mode="r") as f:
-        framework_info = json.load(f)
-        ios_deployment_target = framework_info["apple_deployment_target"]
+    with open(xcframework_info_file, mode="r") as f:
+        xcframework_info = json.load(f)
+        ios_deployment_target = xcframework_info["iphonesimulator"]["apple_deployment_target"]
```
skottmckay, line 86: If I was to do a build with just 'iphoneos' would this work? 

Could we be more generic/consistent and iterate the keys with a case for each instead of hardcoding one key here and processing the macos value separately?

e.g. maybe one case for key.startswith('iphone') to cover 'iphoneos' and 'iphonesimulator' (or a comment explaining why we only check iphonesimulator and not iphoneos) and one case with 'MacOSX' in key?

tools/ios/build_xcframework.py
```diff
@@ -50,38 +65,49 @@ def build_framework_for_platform_and_arch(
...
+                str(_repo_dir / "tools" / "build.py"),
+                f"--build_dir={build_dir}",
+                f"--config={config}",
+                "--update",
+                "--build",
+                "--parallel",
+                "--test",
+                "--build_apple_framework",
+                f"--apple_sysroot={platform}",
+                f"--apple_arch={arch}",
```
skottmckay, line 83: nit: indentation looks off. would have expected it to align with the start of `sys.executable` given these are all items in the same array.

tools/ios/assemble_pod_package.py
YUNQIUGUO: yeah good point technically speaking, it's possible to only specify `--platform iphoneos arm64`for the build. and this won't work. though currently we have a default set where all archs are included for the case platform argument is not specified.

updated as suggested.

YUNQIUGUO: yeah good point. it's possible to do a build with only specifying `--platform iphoneos` and this probably won't work.

updated as suggested.

tools/ios/assemble_pod_package.py
```diff
@@ -81,16 +81,25 @@ def assemble_pod_package(
     pod_name = "onnxruntime-extensions-c"
     pod_summary = "ONNX Runtime Extensions C/C++ Pod"
     pod_description = "Pod containing pre and post processing custom ops for onnxruntime."
-    with open(framework_info_file, mode="r") as f:
-        framework_info = json.load(f)
-        ios_deployment_target = framework_info["apple_deployment_target"]
+    with open(xcframework_info_file, mode="r") as f:
+        xcframework_info = json.load(f)
+        ios_deployment_target = ""
```
skottmckay, line 86:
```suggestion
        print(f"Setting deployment target from available frameworks: {', '.join(xcframework_info.keys())}")
        ios_deployment_target = ""
```

To aid debugging if there's an issue with an empty deployment target value later on it might help to output which frameworks were available here.

skottmckay (approved):

PR #629, opened 2024-01-04
Fix extensions nuget pipeline

YUNQIUGUO:
- successful build: https://aiinfra.visualstudio.com/Lotus/_build/results?buildId=394396&view=results


wenbingl (dismissed):

.pipelines/templates/build-package-for-ios-cocoapods.yml
```diff
@@ -122,6 +125,7 @@ jobs:
 
   - script: |
       ORT_EXTENSIONS_LOCAL_POD_PATH=$(Build.BinariesDirectory)/pod_staging \
+      EXCLUDE_MACOS_TARGET=true \
```
skottmckay, line 128: out of interest if we don't build the macos frameworks, why do we need to explicitly exclude them? 

that seems like it will be a painful experience if someone is trying to do a local build with a subset of platforms

YUNQIUGUO: in our current test app setup, the Podfile by default will include both macos/ios testing targets so we can just share one single testing app.

And specifying this variable is just to let the CI test project pass if MacOS arch not included in the current build.

If a user doing a local build with their own project with correct specifying supported platform Podfiles, wouldn't need to explicitly setting this variable.

YUNQIUGUO: more natural way would be I can modify to change it to include the macos target if we build for macos:)

skottmckay: Can we add a comment to the script for `pod install` with this info? 

something along the lines of 'we exclude macos in the build_framework.py command in this CI but the test app's podfile is setup for all platforms, and due to that we have to explicitly exclude the macos target when installing the pod for the test app'. not sure that's completely accurate so please adjust as needed.

YUNQIUGUO: yes, I added a note.  also added for another place in CI.yml file for the case where we only build for a subset of platforms.

skottmckay (approved):

PR #632, opened 2024-01-09
add a gen_processing_model option to cast token-id for int64

wenbingl:
The sentencepiece based tokenizer token-id is int32. To be consistent, add an option to cast the output token-id.


test/test_autotokenizer.py
```diff
@@ -28,6 +28,20 @@ def test_llama_tokenizer(self):
...
+        tokenizer = AutoTokenizer.from_pretrained("hf-internal-testing/llama-tokenizer")
+        text = "I was born in 92000, and this is falsé."
+        ids = tokenizer.encode(text, return_tensors="np")
+
+        ort_tok = OrtPyFunction.from_model(gen_processing_models(
+            tokenizer,
+            pre_kwargs={"WITH_DEFAULT_INPUTS": True,
+                        "CAST_TOKEN_ID": True})[0])
+        actual_ids = ort_tok([text])[0]
+        self.assertEqual(actual_ids.dtype, np.int64)
```
sayanshaw24, line 42: what was the type before casting? are we adding the `CAST_TOKEN_ID` option just for this Llama use case?

i suggest changing the `CAST_TOKEN_ID` option to a string type rather than boolean, and changing the implementation to allow casting to _all_ supported TensorProto types. Then we can feed the string input (ex: "int64") to a dict similar to the one we have for tokenizers in `_hf_cvt.py` and cast to the respective type. That way this option can be used to cast for all supported types.

Otherwise, we may rename the option to something like `CAST_TO_INT64` for now.

onnxruntime_extensions/util.py
```diff
@@ -112,11 +112,13 @@ def remove_unused_initializers(subgraph, top_level_initializers=None):
...
     # Filter the initializers by checking if their names are in the list of used input tensors
-    used_initializers = [init for init in all_initializers if init.name in input_tensors]
+    used_initializers = [
+        init for init in all_initializers if init.name in input_tensors]
 
     # Update the subgraph's initializers
     del subgraph.initializer[:]
-    subgraph.initializer.extend([init for init in used_initializers if init in initializers])
+    subgraph.initializer.extend(
+        [init for init in used_initializers if init in initializers])
```
sayanshaw24, line 121: for all the indentation/line changes, are you using a specific standard (ex: 80 chars max per line)? i can also make sure to do this in the future so we can minimize such cleanup in the future.

wenbingl: it's not about the maximum line length, I think some AI tools are trying to beautify the code and make it be more readable.

test/test_autotokenizer.py
wenbingl: it's not LLAMA only, right now, some tokenizer generated int64 token-id, other did it with int32. But make the data type changes on the existing tokenizer output may cause some unexpected compatible issue. So this option was added is to give user a chance to have a consistent experience.

The full name should be CAST_TOKEN_ID_TO_INT64, to be short, CAST_TOKEN_ID was chose to let the user know only token-id was casted, not any outputs.

onnxruntime_extensions/util.py
sayanshaw24: sounds good!

PR #640, opened 2024-01-26
Fix batching in fairseq SentencepieceTokenizer

### LINKED ISSUE #639: Fairseq support in SentencepieceTokenizer doesn't work with batches
I've exported some HF Sentencepiece tokenizers that use the fairseq support using the SentencepieceTokenizer op (setting the fairseq flag as an input to the tokenizer op). This works when there is a single element that I'm tokenizing, but fails for batches (I have logic elsewhere in my ONNX export which appropriately batches the sentencepiece output and adds the right padding). The issue is that the fairseq transformation [here](https://github.com/microsoft/onnxruntime-extensions/blob/main/operators/tokenizer/sentencepiece_tokenizer.cc#L98) runs from `content.begin()` to `content.end()`, but it's inside the loop over input strings, so it is applied multiple times. I think the fix is just to move the fairseq correction outside the for loop (guarded by add_rev as well), and I can send a PR for that if you want.

For example, on this batch of three sentences the first sentence has the BOS 4 higher, and the rest of the tokens 2 higher, the second sentence has it's BOS 3 higher and the rest of the tokens 1 higher, while the last sentence is correct. The example uses [this tokenizer](https://huggingface.co/pkshatech/GLuCoSE-base-ja), though the issue is present in all things which hit the fairseq correction.

```
HF:
[0, 1514, 5713, 5655, 716, 4783, 13, 2148, 9407, 7, 2],
[0, 24340, 4197, 29833, 427, 6125, 29841, 24135, 29829, 77, 30084, 752, 3000, 2],
[0, 29834, 30009, 3579, 3375, 29832, 5713, 5655, 716, 29843, 2148, 1567, 742, 3122, 7, 2]

ONNX:
[4, 1516, 5715, 5657, 718, 4785, 15, 2150, 9409, 9, 2],
[3, 24341, 4198, 29834, 428, 6126, 29842, 24136, 29830, 78, 30085, 753, 3001, 2],
[0, 29834, 30009, 3579, 3375, 29832, 5713, 5655, 716, 29843, 2148, 1567, 742, 3122, 7, 2]
```

Craigacp:
This fixes #639 by moving the fairseq id patching out of the loop so it's not applied multiple times to sequences earlier in the batch.

I've added the check to ensure it's not applied when `add_rev` is true mirroring its position in the for loop, but I'm unsure if that's required (presumably there are no HF tokenizers which emit a reversed string and have this fairseq hack?).

I tested it against the example in #639 and both batches and single sequences work correctly now.

### DIFF
--- a/operators/tokenizer/sentencepiece_tokenizer.cc
+++ b/operators/tokenizer/sentencepiece_tokenizer.cc
@@ -85,6 +85,9 @@ OrtStatusPtr KernelSentencepieceTokenizer::Compute(const ortc::Tensor<std::strin
       }
     }
   }
+  instance_indices.push_back(content.size());
+
+  // Patch fairseq indices
   if (fairseq.has_value() && (*fairseq) && !add_rev) {
     // HF Fairseq Example (XLMRobertaTokenizer) : https://huggingface.co/transformers/v4.6.0/_modules/transformers/models/xlm_roberta/tokenization_xlm_roberta.html#XLMRobertaTokenizer
     //
@@ -106,7 +109,6 @@ OrtStatusPtr KernelSentencepieceTokenizer::Compute(const ortc::Tensor<std::strin
       }
     });
   }
-  instance_indices.push_back(content.size());
 
   // Setup output
   std::vector<int64_t> size_content(1);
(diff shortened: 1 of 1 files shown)


Craigacp:
I ran the Python tests locally on macOS ARM64 and everything passed. The access violation in the Windows tests happens either in `test_cliptok` or `test_cv2` both of which passed fine on my machine and don't use the code I changed, and they also passed fine in Linux and macOS in the CI. Is there anything more I can do to run the failure down?

wenbingl:
@sayanshaw24 , can you review this PR?

sayanshaw24 (approved):
good catch! missed the `!add_rev` check in the if-statement - this looks good, thanks for this.

PR #653, opened 2024-02-14
make onnx package to be optional.

wenbingl:
ONNX dependency is only used on offline model generation, so make it be optional to remove the dependency on building ort-extensions package.


onnxruntime_extensions/_ortapi2.py
```diff
@@ -37,10 +114,10 @@ def get_opset_version_from_ort():
         "1.12": 17,
         "1.13": 17,
         "1.14": 18,
-        "1.15": 18
+        "1.15": 18,
```
sayanshaw24, line 117: nit: why the comma here?

sayanshaw24 (approved):

onnxruntime_extensions/_ortapi2.py
wenbingl: VS code formatter did it. no sure why it is

sayanshaw24 (approved):

PR #663, opened 2024-02-26
Add args to original Yolo script to be able to run it standalone

natke:


tutorials/yolo_e2e.py
```diff
@@ -57,7 +60,19 @@ def run_inference(onnx_model_file: Path):
 if __name__ == '__main__':
     # YOLO version. Tested with 5 and 8.
     version = 8
-    onnx_model_name = Path(f"../test/data/yolov{version}n.onnx")
+
+    parser = argparse.ArgumentParser(
+        """Add pre and post processing to the YOLOv8 model.
+        """)
```
skottmckay, line 66:
```suggestion
    parser = argparse.ArgumentParser("Add pre and post processing to the YOLOv8 model.")
```

tutorials/yolo_e2e.py
```diff
@@ -57,7 +60,19 @@ def run_inference(onnx_model_file: Path):
 if __name__ == '__main__':
     # YOLO version. Tested with 5 and 8.
     version = 8
```
skottmckay, line 62: Do we need this anymore?

natke: The script would allow them to switch back to 5?

skottmckay: more that we don't have an arg to set the version so separating it from the default model name doesn't seem to add much value. don't care too much about this though.

skottmckay (approved):

PR #665, opened 2024-03-01
Update ort version to 1.17.1

sayanshaw24:


.pipelines/ci.yml
```diff
@@ -47,10 +51,6 @@ stages:
           python.version: '3.9'
           torch.version: 'torch==1.13.1+cpu torchvision==0.14.1+cpu torchaudio==0.13.1 --index-url https://download.pytorch.org/whl/cpu'
           ort.version: '1.14.1'
-        py38-1131:
```
wenbingl, line 50: can we still keep Python version 3.8?

sayanshaw24: yeah sounds good was planning on doing the same, had originally updated them to add python 3.12 like you did in your recent PR - just reverted most of the python versions for the moment though as it seems 3.12 might also be causing some CI problems with ort 1.17.1 on WindowsPython and MacOSPython (which we can address in a different PR).

.pipelines/ci.yml
```diff
@@ -35,22 +35,22 @@ stages:
 
     strategy:
       matrix:
-        py311-1163:
+        py312-1171:
           python.version: '3.11'
```
wenbingl, line 39: the version doesn't match the job name. what's the issue of 3.12 here?

sayanshaw24: [Pipelines - Run 20240229.1 logs (azure.com)](https://dev.azure.com/onnxruntime/onnxruntime/_build/results?buildId=1309882&view=logs&j=498307ac-daeb-53b9-75d4-6986ffc4c71d&t=8eef9f10-c422-50ef-60ef-1c3882233af9): error building onnx wheel because onnx doesn't have python 3.12 package; skipping requirements install and unit test for WindowsPython in latest commit.

.pipelines/ci.yml
```diff
@@ -417,25 +429,14 @@ stages:
...
-      displayName: Install requirements{-dev}.txt and cmake python modules
+      displayName: Install requirements and cmake python modules
 
     - script: |
         call activate pyenv
         set CMAKE_ARGS=-DOCOS_ONNXRUNTIME_VERSION=$(ort.version)
         python -m pip install .
       displayName: Build the wheel
 
-    - script: |
```
wenbingl, line 429: will all python version skip the unit test, 
you can try to   "condition: neq(variables['python.version'], '3.12')" to only skip if python version is 3.12

sayanshaw24: ah of course, sorry - fixed.

wenbingl (approved):

Issue #561, opened 2023-09-14
SentencePiece Tokenizer has difference result if input text contains '<unk>'


wenbingl:

wenbingl:
fixed

PR #666, opened 2024-03-01
Unify the spm/bpe tokenizers

wenbingl:
#561 fixed

### DIFF
--- a/onnxruntime_extensions/_cuops.py
+++ b/onnxruntime_extensions/_cuops.py
@@ -51,13 +51,15 @@ class GPT2Tokenizer(CustomOp):
 
     @classmethod
     def get_inputs(cls):
-        return [cls.io_def("input_text", onnx_proto.TensorProto.STRING, [None])]
+        return [
+            cls.io_def('input_text', onnx_proto.TensorProto.STRING, [None])
+        ]
 
     @classmethod
     def get_outputs(cls):
         return [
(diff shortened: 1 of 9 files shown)


sayanshaw24 (approved):
All looks good. My only comment is that `KernelBpeTokenizer::SpmTokenize` does not support fairseq or computing instance indices at the moment.

But I noticed you added the SPM implementation to bpe_kernels.cc but did not delete sentencepiece_tokenizer.cc, sentencepiece_tokenizer.hpp, or sentencepiece_decoder.hpp. I assume we will do this in the future as that is the main incentive of this unification, and when we do, we can also take care of the remaining SPM stuff (fairseq and instance indices) in the bpe source.

wenbingl:
> All looks good. My only comment is that `KernelBpeTokenizer::SpmTokenize` does not support fairseq or computing instance indices at the moment.
> 
> But I noticed you added the SPM implementation to bpe_kernels.cc but did not delete sentencepiece_tokenizer.cc, sentencepiece_tokenizer.hpp, or sentencepiece_decoder.hpp. I assume we will do this in the future as that is the main incentive of this unification, and when we do, we can also take care of the remaining SPM stuff (fairseq and instance indices) in the bpe source.

There will be a transition period. The old sentencepiece tokenizer code needs to be kept for Custom-Op backward compatibility. The faireseq indices will be added in the future.

PR #676, opened 2024-03-22
support GroupQueryAttention

jslhcl:
support GroupQueryAttention

### DIFF
--- a/test/cuda/test_cudaops.py
+++ b/test/cuda/test_cudaops.py
@@ -1,12 +1,602 @@
 import unittest
 import numpy as np
 from numpy.testing import assert_almost_equal
-from onnx import helper, onnx_pb as onnx_proto
+from onnx import helper, onnx_pb as onnx_proto, TensorProto
 from onnxruntime_extensions import make_onnx_model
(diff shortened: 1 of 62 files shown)


test/cuda/test_cudaops.py
```diff
@@ -1,12 +1,602 @@
...
+        np.allclose(
+            out,
+            out_ref,
+            rtol=rtol,
+            atol=atol,
+            equal_nan=True,
+        ),
+    )
 
 class TestCudaOps(unittest.TestCase):
```
sayanshaw24, line 601: In the future (or potentially even in this PR), we can separate our Python tests for CUDA by different ops (like we do for cpu tests). Otherwise, in the long term this file will get too long (GitHub is already not showing its preview by default on this PR because of file size).

nit: regardless, we should probably rename this file as it is already under `test/cuda`, so `test_cudaops.py` is redundant.

sayanshaw24 (commented):
Could you add a C++ unit test for GroupQueryAttention as well under `test/shared_test`?

PR #677, opened 2024-03-24
Improve handling of missing `vocab_file` attribute in HFTokenizerConverter

kazssym:
This commit updates `HFTokenizerConverter` to handle cases where the `hf_tokenizer` object might not have a `vocab_file` attribute.

Changes:

* Uses `getattr` to retrieve the `vocab_file` attribute for flexibility
* Stores the retrieved value in a separate variable `vocab_file` for clarity
* Checks if `vocab_file` is `None` before checking its existence

This ensures the converter works correctly even with tokenizers that don't define a `vocab_file` attribute.


kazssym:
I'm not sure but it looks like `GPT2Tokenizer/-Fast` lacks the `vocab_file` attribute.

wenbingl (approved):
Thanks for the fixing!

PR #682, opened 2024-03-27
Ensure ONNX model generation considers opset version

kazssym:
This commit modifies the `gen_processing_models` function in `onnxruntime_extensions/cvt.py` to explicitly include the `opset_version` argument when calling `make_onnx_model` for both pre-processing and post-processing graphs.

Previously, the function only generated models if `pre_g` or `post_g` existed, without specifying the opset version. This update ensures that generated ONNX models adhere to the desired opset level, improving compatibility and deployment options.

### DIFF
--- a/onnxruntime_extensions/cvt.py
+++ b/onnxruntime_extensions/cvt.py
@@ -90,7 +90,7 @@ def gen_processing_models(processor: Union[str, object],
             **pre_kwargs) if pre_kwargs is not None else None
         post_g = _converter.post_processing(
             **post_kwargs) if post_kwargs is not None else None
-        return make_onnx_model(pre_g) if pre_g else None, \
-            make_onnx_model(post_g) if post_g else None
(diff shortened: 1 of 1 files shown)


wenbingl (approved):

onnxruntime_extensions/cvt.py
```diff
@@ -90,7 +90,7 @@ def gen_processing_models(processor: Union[str, object],
             **pre_kwargs) if pre_kwargs is not None else None
         post_g = _converter.post_processing(
             **post_kwargs) if post_kwargs is not None else None
-        return make_onnx_model(pre_g) if pre_g else None, \
-            make_onnx_model(post_g) if post_g else None
+        return make_onnx_model(pre_g, opset_version=opset) if pre_g else None, \
+            make_onnx_model(post_g, opset_version=opset) if post_g else None
```
wenbingl, line 94: it looks it also need some fixing for the case of opset==None, check CI pipeline result for more details. Would you like to do that, or wait for my help?

PR #687, opened 2024-04-07
Add support for YOLO to output bounding boxes

louisfacun:
Implemented option for users to choose between two output formats: byte conversion of the image with drawn bounding boxes or a list of bounding boxes for further processing.

Also, based on this: https://onnxruntime.ai/docs/tutorials/mobile/pose-detection.html, if we didn't modify `yolo_detection()`, how can it can return two values (rawImage and boxOutput)? It looks like the used `yolo_detection()` from `yolo_e2e.py` is ahead from what I used, any ideas?

```
val rawOutput = (output?.get(0)?.value) as ByteArray
val boxOutput = (output?.get(1)?.value) as Array<FloatArray> // Index out of bounds
```

### DIFF
--- a/onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
+++ b/onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
@@ -163,7 +163,8 @@ def superresolution(model_file: Path, output_file: Path, output_format: str, onn
 
 
 def yolo_detection(model_file: Path, output_file: Path, output_format: str = 'jpg',
-                   onnx_opset: int = 16, num_classes: int = 80, input_shape: List[int] = None):
+                   onnx_opset: int = 16, num_classes: int = 80, input_shape: List[int] = None,
+                   draw_bboxes: bool = True):
     """
     SSD-like model and Faster-RCNN-like model are including NMS inside already, You can find it from onnx model zoo.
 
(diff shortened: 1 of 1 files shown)


skottmckay:
> Also, based on this: https://onnxruntime.ai/docs/tutorials/mobile/pose-detection.html, if we didn't modify `yolo_detection()`, how can it can return two values (rawImage and boxOutput)? It looks like the used `yolo_detection()` from `yolo_e2e.py` is ahead from what I used, any ideas?

I'm not quite sure what you're asking here and whether it's about the pose model or the object detection model. Each has different output. And can you clarify what 'rawImage' equates to?

louisfacun:
> > Also, based on this: https://onnxruntime.ai/docs/tutorials/mobile/pose-detection.html, if we didn't modify `yolo_detection()`, how can it can return two values (rawImage and boxOutput)? It looks like the used `yolo_detection()` from `yolo_e2e.py` is ahead from what I used, any ideas?
> 
> I'm not quite sure what you're asking here and whether it's about the pose model or the object detection model. Each has different output. And can you clarify what 'rawImage' equates to?

Sorry, it's for object detection model, and the `rawImage` I am referring is the detected image with drawn bounding boxes.

skottmckay:
You can add an Identity step at the end to produce an additional model output from the ScaleBoundingBoxes step.

e.g. after the last post-processing step, add an Identity step like below

```py
        # Encode to jpg/png
        ConvertBGRToImage(image_format=output_format),
        
        # also return bounding boxes
        (Identity(name="bounding_boxes"),
         [utils.IoMapEntry("ScaleBoundingBoxes", producer_idx=0, consumer_idx=0)]),
```

onnxruntime_extensions/tools/add_pre_post_processing_to_model.py
```diff
@@ -185,6 +186,7 @@ def yolo_detection(model_file: Path, output_file: Path, output_format: str = 'jp
     :param onnx_opset: The opset version of onnx model, default(16).
     :param num_classes: The number of classes, default(80).
     :param input_shape: The shape of input image (height,width), default will be asked from model input.
+    :param draw_bboxes: Whether to draw bounding boxes on the image, default(True).
```
skottmckay, line 189: Would be good for this to explain how this affects the model output (returning scaled bounding boxes vs. a jpg/png image). Should probably also call out that if this is False the `output_format` value becomes unused.

Issue #679, opened 2024-03-25
Add compatibility in java ext for arm64


github-gauthier-perrod:
We don't check linux system for arm64 architecture in java-ext:

https://github.com/microsoft/onnxruntime-extensions/blob/main/cmake/ext_java.cmake#L98-L99

It could be very easy to correc this by adding those two lines just below:

`elseif (CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64")
   set(JNI_ARCH aarch64)`
   
  I've tested in on linux arch64 architecture machine and it worked well

wenbingl:
thanks for the suggestion, I have created a PR for that.

PR #688, opened 2024-04-08
Update java package script for linux arm64

wenbingl:
#679


sayanshaw24 (approved):
thanks for adding this!

PR #695, opened 2024-04-22
Test triton cmake with different deps location to avoid gtest clash

sayanshaw24:


cmake/externals/triton.cmake
```diff
@@ -1,10 +1,10 @@
 include(ExternalProject)
 
-set(triton_PREFIX ${CMAKE_CURRENT_BINARY_DIR}/_deps/triton)
+set(triton_PREFIX ${CMAKE_CURRENT_BINARY_DIR}/_deps_triton/triton)
```
wenbingl, line 3: this change may not work, I guess the include path will be changed to this new _deps_triton too.

PR #708, opened 2024-05-03
Introduce flash attention and cutlass library

jslhcl:
Introduce flash attention and cutlass library


CMakeLists.txt
```diff
@@ -3,6 +3,7 @@
 
 # Minimum CMake required
 cmake_minimum_required(VERSION 3.25)
+cmake_policy(SET CMP0104 OLD)
```
wenbingl, line 6: should we set the policy to be new?

cmake/ext_cuda.cmake
```diff
@@ -6,6 +6,14 @@ enable_language(CUDA)
 
 set(CMAKE_CUDA_RUNTIME_LIBRARY Shared)
 set(CMAKE_CUDA_STANDARD 17)
+include(CMakeDependentOption)
```
wenbingl, line 9: should these options be moved into the main CMakeLists.txt?

cmake/ext_cuda.cmake
```diff
@@ -6,6 +6,14 @@ enable_language(CUDA)
...
+cmake_dependent_option(USE_FLASH_ATTENTION "Build flash attention kernel for scaled dot product attention" ON "NOT WIN32" OFF)
+option(USE_MEMORY_EFFICIENT_ATTENTION "Build memory efficient attention kernel for scaled dot product attention" ON)
```
wenbingl, line 11: Usually there is a prefix before these cmake options to avoid the potential conflictions.

cmake/externals/cutlass.cmake
```diff
@@ -0,0 +1,11 @@
+include(FetchContent)
```
wenbingl, line 1: already used in CMakeLists.txt

operators/cuda/attention_lib/cutlass_fmha/fmha_launch_template.h
```diff
@@ -0,0 +1,276 @@
...
+#if defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-parameter"
+#pragma GCC diagnostic ignored "-Wstrict-aliasing"
+#endif
+
+#include "memory_efficient_attention.h"
+#include "41_fused_multi_head_attention/kernel_forward.h"
+
+namespace contrib {
```
wenbingl, line 15: a more meaningful namespace, of the default one: ort_extensions

wenbingl (commented):
Can we have some basic testing code for them?

CMakeLists.txt
jslhcl: "SET CMP0104 NEW" will initialize CMAKE_CUDA_ARCHITECTURES inside cmake. Since now both cmdclass.py and build.sh will set it explicitly, I am fine to set the policy to be new. 

Just remind that build.bat does not set CMAKE_CUDA_ARCHITECTURES, so build ORT-extensions in Windows with cuda enabled will have some strange value for this variable. 

Tested just now, running "nvidia-smi --query-gpu=compute_cap --format=csv,noheader,nounits" in my win machine has the result 7.5, but "SET CMP0104 NEW" will make it 52. Are you OK with that?

CMakeLists.txt
```diff
@@ -581,6 +583,10 @@ target_include_directories(ocos_operators PUBLIC
   ${PROJECT_SOURCE_DIR}/base
   ${PROJECT_SOURCE_DIR}/operators)
 
+if (OCOS_USE_CUDA) 
+  target_include_directories(ocos_operators PUBLIC ${cutlass_SOURCE_DIR}/include ${cutlass_SOURCE_DIR}/examples)
```
wenbingl, line 587: interesting, why does the examples are needs here?

wenbingl (approved):

CMakeLists.txt
jslhcl: some headers are in the example folder, same as ORT. Not sure why

PR #709, opened 2024-05-06
Add kernel NegXPlus1 = 1 - X

xadupre:


operators/cuda/negxplus1.h
```diff
@@ -0,0 +1,33 @@
...
+
+namespace contrib {
+
+template <typename T>
+struct NegXPlus1 {
+  template <typename TDict>
+  OrtStatusPtr OnModelAttach(const TDict& /*dict*/) {
+    return nullptr;
+  }
+  OrtStatusPtr Compute(Ort::Custom::CUDAKernelContext* ctx,
```
wenbingl, line 16: If you want to support Eager mode debugging, you need to return OrtxStatus, instead of OrtStatusPtr here. But we can replace all of them later.

wenbingl (approved):

operators/cuda/negxplus1.h
xadupre: I'll do that in a different PR.

PR #711, opened 2024-05-07
Return added_tokens_ by reference

baijumeswani:
https://github.com/microsoft/onnxruntime-extensions/blob/c58c930739ffa481744d59a3e170895e1461b9b4/operators/tokenizer/bpe_streaming.hpp#L34 is expecting a reference, but the function does not return a reference resulting in a compiler error.

This PR addresses this issue.

### DIFF
--- a/operators/tokenizer/bpe_kernels.h
+++ b/operators/tokenizer/bpe_kernels.h
@@ -115,7 +115,7 @@ class JsonFastTokenizer : KernelBpeTokenizer {
                      std::optional<ortc::Tensor<int64_t>*> offset_mapping) const;
 
  public:
-  auto& GetAddedTokens() const { return added_tokens_; }
+  const auto& GetAddedTokens() const { return added_tokens_; }
   const ort_extensions::BpeModel& GetEncoder() const { return *bbpe_tokenizer_; }
 
  private:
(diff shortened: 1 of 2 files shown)


wenbingl (approved):

operators/tokenizer/bpe_kernels.h
```diff
@@ -115,7 +115,7 @@ class JsonFastTokenizer : KernelBpeTokenizer {
                      std::optional<ortc::Tensor<int64_t>*> offset_mapping) const;
 
  public:
-  auto GetAddedTokens() const { return added_tokens_; }
+  auto& GetAddedTokens() const { return added_tokens_; }
```
wenbingl, line 118: Is it possible to share more detail what kind of C++ compiler or flag are triggering error? We can add it in ort-extensions to avoid this kind of issue in the future.

baijumeswani: I am using linux GCC compiler 12.x for this. I think this would be an error regardless of the compiler flags being used. Not sure what the best way it to catch this error.

wenbingl: Thanks, it looks like only GCC 12 catch this error, and the pipeline are only using

 -- The CXX compiler identification is GNU 9.4.0

edgchen1, line 118: nit: would suggest being explicit with the const as this is a const function
```suggestion
  const auto& GetAddedTokens() const { return added_tokens_; }
```

edgchen1 (approved):

PR #713, opened 2024-05-10
Fix warnings around usage of virtual.

skottmckay:
Fix warnings like this:

```
D:/Android/ndk/25.1.8937393/toolchains/llvm/prebuilt/windows-x86_64/sysroot/usr/include/c++/v1/memory:2427:5: warning: delete called on 'Ort::Custom::IStringTensorStorage<std::basic_string_view<char>>' that is abstract but has non-virtual destructor [-Wdelete-abstract-non-virtual-dtor]
    delete __ptr;
    ^
...
D:/Android/ndk/25.1.8937393/toolchains/llvm/prebuilt/windows-x86_64/sysroot/usr/include/c++/v1/memory:2427:5: warning: delete called on 'Ort::Custom::ITensorStorage' that is abstract but has non-virtual destructor [-Wdelete-abstract-non-virtual-dtor]
    delete __ptr;
```

Cleanup virtual/override usage. 
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-override
https://github.com/isocpp/CppCoreGuidelines/pull/1448

### DIFF
--- a/include/custom_op/tensor_api.h
+++ b/include/custom_op/tensor_api.h
@@ -64,6 +64,7 @@ public:
   virtual const void* DataRaw() const = 0;
   virtual bool IsInitialized() const = 0;
   virtual void* Initialize(const std::vector<int64_t>& shape, size_t element_size) = 0;
+  virtual ~ITensorStorage() = default;
 };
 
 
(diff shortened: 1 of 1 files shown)


wenbingl (approved):
Thanks for fixing.
Is it possible that we can setup a pipeline for detecting this kind of style issue?

skottmckay:
Possibly it's just a case of setting 'warnings as errors' to detect.

PR #716, opened 2024-05-14
Update OpenCV build to just include libspng and libjpeg libraries

sayanshaw24:
updates OpenCV build to just include the two necessary libraries and remove unnecessary peripherals.


sayanshaw24:
Handled in #800.

PR #717, opened 2024-05-15
add move constructor and Release API for tensor

souptc:
1. Add move constructor, 
2. Disable copy constructor
3. Add Release API so in pytorch integration we can transfer the tensor ownership to torch tensor.


wenbingl (approved):
I also want to discuss with you about this feature requirement, and it happens

PR #719, opened 2024-05-16
Introduce ONNXRUNTIME_PKG_DIR parameter to pip install

jslhcl:
Introduce ONNXRUNTIME_PKG_DIR parameter to pip install. This change is helpful when pip install against the latest ORT package.
After this change, pip install can support the format like:

pip install . --config-settings "ortx-user-option=use-cuda,cc_debug,ort_pkg_dir=/home/leca/ort_pkg_19"


wenbingl (commented):
Thanks!
Can you also update the doc here too: https://github.com/microsoft/onnxruntime-extensions/blob/main/docs/development.md#python-package?

wenbingl (approved):

PR #721, opened 2024-05-17
Remove C++ filesystem library dependency for the compatibility of old system

wenbingl:
… OS.


sayanshaw24 (approved):

base/file_sys.h
```diff
@@ -0,0 +1,83 @@
...
+
+#include <string>
+#include <fstream>
+
+namespace ort_extensions {
+
+class path {
+ public:
+  path() = default;
+  path(const std::string& path) : path_(path){};
```
snnn, line 20: On Windows it should be wstring. Otherwise you could hit path length size limit

wenbingl: Yes, it was done differently. the input string here is UTF-8 and on Windows it will be converted to UTF-16 before any file access API.

PR #725, opened 2024-05-23
new APIs for ORT-genai

jslhcl:
Add new APIs for ORT-genai project

### DIFF
--- a/include/custom_op/kernel_context.h
+++ b/include/custom_op/kernel_context.h
@@ -5,6 +5,7 @@
 #include <optional>
 #include <numeric>
 #include <type_traits>
+#include "onnxruntime_c_api.h"
 
 namespace Ort {
 namespace Custom {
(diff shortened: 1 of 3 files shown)


include/custom_op/kernel_context.h
```diff
@@ -29,6 +30,7 @@ class CUDAKernelContext :  public KernelContext {
   virtual void* GetCudaStream() const = 0;
   virtual void* GetCublasHandle() const = 0;
   virtual int GetCudaDeviceId() const = 0;
+  virtual void* GetScratchBufferUnderMultiStream(const OrtMemoryInfo* , size_t ) { return nullptr; }
```
wenbingl, line 33: I can you cannot use ORT type here since these classes will be used in eager mode which doesn't depend on ORT

souptc: yes, we need to decouple it from ORT types.

PR #729, opened 2024-05-24
Add initial tiktoken and Phi3SmallTokenizer support

sayanshaw24:
This PR adds official support in ORT Extensions for the new fast [tiktoken](https://github.com/openai/tiktoken) format introduced by OpenAI for fast loading for tokenizers, as used in the Phi3SmallTokenizer. Currently, encoding is fully supported and tested. Will add decoder support in a separate PR.


operators/tokenizer/bpe_kernels.cc
```diff
@@ -641,3 +643,147 @@ OrtxStatus JsonFastTokenizer::Compute(const ortc::Tensor<std::string>& input,
...
+};
+
+OrtxStatus TikTokenizer::Load(const ort_extensions::bpe::TokenJsonConfig& config) {
+  std::string voc_file = config.GetVocabDataFile();
+  std::ifstream ifs = path(voc_file).open();
+  if (!ifs.is_open()) {
+    return OrtxStatus(kOrtxErrorInvalidFile, "Failed to open vocab file: " + voc_file);
+  }
+
+  std::unordered_map<std::vector<uint8_t>, uint32_t, VectorHash, VectorEqual> bpe_ranks;
```
wenbingl, line 683: why not directly use std::string instead of std::vector<uint8_t>?
std::string also stores the length of the string so you don't need to worry about '\x0' char in the string

operators/tokenizer/bpe_kernels.cc
```diff
@@ -641,3 +643,147 @@ OrtxStatus JsonFastTokenizer::Compute(const ortc::Tensor<std::string>& input,
...
+    return status;
+  }
+
+  add_bos_token_ = config.add_bos_token_;
+  add_eos_token_ = config.add_eos_token_;
+
+  return status;
+}
+
+OrtxStatus TikTokenizer::Compute(const ortc::Tensor<std::string>& input,
```
wenbingl, line 784: instead of the origin BPE algorithm, can we just use tiktoken tokenize method to compute?

sayanshaw24: i believe after loading the bpe ranks from the .tiktoken file, they also use BPE with a similar algorithm, but i will investigate and improve the compute more after ensuring the loading works firstly.

operators/tokenizer/bpe_kernels.cc
sayanshaw24: because we are decoding the strings in base64 in the .tiktoken file from std::string to std::vector<uint8_t> first - let me know if you think this thinking is correct.

operators/tokenizer/bpe_kernels.cc
wenbingl: It looks like tiktoken compute equivalent is efficient, https://github.com/openai/tiktoken/blob/c0ba74c238d18b4824c25f3c27fc8698055b9a76/src/lib.rs#L226
here, it worths to have a try.

operators/tokenizer/bpe_kernels.cc
wenbingl: you can convert std::vector<int8_t> into a std::string, so that we don't need these map function and a simpler type define,

operators/tokenizer/bpe_json.hpp
```diff
@@ -9,6 +9,8 @@
 
 #include "bpe_types.h"
 
+#include <filesystem>
```
wenbingl, line 12: filesystem is not available for all platforms that ort-extensions needs to support, try to use some C API.

wenbingl (commented):
Where is the unit test?

sayanshaw24:
> Where is the unit test?

the unit test you included for pp_api_tests in your PR here: https://github.com/microsoft/onnxruntime-extensions/pull/726/files#diff-f4495557b563fdd392dbf72a0969d541e8288335dc2cd2e48b44d093de843cbfR121 will use this implementation as I moved the test data files into test/data2 so it will run that test, recognize there is no tokenizer.json and use the TikTokenizer.

it works successfully as all pp_api_tests pass here: https://dev.azure.com/onnxruntime/onnxruntime/_build/results?buildId=1398194&view=logs&j=719f0e07-ada6-5167-1399-95b4f6fab546&t=8a37696c-a395-5292-3877-9451da366bed&l=1653

operators/tokenizer/bpe_kernels.cc
sayanshaw24: another (and more crucial) reason that the bpe_ranks need to have the vector of bytes is because those bytes get encoded using our ByteEncoder first; after, actually the vocab map here: https://github.com/microsoft/onnxruntime-extensions/pull/729/files#diff-525f03c6db82461d5659425f865848836162b44f0559d7338f11a93f9c3f28bdR701 uses strings - so the string conversion does happen for simpler types, just later in the primitive.

wenbingl:
> > Where is the unit test?
> 
> the unit test you included for pp_api_tests in your PR here: https://github.com/microsoft/onnxruntime-extensions/pull/726/files#diff-f4495557b563fdd392dbf72a0969d541e8288335dc2cd2e48b44d093de843cbfR121 will use this implementation as I moved the test data files into test/data2 so it will run that test, recognize there is no tokenizer.json and use the TikTokenizer.
> 
> it works successfully as all pp_api_tests pass here: https://dev.azure.com/onnxruntime/onnxruntime/_build/results?buildId=1398194&view=logs&j=719f0e07-ada6-5167-1399-95b4f6fab546&t=8a37696c-a395-5292-3877-9451da366bed&l=1653

that was testing a converted HF json file, instead of tiktoken file. Please create your own test.

operators/tokenizer/bpe_kernels.cc
sayanshaw24: sounds good, will have another PR for this in the future to optimize the initial implementation.

sayanshaw24:
> > > Where is the unit test?
> > 
> > 
> > the unit test you included for pp_api_tests in your PR here: https://github.com/microsoft/onnxruntime-extensions/pull/726/files#diff-f4495557b563fdd392dbf72a0969d541e8288335dc2cd2e48b44d093de843cbfR121 will use this implementation as I moved the test data files into test/data2 so it will run that test, recognize there is no tokenizer.json and use the TikTokenizer.
> > it works successfully as all pp_api_tests pass here: https://dev.azure.com/onnxruntime/onnxruntime/_build/results?buildId=1398194&view=logs&j=719f0e07-ada6-5167-1399-95b4f6fab546&t=8a37696c-a395-5292-3877-9451da366bed&l=1653
> 
> that was testing a converted HF json file, instead of tiktoken file. Please create your own test.

I thought this one was for the converted HF json file?: https://github.com/microsoft/onnxruntime-extensions/blob/main/test/pp_api_test/test_tokenizer.cc#L93 i remember when you added this and data/tiktoken even has the converted tokenizer.json file and when I asked you mentioned this was the converted HF json file.

This one is different: https://github.com/microsoft/onnxruntime-extensions/blob/main/test/pp_api_test/test_tokenizer.cc#L120 maybe you added it twice accidentally or even for local testing? I had basically the same test locally but did not stage it since you already added it. It's not being run as it stands right now because there is nothing in test/data2 folder, so it is skipped. With this PR, it is being run because I added the necessary .tiktoken and tokenizer_config.json files in test/data2/phi-3-small.

But regardless, I have no issues adding another UT - let me know if you would like me to. Just wanted to clarify about this benign test at the moment since Phi3_S_Tokenizer right now is not run (I thought you added it for this purpose).

Sorry for the confusion.

test/data/phi-3-small/tokenizer_module.json
```diff
@@ -0,0 +1,185 @@
...
+      "content": "<|dummy_id_0|>",
+      "single_word": false,
+      "lstrip": false,
+      "rstrip": false,
+      "normalized": false,
+      "special": true
+    },
+    {
+      "id": 100265,
+      "content": "<|dummy_id_1|>",
```
wenbingl, line 78: any idea where are these dummy_ids used?

operators/tokenizer/bpe_kernels.h
```diff
@@ -107,6 +107,10 @@ struct SpmTokenizer : KernelBpeTokenizer {
 class JsonFastTokenizer : KernelBpeTokenizer {
  public:
   JsonFastTokenizer();
+  bool tiktoken;
```
wenbingl, line 110: member variable naming convention, and the initial value

operators/tokenizer/bpe_json.hpp
```diff
@@ -30,7 +32,24 @@ class TokenJsonConfig final {
...
+      auto module_file_path = path(json_path) / "tokenizer_module.json";
+      module_path_ = module_file_path.string();
+      std::ifstream tok_module_ifs = module_file_path.open();
+      if (!tok_module_ifs.is_open()) {
+        return OrtxStatus(kOrtxErrorInvalidFile, "No tokenizer.json or tokenizer_module.json file found.");
+      } else {
+        nlohmann::json tok_module_json_config = nlohmann::json::parse(tok_module_ifs);
+        auto tiktoken_path = tok_module_json_config.value("tiktoken_file", "");
+        vocab_file_path = path(json_path) / tiktoken_path.c_str();
+        vocab_file_path = path(json_path) / "cl100k_base.tiktoken";
```
wenbingl, line 49: why not get it from json config file?

operators/tokenizer/bpe_json.hpp
wenbingl, line 12: it looks like we don't use any fancy feature from filesystem, can we remove the dependency now and use file_sys.h instead?

operators/tokenizer/bpe_kernels.cc
```diff
@@ -552,13 +554,140 @@ SpmTokenizer::SpmTokenizer()
 
 JsonFastTokenizer::JsonFastTokenizer() : KernelBpeTokenizer(kGPT2Configuration) {}
 
+/*
+Read more here: https://github.com/huggingface/transformers/blob/60bb571e993b7d73257fb64044726b569fef9403/src/transformers/convert_slow_tokenizer.py#L1454
+
+Note: this is similar to the BPE CreateByteEncoder, however for decoding the .tiktoken bytes
+we need to store the strings rather than their IDs, and thereby need a separate map.
+*/
+void JsonFastTokenizer::CreateUnicodeByteEncoder() {
```
wenbingl, line 563: what's the difference from BpeModel::CreateByteEncoder?

operators/tokenizer/bpe_kernels.cc
```diff
@@ -570,18 +699,27 @@ OrtxStatus JsonFastTokenizer::Load(const ort_extensions::bpe::TokenJsonConfig& c
...
-  auto model_node = tok_json.find("model");
-  if (model_node == tok_json.end()) {
-    return OrtxStatus(kOrtxErrorCorruptData, "Failed to get model node from tokenizer.json");
-  }
+  OrtxStatus status;
+  if (tiktoken){
+    status = bbpe_tokenizer_->Load(vocab,
+                                   merges,
+                                   bpe_conf_.get().GetSpecialTokens().c_str(),
+                                   IsSpmModel(ModelName()));
```
wenbingl, line 707: it should be always false

operators/tokenizer/bpe_kernels.cc
```diff
@@ -552,13 +554,140 @@ SpmTokenizer::SpmTokenizer()
...
+        }
+      }
+    }
+
+    std::vector<std::tuple<std::vector<uint8_t>, std::vector<uint8_t>, uint32_t>> byte_merges;
+
+    bbpe_tokenizer_ = std::make_unique<BpeModel>();
+    JsonFastTokenizer::CreateUnicodeByteEncoder();
+    
+    for (const auto& item : bpe_ranks) {
```
wenbingl, line 647: defining bpe_ranks as a vector seems more simple, why did it have to be a map?

operators/tokenizer/bpe_kernels.cc
```diff
@@ -552,13 +554,140 @@ SpmTokenizer::SpmTokenizer()
...
+      byte_merges.insert(byte_merges.end(), local.begin(), local.end());
+    }
+
+    // Custom comparator that compares the third element of the tuples
+    auto compare_merge_tuples = [&](const std::tuple<std::vector<uint8_t>, std::vector<uint8_t>, uint32_t>& a,
+                                        const std::tuple<std::vector<uint8_t>, std::vector<uint8_t>, uint32_t>& b) {
+      return std::get<2>(a) < std::get<2>(b);
+    };
+    
+    std::sort(byte_merges.begin(), byte_merges.end(), compare_merge_tuples);
```
wenbingl, line 683: were the ranks already sorted in titoken file?

test/data/phi-3-small/tokenizer_module.json
sayanshaw24: explanation by HF here: https://huggingface.co/microsoft/Phi-3-small-8k-instruct/blob/main/tokenization_phi3_small.py#L59, but TLDR is that it is for continuous tokenization.

operators/tokenizer/bpe_json.hpp
sayanshaw24: good catch, just missed deleting the second line, will update in latest commit.

operators/tokenizer/bpe_json.hpp
sayanshaw24: yep! good catch, removed.

operators/tokenizer/bpe_kernels.cc
sayanshaw24: BpeModel::CreateByteEncoder creates a map where the values are the token ids (uint32s) and uses the input ids for this map, e.g. `byte_encoder_[i] = GetTokenId(ustring::EncodeUTF8Char(i))`.

CreateUnicodeByteEncoder creates a map where the values are the token strings, and needs to be used just to convert bytes to string (including unicodes), and therefore does not need to get the token ids (which it also cannot at that step because we need to process the tiktoken file first).

operators/tokenizer/bpe_kernels.cc
sayanshaw24: no, afaik from debugging, i do not believe so; this step is also done by HF here: https://github.com/huggingface/transformers/blob/60bb571e993b7d73257fb64044726b569fef9403/src/transformers/convert_slow_tokenizer.py#L1520

operators/tokenizer/bpe_kernels.cc
sayanshaw24: because we need a dict to get the rank given a token: https://github.com/openai/tiktoken/blob/c0ba74c238d18b4824c25f3c27fc8698055b9a76/tiktoken/load.py#L143

wenbingl (approved):

PR #731, opened 2024-05-28
Fix component governance issues.

skottmckay:
Fix component governance warnings:
- update protobuf version being used by sentencepiece and the java tests
- ignore unused language bindings from protobuf and triton
- specify the CG config file with ignored directories where required

Fix cgmanifest.json:
- 'git' entries require a commit hash not version
- use 'other' for opencv third party code that is included directly in the opencv repo
  - the path isn't a valid repositoryUrl value to be provided as a 'git' entry
- update version numbers/commit hashes to match the latest code

### DIFF
--- a/cgmanifest.json
+++ b/cgmanifest.json
@@ -5,18 +5,20 @@
       "component": {
         "type": "git",
         "git": {
-          "commitHash": "v1.14.1",
+          "commitHash": "8f5c79cb63f09ef1302e85081093a3fe4da1bc7d",
           "repositoryUrl": "https://github.com/microsoft/onnxruntime.git"
-        }
+        },
+        "comments": "v1.17.1"
       }
     },
     {
       "component": {
         "type": "git",
         "git": {
-          "commitHash": "6e511679de8ab0feefc1cdac1505b2fac5548e42",
+          "commitHash": "fe271ab76f2ad2b2b28c10443865d2af21e27e0e",
           "repositoryUrl": "https://github.com/protocolbuffers/protobuf.git"
-        }
+        },
+        "comments": "v3.20.3"
       }
     },
     {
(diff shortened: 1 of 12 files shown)


java/build.gradle
```diff
@@ -148,7 +148,7 @@ if (cmakeBuildDir != null) {
 dependencies {
     testImplementation 'org.junit.jupiter:junit-jupiter-api:5.7.0'
     testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.7.0'
-    testImplementation 'com.google.protobuf:protobuf-java:3.20.1'
+    testImplementation 'com.google.protobuf:protobuf-java:3.21.7'
```
skottmckay, line 151: This matches the version used in build-android.gradle.

cgmanifest.json
```diff
@@ -50,16 +52,17 @@
...
+        },
+        "comments": "2021-06-01"
       }
     },
     {
       "component": {
         "type": "git",
         "git": {
-          "commitHash": "release-1.11.0",
+          "commitHash": "9406a60c7839052e4944ea4dbc8344762a89f9bd",
```
skottmckay, line 65: Not sure what tag this commitHash matches as there's no comment in the cmake file. 

https://github.com/microsoft/onnxruntime-extensions/blob/a1e342acb94c4b28c74c38e5b3a235824edb98fc/cmake/externals/googletest.cmake#L1-L5

The commit id doesn't match release-1.11.0. The readme.md in the googletest source says it's release-1.12.0, but the commit id for that is different.

https://github.com/google/googletest/releases/tag/release-1.12.0

wenbingl: Yes, it was upgraded due to conflict with AzureOP build in CI pipeline.

wenbingl (approved):

PR #738, opened 2024-06-06
Add custom ops Rotary

xadupre:


wenbingl (dismissed):

operators/cuda/cuda_ops.cc
```diff
@@ -7,7 +7,11 @@
 #include "cuda/add_mul.h"
 #include "cuda/fast_gelu.h"
 #include "cuda/negxplus1.h"
+<<<<<<< HEAD
```
wenbingl, line 10: A git merge issue.

xadupre: thanks, fixed.

operators/cuda/rotary_impl.cu
```diff
@@ -0,0 +1,81 @@
...
+
+template <typename T> __device__ __inline__ T _neg(const T x) { return -x; }
+
+#if __CUDA_ARCH__ < 700
+template <> __device__ __inline__ half _neg(const half x) {
+  return __float2half(-__half2float(x));
+}
+#endif
+
+template <typename T, RotarySide side>
```
wenbingl, line 23: A question: RotarySide being a template parameter is more efficient for CUDA, or any other reason?

wenbingl (dismissed):

operators/cuda/rotary_impl.cu
xadupre: Yes, to avoid doing the test during the execution.

operators/cuda/rotary.h
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+*  Y = Rotary(X) is equivalent to if side == LEFT:
+*
+*   N = X.shape[-1]
+*   Y = X.copy()
+*   Y[...,:N/2] = X[...,N/2:]
+*   Y[...,N/2:] = -X[...,:N/2]
+*
+*   And the opposite if side == RIGHT.
```
wschin, line 19: Could you write the formula down? `opposite` could mean
```
N = X.shape[0]
```
or
```
Y[:N/2, ...] = X[N/2:, ...]
```
etc.

operators/cuda/rotary.h
```diff
@@ -0,0 +1,83 @@
...
+*/
+template <typename T>
+struct Rotary {
+  template <typename TDict>
+  OrtxStatus OnModelAttach(const TDict& dict) {
+    std::string empty;
+    std::string side = dict.TryToGetAttributeWithDefault("side", empty);
```
wschin, line 26: Maybe we should do `LeftRotary` and `RightRotary` so that the compute function becomes stateless.

operators/cuda/rotary.h
```diff
@@ -0,0 +1,83 @@
...
+    else if (side == "right") {
+      side_ = RotarySide::RIGHT;
+    }
+    else {
+      return {kOrtxErrorInvalidArgument, "side must be 'left' or 'right'."};
+    }
+
+    return {};
+  }
+  OrtxStatus Compute(Ort::Custom::CUDAKernelContext* ctx,
```
wschin, line 39: How is `split` related to the formula above?

operators/cuda/rotary.h
```diff
@@ -0,0 +1,83 @@
...
+      return {};
+    }
+
+    auto shape_split = split.Shape();
+    if (shape_split.size() != 1 || shape_split[0] != 2) {
+      return {kOrtxErrorInvalidArgument, "Rotary only works when there are two sides."};
+    }
+    const int64_t* split_data = split.Data();
+    if (split_data[0] != split_data[1]) {
+      return {kOrtxErrorInvalidArgument, "Only equal split are allowed."};
```
wschin, line 57: are -> is?

operators/cuda/rotary.h
```diff
@@ -0,0 +1,83 @@
...
+    const T* input_data = input.Data();
+    auto input_shape = input.Shape();
+    T* output_data = output.Allocate(input_shape);
+    auto input_length = input.NumberOfElement();
+    if (0 == input_length) {
+      return {};
+    }
+
+    auto shape_split = split.Shape();
+    if (shape_split.size() != 1 || shape_split[0] != 2) {
```
wschin, line 52: What are `shape_split[0]` and `shape_split[1]`?

operators/cuda/rotary.h
```diff
@@ -0,0 +1,83 @@
...
+      return {kOrtxErrorInvalidArgument, "Rotary only works when there are two sides."};
+    }
+    const int64_t* split_data = split.Data();
+    if (split_data[0] != split_data[1]) {
+      return {kOrtxErrorInvalidArgument, "Only equal split are allowed."};
+    }
+    if (split_data[0] * 2 != input_shape[input_shape.size()-1]) {
+      return {kOrtxErrorInvalidArgument, "Sum of the splits are not equal to the last dimension."};
```
wschin, line 60: Sum of split sounds `split_data[0] + split_data[1]`. However, here it uses `split_data[0] * 2`.

operators/cuda/rotary.h
```diff
@@ -0,0 +1,83 @@
...
+    }
+    else {
+      return {kOrtxErrorInvalidArgument, "side must be 'left' or 'right'."};
+    }
+
+    return {};
+  }
+  OrtxStatus Compute(Ort::Custom::CUDAKernelContext* ctx,
+                       const ortc::Tensor<T>& input,
+                       const ortc::Tensor<int64_t>& split,
```
wschin, line 41: Maybe throw if this tensor is not on CPU.

operators/cuda/rotary.h
```diff
@@ -0,0 +1,83 @@
...
+    if (split_data[0] != split_data[1]) {
+      return {kOrtxErrorInvalidArgument, "Only equal split are allowed."};
+    }
+    if (split_data[0] * 2 != input_shape[input_shape.size()-1]) {
+      return {kOrtxErrorInvalidArgument, "Sum of the splits are not equal to the last dimension."};
+    }
+
+    LaunchRotaryKernel<T>(reinterpret_cast<cudaStream_t>(ctx->GetCudaStream()),
+                             input_length,
+                             static_cast<int>(input_shape[input_shape.size()-1]),
```
wschin, line 65: Consider format all files once.

wschin (changes requested):
My comments fall into two categories. For documentation, they are must-to-have. The rest are nit and optional. Thank you.

operators/cuda/rotary.h
xadupre: The Rotary replaces something like:

```
x1, x2 = Split(X, split, axis=-1)
Y = Concat(-x2, x1, axis=-1)
```

When I implemented this kernel, I found it was equal splits on llama but I wondered if it could be not equal splits so I chose to leave this parameter so that I could still use the optimized and let the kernel fails when the split are not equal. Then I would know not equal splits must be implemented. I can remove if you know this case never happens.

operators/cuda/rotary.h
xadupre: done

operators/cuda/rotary.h
xadupre: True, the condition is valid because the test before checks that they are equal but it is better to change it if the test is moved.

operators/cuda/rotary.h
xadupre: I assumed ort would ensure that condition is true since the kernel requires that input to be on CPU. Is the check still necessary?

operators/cuda/rotary.h
xadupre: done

operators/cuda/rotary.h
xadupre: See my comment above.

operators/cuda/rotary.h
xadupre: done

operators/cuda/rotary.h
xadupre: I think this is done later here: https://github.com/microsoft/onnxruntime-extensions/pull/738/files#diff-643fdcb552aafbcb9a86bc8d48cab50dea3723b20b3105880e081b2e5b1b8da9R24. I hope the compiler is able to remove the unnecessary code at compilation time.

operators/cuda/rotary.h
wschin: Please make sure the `exact` behavior is described in comments for this op. I don't see the use and explanation of `split` in the equations. In addition to math symbols, you can use ONNX sub-graph to describe them.

operators/cuda/rotary.h
wschin: No idea. nit. :)

operators/cuda/rotary.h
wschin: I think we can assume even split and throw otherwise. With `split`, you will need to introduce a sub-graph to compute the right `split` from the shape of `X`. That will slows down ORT..

operators/cuda/rotary.h
wschin: It becomes more reasonable to split them in the beginning instead of relying on compiler. If they are two ops, we define two ops. No need to make a mega op to support several cases; this will make future improvement harder and error-prone.

operators/cuda/rotary.h
xadupre: I don't think so. The formula remains the same, whether or not the splits are equal or not.

operators/cuda/rotary.h
xadupre: That's not the only place I did that. There is only one op from ort point of view and it is Rotary but we have different implementation depending the argument value, the same goes with the input types. Rotary(X, side=LEFT) = Neg(Rotary(X, side=RIGHT)). Do you want to have distinct onnx names as well?

xadupre: I suggest keeping the names this way. I don't think it is worth adding new names for left and right.

PR #739, opened 2024-06-06
Add custom ops ReplaceZero

xadupre:


operators/cuda/cuda_ops.cc
```diff
@@ -31,13 +32,15 @@ FxLoadCustomOpFactory LoadCustomOpClasses_Contrib = []() -> CustomOpArray& {
       CustomCudaStructV2("FastGelu", contrib::FastGelu<float>),
       CustomCudaStructV2("MulSharedInput", MulSharedInputFloat32Type),
       CustomCudaStructV2("NegXPlus1", contrib::NegXPlus1<float>),
+      CustomCudaStructV2("ReplaceZero", contrib::ReplaceZero<float>),
```
wschin, line 35: Not sure how the kernel selection works. Do we need `half`, `int64_t`, etc. as template parameters?

wschin (commented):
Please consider
1. format the code
2. add a few more details. (1) how to run this test, (2) kernel selection mechanism to explain why you define two kernels for fp16 and fp32.
3. avoid abbreviations as function name.

wschin:
Please ping me when you want me to review again. Thanks.

operators/cuda/cuda_ops.cc
xadupre: All templates are implemented for float and float16 because that's the only types i need. Adding bfloat16 makes sense as well. Integer does not really make sense. I kept the list short to avoid growing the binary. I can add bfloat16 if you think that makes sense.

operators/cuda/replace_zero_impl.cu
```diff
@@ -0,0 +1,64 @@
...
+
+template <typename T>
+__global__ void ReplaceZeroKernel(T *output_data, const T *input_data, CUDA_LONG N, const T by) {
+  CUDA_LONG id = blockDim.x * blockIdx.x + threadIdx.x;
+  if (id >= N)
+    return;
+  output_data[id] = _replace_zero(input_data[id], by);
+}
+
+template <typename T> T _cvt(float value) { return (T)value; }
```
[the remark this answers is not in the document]
xadupre: renamed into _cast

xadupre:
I chose cvt for convert, I thought it was a common abbrevation. I replaced by cast. For float32, float16, we can extend to bfloat16 if needed. I chose not to reduce compilation time but we definitly can add it. What do you mean by *how to run the unit tests*? There is one file associated to all the kernel implemented in that folder: ``test/cuda/test_cudaops.py``. I just extended it with one of two models per kernel.

wschin (approved):
Thanks!

PR #743, opened 2024-06-10
Add more fused element-wise operators

xadupre:


operators/cuda/add_mul.h
```diff
@@ -20,27 +41,156 @@ struct AddOrMulSharedInput {
...
+                                       tensor_a.Data(), tensor_b.Data(), tensor_c.Data(),
                                        output_data_ab, output_data_ac,
                                        length_a, length_b, length_c,
                                        addition);
     return {};
   }
 };
 
+template <typename T, bool addition>
+struct AddOrMulTwice {
```
wschin, line 65: what's formulation implemented? Every math kernel has to have exact math equation implemented.

operators/cuda/add_mul.h
```diff
@@ -20,27 +41,156 @@ struct AddOrMulSharedInput {
...
+    }
+    return {};
+  }
+
+ private:
+  bool switchMiddelAxis_;
+};
+
+template <typename T, bool subtract_first>
+struct SubAndMul {
```
wschin, line 158: Please add exact math equation of this kernel.

operators/cuda/add_mul.h
xadupre: done

operators/cuda/add_mul.h
xadupre: done

operators/cuda/add_mul.h
```diff
@@ -8,6 +8,35 @@
 
 namespace contrib {
 
+inline void _FillOutputShape3Op(std::vector<int64_t>& dimsA,
```
wschin, line 11: Consider using `const std::vector<int64_t>&` for fixed inputs.

operators/cuda/add_mul.h
```diff
@@ -8,6 +8,35 @@
...
+    }
+}
+
+/**
+*  AddOrMulSharedInput(A, B, C) = A + B, A + C ifaddition is true
+*  AddOrMulSharedInput(A, B, C) = A * B, A * C ifaddition is false
+*
+*  The operator supports broadcast on first dimensions.
+*  A[1, J] + B[I, J] is supported,
+*  A[1, J, 1] + B[I, J, K] is not supported,
```
wschin, line 38:
```suggestion
*  A[1, J, 1] + B[I, J, K] is not supported,
*  In all other cases, all tensors must have the same shape.
```

operators/cuda/add_mul.h
```diff
@@ -20,27 +49,186 @@ struct AddOrMulSharedInput {
...
                      ortc::Tensor<T>& output_ac) const {
-    const T* input_data_a = tensor_a.Data();
-    const T* input_data_b = tensor_b.Data();
-    const T* input_data_c = tensor_c.Data();
-
     auto length_a = tensor_a.NumberOfElement();
     auto length_b = tensor_b.NumberOfElement();
     auto length_c = tensor_c.NumberOfElement();
 
+    if (0 == length_a || 0 == length_b || 0 == length_c) {
```
wschin, line 56: Should this be an assert?

operators/cuda/add_mul.h
xadupre: None of them are fixed.

operators/cuda/add_mul.h
xadupre: I used fast_gelu.h as an example and it returns a null tensor when the input is null. I followed the same design.

operators/cuda/add_mul.h
```diff
@@ -20,27 +50,186 @@ struct AddOrMulSharedInput {
...
+
+/**
+*  SubAndMul(A, B, C) = (A - B) * C if subtract_first is true
+*  SubAndMul(A, B, C) = A * B - C if subtract_first is false
+*
+*  The operator supports broadcast on first dimensions.
+*  A[1, J] + B[I, J] is supported,
+*  A[1, J, 1] + B[I, J, K] is not supported,
+*
+* If negative is true, then the output is transposed, then
```
wschin, line 193: The name `negative` sounds confusing. Why does `negative` control `transpose`? From the equation below, it looks like a typo.

wschin (dismissed):
Overall LGTM but please still fix the typo.

operators/cuda/add_mul.h
xadupre: There is no transpose on this one. Instead of doing A - B, it does B - A. I'll remove the confusing help. Some wrong copy/pasting.

operators/cuda/add_mul.h
```diff
@@ -190,7 +190,7 @@ struct AddAndMul {
 *  A[1, J] + B[I, J] is supported,
 *  A[1, J, 1] + B[I, J, K] is not supported,
 *
-* If negative is true, then the output is transposed, then
```
wschin: Does this op support `subtract_first=false` and `negative=true` and all other combinations?

wschin (commented):
Please make sure the math documentation covers all expected behaviors of the op (not only in this PR). They are the golden standard needed for code quality. Once you fix them, please ping me again. Thanks.

operators/cuda/add_mul.h
xadupre: Yes, I chose the following convention: a parameter is a template parameter if it changes the kernel name. So substract_first is not seen by the user but it implicitly set it up by choosing op SubMul or MulSub. A parameter is an argument of the kernel if the name remains the same.

PR #745, opened 2024-06-12
Add a generic image processor and its C API

wenbingl:


.clang-format
```diff
@@ -1,10 +1,7 @@
 ---
 # Defaults for all languages.
 BasedOnStyle:  Google
-
-# Setting ColumnLimit to 0 so developer choices about where to break lines are maintained.
-# Developers are responsible for adhering to the 120 character maximum.
-ColumnLimit:     0
```
wenbingl, line 7: it will cause VS auto-formatter generating a very long line...

shared/api/image_transforms_phi_3.hpp
```diff
@@ -0,0 +1,209 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
sayanshaw24, line 1: nit: as our list of processor files slowly grows, should we perhaps group the files inside shared/api? perhaps into `c_api`, `imaging`, and `tokenizing` - we can do this organizational work in a future PR.

sayanshaw24 (approved):

shared/api/image_transforms_phi_3.hpp
wenbingl: After audio C abi, we may revisit it and re-org the folder if needed

sayanshaw24 (approved):

PR #749, opened 2024-06-20
Suppress C4459 warning in custom_op_lite.h

skyline75489:
Seems to a false positive. I found no `api` in global namespace.

### DIFF
--- a/include/custom_op/custom_op_lite.h
+++ b/include/custom_op/custom_op_lite.h
@@ -11,6 +11,9 @@
 #include "ort_c_to_cpp.h"
 #include "onnxruntime_f16.h"
 
+#pragma warning(push)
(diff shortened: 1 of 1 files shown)


include/custom_op/custom_op_lite.h
```diff
@@ -11,6 +11,9 @@
 #include "ort_c_to_cpp.h"
 #include "onnxruntime_f16.h"
 
+#pragma warning(push)
+#pragma warning(disable : 4459)  // declaration of 'api' hides global declaration
```
wenbingl, line 15: cannot disable the warning, which can be detected some security check tools, and it eventually become a server security issue in some dashboard.

skyline75489: OK I'll see if there's other way to fix it

PR #755, opened 2024-06-27
Feature extraction C API for whipser model

wenbingl:
1. Add a feature extraction API for speech data processing.
2. Refactor the existing audio processing kernels
3. Adjust image processing C API for unification of APIs.


shared/api/c_api_processor.cc
```diff
@@ -67,59 +69,38 @@ extError_t ORTX_API_CALL OrtxImagePreProcess(OrtxProcessor* processor, OrtxRawIm
...
-  auto result_ptr = static_cast<ImageProcessorResult*>(result);
-  status = result_ptr->IsInstanceOf(extObjectKind_t::kOrtxKindImageProcessorResult);
-  if (!status.IsOk()) {
-    return status.Code();
-  }
-
-  ImageProcessor::ClearOutputs(result_ptr);
-  return extError_t();
-}
+// extError_t ORTX_API_CALL OrtxImageGetTensorResult(OrtxImageProcessorResult* result, size_t index, OrtxTensor**
```
sayanshaw24, line 84: nit: did you want to delete this commented code or do we need this OrtxImageGetTensorResult impl checked in? maybe we can just remove it to clean up the code (can be done in future PR) and if we ever need to look at it for reference we could always check the git blame?

sayanshaw24 (dismissed):

shared/api/c_api_processor.cc
wenbingl: cleaned up

sayanshaw24 (approved):

PR #757, opened 2024-07-01
sync to flash attention kernel 2.5.9 and add document of how to write custom op

jslhcl:
sync to flash attention kernel 2.5.9 and add document of how to write custom op

### DIFF
--- a/onnxruntime_extensions/__init__.py
+++ b/onnxruntime_extensions/__init__.py
@@ -10,6 +10,10 @@ This enables more flexibility and control over model execution, thus expanding t
 
 __author__ = "Microsoft"
 
+import importlib.util
+spec = importlib.util.find_spec('pybind11')
+if spec is None:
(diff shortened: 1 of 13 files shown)


onnxruntime_extensions/__init__.py
```diff
@@ -10,6 +10,10 @@
 
 __author__ = "Microsoft"
 
+import importlib.util
+spec = importlib.util.find_spec('pybind11')
```
wenbingl, line 14: why is pybind11 needed?

jslhcl: pybind11 is required by _extensions_pydll. It will throw an exception: ModuleNotFoundError: No module named 'onnxruntime_extensions._extensions_pydll when importing onnxruntime_extensions when pybind11 is not installed

jslhcl: removed the change for pybind11 as we discussed

wenbingl (approved):

PR #762, opened 2024-07-12
switch cmake cmp0169 flag to new

wenbingl:
cmp0169 is to deprecate FetchContent_Popluate, this PR will follow this guideline and only use FetchContent_MakeAvailable


cmake/externals/dlib.cmake
```diff
@@ -3,10 +3,7 @@ FetchContent_Declare(dlib
     # there is non an official tag which supports STFT,
     # choose a relatively stable commit id for that.
     GIT_TAG        a12824d42584e292ecb3bad05c4b32c2015a7b89
+    SOURCE_SUBDIR  not_set
```
snnn, line 6: ”SOURCE_SUBDIR not_set“ 
Is it needed?

wenbingl: Yes, it's intentional for the library which we want to use it in a C++ header-only way.

cmake/externals/opencv.cmake
```diff
@@ -98,6 +98,7 @@ if (OCOS_ENABLE_OPENCV_CODECS)
 
   set(WITH_JPEG               ON CACHE INTERNAL "")
   set(WITH_PNG                ON CACHE INTERNAL "")
+  set(BUILD_JPEG_TURBO_DISABLE  ON CACHE INTERNAL "")
```
wenbingl, line 101: libjpeg-turbo is not listed in cgmanifest

snnn (approved):

PR #780, opened 2024-07-31
Upgrade ESRP signing task from v2 to v5 

wenbingl:
Upgrade ESRP signing task from v2 to v5 because 'OnnxRuntime CodeSign 20190817' doesn't exist anymore.

Configuration is copied from this PR: https://github.com/microsoft/onnxruntime/pull/20995


sayanshaw24 (approved):

mszhanyi (approved):

wenbingl:
> It‘d better to double check the generated binaries

Thanks, Do we have any PRs that can be used as a reference?

PR #781, opened 2024-07-31
Fix the case that bos_token is null, like qwen2 tokenizer

wenbingl:


operators/tokenizer/bpe_json.hpp
```diff
@@ -40,14 +40,16 @@ class TokenJsonConfig final {
...
     auto tok_iter = json_config.find("bos_token");
-    if (tok_iter != json_config.end() && tok_iter->is_object()) {
-      bos_token_ = tok_iter->value("content", "");
-      eos_token_ = json_config.value("/eos_token/content"_json_pointer, "");
-      unk_token_ = json_config.value("/unk_token/content"_json_pointer, "");
-    } else {
-      bos_token_ = json_config.value("bos_token", "");
-      eos_token_ = json_config.value("eos_token", "");
-      unk_token_ = json_config.value("unk_token", "");
+    if (tok_iter != json_config.end() && !tok_iter->is_null()) {
```
yufenglee, line 43: if(tok_iter != json_config.end() && !tok_iter->is_null() && tok_iter->is_object()) then else?

wenbingl: the condition and the previous impl are buggy since the else statement could read bos_token for a null or non-existing json node.

yufenglee (approved):

PR #800, opened 2024-08-30
Remove OpenCV dependency from C_API mode

wenbingl:
Needs some unit tests.


cmake/ext_imgcodecs.cmake
```diff
@@ -0,0 +1,131 @@
...
+  if(NOT M_LIBRARY)
+    message(STATUS "math lib 'libm' not found; floating point support disabled")
+  endif()
+else()
+  # not needed on windows
+  set(M_LIBRARY "")
+endif()
+
+set(lib_srcs
+   ${libPNG_SOURCE_DIR}/arm/arm_init.c
```
sayanshaw24, line 28: is there a reason we include all these files individually rather than put them in folders and include the folders instead for efficiency? (same for libjpeg)

wenbingl: Yes, there are some unnecessary source files in the folder, and cannot be compiled easily. e.g. a lot of build configuraiton and tunning.

sayanshaw24 (approved):

PR #802, opened 2024-09-05
Fix unused result warnings

skyline75489:
Fixed unused result warning from clang.


include/custom_op/custom_op_lite.h
```diff
@@ -458,16 +458,16 @@ class OrtGraphCudaKernelContext : public CUDAKernelContext {
   static const int cuda_resource_ver = 1;
 
   OrtGraphCudaKernelContext(const OrtApi& ort_api, const OrtKernelContext& ctx) : api_(ort_api) {
-    api_.KernelContext_GetResource(&ctx, cuda_resource_ver, CudaResource::cuda_handle_t, &cuda_stream_);
+    OrtStatusPtr result = api_.KernelContext_GetResource(&ctx, cuda_resource_ver, CudaResource::cuda_handle_t, &cuda_stream_);
     if (!cuda_stream_) {
```
wenbingl, line 462: should we also check the result is not nullptr here to avoid a potential unused the variable warning?

skyline75489: Yeah sure. I also added the check for `resource != nullptr`, just in case.

skyline75489:
I wonder if the CUDA test failures are related to my change ... Or am I misunderstanding the `result` here.

xadupre:
You can split the test to make sure with a different error message

```c++
if (result) {
      ORTX_CXX_API_THROW("Failed to fetch cuda stream from context (status)", ORT_RUNTIME_EXCEPTION);
}
if (!cuda_stream_) {
      ORTX_CXX_API_THROW("Failed to fetch cuda stream from context (null pointer)", ORT_RUNTIME_EXCEPTION);
}
```

skyline75489:
Turns out `device_id_t` can in fact be `0`. I think this is ready now.

xadupre (approved):

PR #803, opened 2024-09-05
add(tutorials): exporting yolo world model

legendof-selda:
This allows us to export yolo world onnx model which can be later used in mobile inference.


legendof-selda:

tutorials/yolov8_world_e2e.py
```diff
@@ -0,0 +1,405 @@
...
+        By default the post processing will scale the bounding boxes and key points to the original image.
+    """
+    )
+    parser.add_argument("model", type=Path, help="The ONNX YOLOv8 World model.")
+    parser.add_argument(
+        "--classes",
+        type=Path,
+        default="classes.json",
+        help="JSON file containing list of classes that will be set as yolo world prompt.",
+    )
```
skottmckay, line 313: Where should a user get this file from in order to run the script?

legendof-selda: In yolo world you specify prompts to the model. So, it's just a list of text/classes that you will prompt the model for zero-shot detection. So, it's user specified. The user has to create the list of prompts.

skottmckay: Ideally a tutorial is self-contained with default inputs that allow running it end-to-end. We want it to make it as simple as possible for the user to see it working.

Is there an example classes.json that could be added that works with the default `--test_image` value of data/stormtroopers.jpg?

legendof-selda: I can add something. Should I create a file and commit it or have user pass in a list of values as classes and we can set default example classes for the argument?

legendof-selda: I have provided default values for classes as requested. Please review @skottmckay

natke (approved):

PR #804, opened 2024-09-06
Add C++ regex support for Llama3, Standard Library, and Custom Cases

sayanshaw24:
Currently, our regular expression pattern matching for BPE tokenization only supports the GPT2 regex.

This PR adds the following functionality:
1. Standard library regex support.
2. Llama3 regex support.
3. Custom regex implementation that will use appropriate regex given a model.

### DIFF
--- a/operators/tokenizer/bpe_kernels.cc
+++ b/operators/tokenizer/bpe_kernels.cc
@@ -276,7 +276,9 @@ std::vector<int64_t> KernelBpeTokenizer::Tokenize(ustring& input,
     }
 
     while (static_cast<int64_t>(res.size()) < max_length) {
-      auto [b, tok] = regcmp.GetNextToken();
+      // TODO: Change regex based on tokenizer type. Currently using GPT2 regex.
+      std::string regex_expr = "'s|'t|'re|'ve|'m|'ll|'d| ?\\p{L}+| ?\\p{N}+| ?[^\\s\\p{L}\\p{N}]+|\\s+(?!\\S)";
+      auto [b, tok] = regcmp.GetNextToken(regex_expr);
 
       if (!b) break;
 
--- a/operators/tokenizer/bpe_kernels.cc
+++ b/operators/tokenizer/bpe_kernels.cc
@@ -278,10 +278,10 @@ std::vector<int64_t> KernelBpeTokenizer::Tokenize(ustring& input,
     while (static_cast<int64_t>(res.size()) < max_length) {
       std::string regex_expr = "";
       if (ModelName() == kModel_Llama){
-        regex_expr = "(?i:'s|'t|'re|'ve|'m|'ll|'d)|[^\\r\\n\\p{L}\\p{N}]?\\p{L}+|\\p{N}{1,3}| ?[^\\s\\p{L}\\p{N}]+[\\r\\n]*|\\s*[\\r\\n]+|\\s+(?!\\S)|\\s+";
+        regex_expr = regcmp.LLAMA_REGEX_PATTERN_1;
       } else {
         // default to GPT2 regex
-        regex_expr = "'s|'t|'re|'ve|'m|'ll|'d| ?\\p{L}+| ?\\p{N}+| ?[^\\s\\p{L}\\p{N}]+|\\s+(?!\\S)";
+        regex_expr = regcmp.GPT2_REGEX_PATTERN;
       }
       auto [b, tok] = regcmp.GetNextToken(regex_expr);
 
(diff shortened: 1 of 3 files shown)


test/pp_api_test/test_tokenizer.cc
```diff
@@ -133,7 +133,7 @@ TEST(OrtxTokenizerTest, Phi3_S_Tokenizer) {
   EXPECT_NE(tokenizer, nullptr);
 
   std::vector<extTokenId_t> EXPECTED_IDS_0 = {2028, 374, 264, 1296, 13};
-  std::vector<std::string_view> input = {"This is a test.", "the second one",
+  std::vector<std::string_view> input = {"This is a test.", "Ich liebe München",
```
sayanshaw24, line 136: modified tests to include previously problematic char like `ü` that would cause "UTF-8 - Can't decode byte" error

wenbingl, line 136: I don't think the problem is related to this regular expression. Is it just an enhancement of unit test?

test/pp_api_test/test_tokenizer.cc
```diff
@@ -346,7 +346,7 @@ TEST(OrtxTokenizerStreamTest, Llama2Tokenizer) {
   // validate tokenizer is not null
   EXPECT_TRUE(tokenizer != nullptr);
 
-  std::vector<std::string_view> input = {"This is a test and the second one. "};
+  std::vector<std::string_view> input = {"This is a test and the second one is in German. Ich liebe München!"};
```
sayanshaw24, line 349: added test case for previously problematic char like ü that would cause the "UnicodeDecodeError: 'utf-8' codec can't decode byte" issue.

operators/tokenizer/bpe_utils.hpp
```diff
@@ -101,9 +103,9 @@ class TokenWithRegularExp {
     m_text = val;
   }
 
-  std::pair<bool, std::u32string_view> GetNextToken() {
+  std::pair<bool, std::u32string_view> GetNextToken(std::string & regex_expr) {
```
wenbingl, line 106: Can you write a unit test for this function, especially when there will be more patterns added?

operators/tokenizer/bpe_utils.hpp
```diff
@@ -114,12 +116,115 @@ class TokenWithRegularExp {
...
+      return u32str;
+  }
+
+  // Use the C++ standard library to handle regex pattern matching for compatible strings
+  std::u32string_view RegexMatchSTD(const std::u32string& regex) {
+    static std::u32string text(m_text);
+    std::wstring wstr = U2Wstring(text);
+    std::wstring wpattern = U2Wstring(regex);
+
+    std::wregex pattern(wpattern);
```
wenbingl, line 193: why do we have to use the wchar edition of std::regex? Since it doesn't handle unicode code, std::regex should be fine too, right?

operators/tokenizer/bpe_kernels.cc
```diff
@@ -276,7 +276,14 @@ std::vector<int64_t> KernelBpeTokenizer::Tokenize(ustring& input,
     }
 
     while (static_cast<int64_t>(res.size()) < max_length) {
-      auto [b, tok] = regcmp.GetNextToken();
+      std::string regex_expr = "";
+      if (ModelName() == kModel_Llama){
+        regex_expr = "(?i:'s|'t|'re|'ve|'m|'ll|'d)|[^\\r\\n\\p{L}\\p{N}]?\\p{L}+|\\p{N}{1,3}| ?[^\\s\\p{L}\\p{N}]+[\\r\\n]*|\\s*[\\r\\n]+|\\s+(?!\\S)|\\s+";
```
wenbingl, line 281: I saw this pattern appeared several time, can we define a const string variable for it?

operators/tokenizer/bpe_utils.hpp
sayanshaw24: even if the pattern is not in unicode, if the text to pattern match with is in unicode, we need to use wstrings, hence we use wregex and have to convert to and fro, because although the pattern does not have unicode chars, the input text does.

wenbingl: That's actually an interested topic. If a pattern is utf-8 encoded string, and it matches well on utf-8 input text as well if you take some time to think about it.

sayanshaw24: Yep! I already had some ideas for a related PR, can discuss with you in our weekly sync.

operators/tokenizer/bpe_utils.hpp
sayanshaw24: done! will add more after adding more functionality in a different PR so users can parse any regex, such as the one long one you sent me earlier.

operators/tokenizer/bpe_kernels.cc
sayanshaw24: good point, done!

wenbingl (approved):

PR #805, opened 2024-09-09
Don't disable vision operators in a catalyst build.

skottmckay:
We need the image decode/encode operators in the nuget package.


rmarinho:
Can I test this package? I was trying to access the azdo but seems I don t have permissions :) 

https://aiinfra.visualstudio.com/530acbc4-21bc-487d-8cd8-348ff451d2ff/_build/results?buildId=550669

edgchen1 (approved):

PR #808, opened 2024-09-13
Added support for native image decoding

skyline75489:
This added support for native image decoding on Windows & Apple platforms. This helps us remove libpng & libjpeg completely on these platforms, and in the meantime support more image formats thanks to OS vendors,


test/pp_api_test/test_imgcodec.cc
```diff
@@ -60,13 +71,40 @@ TEST(ImageDecoderTest, TestJpegDecoder) {
...
   auto status = image_decoder(jpeg_tensor, out_tensor);
-  ASSERT_TRUE(status.IsOk());
+  ASSERT_TRUE(status.IsOk()) << status.ToString();
 
   ASSERT_EQ(out_tensor.Shape(), std::vector<int64_t>({876, 1300, 3}));
   auto out_range = out_tensor.Data() + 0;
   ASSERT_EQ(std::vector<uint8_t>(out_range, out_range + 12),
             std::vector<uint8_t>({48, 14, 5, 48, 14, 5, 48, 14, 5, 48, 14, 5}));
 
+#if OCOS_ENABLE_VENDOR_IMAGE_CODECS
```
skyline75489, line 81: The implementation difference results in tiny difference in the decoding result

test/pp_api_test/test_imgcodec.cc
```diff
@@ -78,4 +116,55 @@ TEST(ImageDecoderTest, TestJpegDecoder) {
...
+            std::vector<uint8_t>({84, 68, 55, 86, 70, 55, 92, 77, 58, 101, 86, 65}));
+
+  out_range = out_tensor.Data() + 875 * 1300 * 3 + 1296 * 3;
+  ASSERT_EQ(std::vector<uint8_t>(out_range, out_range + 12),
+            std::vector<uint8_t>({208, 210, 197, 204, 206, 193, 198, 200, 187, 194, 196, 183}));
+#endif
+}
+
+#if OCOS_ENABLE_VENDOR_IMAGE_CODECS
+TEST(ImageDecoderTest, TestTiffDecoder) {
```
skyline75489, line 136: This is just an example. Many other formats can be supported as long as the OS provides such capability.

shared/api/image_decoder_darwin.hpp
```diff
@@ -0,0 +1,102 @@
...
+  CGImageRef        image = NULL;
+  CGImageSourceRef  imageSource;
+  CFDictionaryRef   imageSourceOptions = NULL;
+  CFStringRef       optionKeys[2];
+  CFTypeRef         optionValues[2];
+
+  optionKeys[0] = kCGImageSourceShouldCache;
+  optionValues[0] = (CFTypeRef)kCFBooleanFalse;
+  // Only Integer image data is currently supported
+  optionKeys[1] = kCGImageSourceShouldAllowFloat;
```
skyline75489, line 36: Will we add support for float numbe  image? I'm not sure about this.

test/data/processor/canoe.tif
skyline75489, line 1: From https://people.math.sc.edu/Burkardt/data/tif/tif.html

shared/api/image_decoder_win32.hpp
```diff
@@ -0,0 +1,142 @@
...
+  const int64_t encoded_image_data_len = input.NumberOfElement();
+
+  // check whether it's too small for a image
+  if (encoded_image_data_len < 8) {
+    return {kOrtxErrorInvalidArgument, "[ImageDecoder]: Invalid image data."};
+  }
+
+  OrtxStatus status{};
+
+  CoInitialize(NULL);
```
wenbingl, line 31: In this case, we change this operator type to be a struct which has an Init method to avoid CoInitialize on every compute call, Here is the example: https://github.com/microsoft/onnxruntime-extensions/blob/8d842d85e39aa36985cedf68f5d9e5dfef6f6d05/shared/api/image_transforms.hpp#L35C1-L36C1

skyline75489: sure

CMakeLists.txt
```diff
@@ -71,6 +71,7 @@ option(OCOS_ENABLE_BERT_TOKENIZER "Enable the BertTokenizer building" ON)
 option(OCOS_ENABLE_BLINGFIRE "Enable operators depending on the Blingfire library" ON)
 option(OCOS_ENABLE_MATH "Enable math tensor operators building" ON)
 option(OCOS_ENABLE_DLIB "Enable operators like Inverse depending on DLIB" ON)
+option(OCOS_ENABLE_VENDOR_IMAGE_CODECS "Enable and use vendor image codecs if supported over libpng & libjpeg" ON)
```
skyline75489, line 74: Not sure this should be enabled by default.

shared/api/image_decoder_win32.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+    pIComponentInfo->Release();
+    pSource->Release();
+    pIWICStream->Release();
+    pIDecoder->Release();
+
+    return status;
+  }
+
+  ~DecodeImage() {
```
skyline75489, line 149: Is destructor the correct place to release resources?

skyline75489:
To reviewers: I'll be out of office next week. Feel free to modify my branch directly.

skottmckay:
For pre/post processing we use the DecodeImage and EncodeImage operators. Is there a plan to add native encoding as well?

skyline75489:
@skottmckay I'm a bit confused by the coexistence of both `ort_extensions::KernelDecodeImage` (use opencv) and `ImageProcessor::DecodeImage` (use native codecs with this PR). Currently ort-genai is only using the image processor with decoding capability and that's why this PR only contains such changes.

I'm not sure if the vision operators currently powered by OpenCV can tolerate the tiny encoding & decoding difference after switching to native codecs. If so, I can probably take some time to unify these implementations.

CMakeLists.txt
```diff
@@ -71,6 +71,7 @@ option(OCOS_ENABLE_BERT_TOKENIZER "Enable the BertTokenizer building" ON)
...
 option(OCOS_ENABLE_OPENCV_CODECS "Enable cv2 and vision operators that require opencv imgcodecs." ON)
 option(OCOS_ENABLE_CV2 "Enable the operators in `operators/cv2`" ON)
```
skyline75489, line 76: If I understand this correctly, we can disable both `OPENCV_CODECS` and `CV2` in GenAI. Because we only use image processor related code. No custom operators injection is needed by GenAI. @baijumeswani Could you be kind enough to confirm this?

wenbingl: After you integrate the last ort-extensions main branch into GenAI, all of them opencv related flags should be turn off.  i.e. OCOS_ENABLE_OPENCV_CODECS OCOS_ENABLE_CV2 OCOS_ENABLE_VISION

shared/api/image_decoder_win32.hpp
```diff
@@ -0,0 +1,156 @@
...
+    _pIWICFactory->Release();
+    CoUninitialize();
+  }
+
+  private:
+    IWICImagingFactory* _pIWICFactory{NULL};
```
wenbingl, line 155: coding converntion: the underscore is at the end of the member name.

shared/api/image_decoder_win32.hpp
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    const int rowStride = uiWidth * sizeof(uint8_t) * channels;
+    hr = pSource->CopyPixels(NULL, rowStride, output.SizeInBytes(), decoded_image_data);
+
+    if (FAILED(hr)) {
+      return {kOrtxErrorInternal, "[ImageDecoder]: Failed when pConvertedFrame->CopyPixels."};
+    }
+
+    pIComponentInfo->Release();
```
wenbingl, line 141: these COM objects cannot be Release() if any error occurred above, like line:126. It seems the error handling in this function can be improved.

shared/api/image_decoder_darwin.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+    if (decoded_image_data == nullptr) {
+      return {kOrtxErrorInvalidArgument, "[ImageDecoder]: Failed to allocate memory for decoded image data."};
+    }
+
+    // CoreGraphics don't support 24BPP. We get 32BPP with alpha first and then extract the 24BPP data we need.
+    const size_t _32bpp_channel = 4;
+    const size_t _32bpp_bytesPerRow = width * _32bpp_channel;
+    const size_t _32bpp_bitmapByteCount = width * height * _32bpp_channel;
+    void* _32bpp_bitmapData = malloc(_32bpp_bitmapByteCount);
```
wenbingl, line 82: why malloc, instead of new? or can we use std::unique_ptr<..>?

wenbingl:
> @skottmckay I'm a bit confused by the coexistence of both `ort_extensions::KernelDecodeImage` (use opencv) and `ImageProcessor::DecodeImage` (use native codecs with this PR). Currently ort-genai is only using the image processor with decoding capability and that's why this PR only contains such changes.
> 
> I'm not sure if the vision operators currently powered by OpenCV can tolerate the tiny encoding & decoding difference after switching to native codecs. If so, I can probably take some time to unify these implementations.

yes, it would be nice that we can unify all of these kernels. The first step now is to remove the OpenCV dependencies for these operator under cv2/ and vision folders.

skottmckay:
> > @skottmckay I'm a bit confused by the coexistence of both `ort_extensions::KernelDecodeImage` (use opencv) and `ImageProcessor::DecodeImage` (use native codecs with this PR). Currently ort-genai is only using the image processor with decoding capability and that's why this PR only contains such changes.
> > I'm not sure if the vision operators currently powered by OpenCV can tolerate the tiny encoding & decoding difference after switching to native codecs. If so, I can probably take some time to unify these implementations.
> 
> yes, it would be nice that we can unify all of these kernels. The first step now is to remove the OpenCV dependencies for these operator under cv2/ and vision folders.

AFAIK we've only ever had accuracy issues from doing a Resize without anti-aliasing, which were resolved when the ONNX Resize added that support.

What's the worst diff you'd get from using the native codecs for decoding? An RGB value being off by one? Not sure anyone would notice that.

test/pp_api_test/test_imgcodec.cc
skyline75489: @skottmckay The test code here was modified to reflect the differences. Several RGB values are slightly off by 1 or 2.

skyline75489: @skottmckay I've also checked the difference in the decoded data by creating a BMP from it. Human eyes can't differentiate between OpenCV data and native codec data. It takes binary diff tool to spot the tiny differences.

shared/api/image_decoder_win32.hpp
skyline75489: sure

shared/api/image_decoder_win32.hpp
skyline75489: yeah, will fix it.

shared/api/image_decoder_darwin.hpp
skyline75489: I just need raw memory area and `free` it after copying it to `output`.

Smart pointer would save us the `free` part. Do you mean like this `std::make_unique<int>(_32bpp_bitmapByteCount)`?

If there benefit from using `new int[_32bpp_bitmapByteCount]` and `delete` instead of malloc/free?

shared/api/image_decoder_win32.hpp
skyline75489: Switch to `winrt::com_ptr` for automatic memory handling.

CMakeLists.txt
wenbingl: If this flag was ON, the build of JPEG/PNG library can be skipped, which can be done in a separated PR.

shared/api/image_decoder_win32.hpp
```diff
@@ -0,0 +1,149 @@
...
+#include <winrt/base.h>
+
+#include "op_def_struct.h"
+#include "ext_status.h"
+
+
+struct DecodeImage {
+  template <typename DictT>
+  OrtxStatus Init(const DictT& attrs) {
+    HRESULT hr = CoInitializeEx(NULL, COINIT_MULTITHREADED);
```
snnn, line 19: Good luck on the COM stuff.

wenbingl (dismissed):

shared/api/image_decoder_darwin.hpp
skyline75489: Fixed with `std::make_unique`

CMakeLists.txt
skyline75489: I decided to turn it off for now and only enable it with GenAI build. I think we can turn it on with future opencv removal work.

CMakeLists.txt
skyline75489: Yeah sure

shared/api/image_decoder_win32.hpp
skyline75489: manually releasing `pIWICFactory_` is not needed with `winrt::com_ptr`.

wenbingl (approved):

skyline75489:
Recently I've been validating this PR with the newly released [Phi-3.5-vision model](https://huggingface.co/microsoft/Phi-3.5-vision-instruct-onnx) and found that the small difference in image decoding doesn't change the model output. 

Removing OpenCV dramatically reduced the final binary size in [GenAI  0.5 release](https://github.com/microsoft/onnxruntime-genai/releases) without compromising the model output. I think now it's safe to say this is the correct direction.

FYI @wenbingl

PR #809, opened 2024-09-13
Fix regex prefast warning

sayanshaw24:
Fixes prefast warning due to "gsl::span or std::string_view created from a temporary will be invalid when the temporary is invalidated (gsl.view)" issue from RegexMatchSTD.

### DIFF
--- a/operators/tokenizer/bpe_utils.hpp
+++ b/operators/tokenizer/bpe_utils.hpp
@@ -198,7 +198,8 @@ class TokenWithRegularExp {
     std::wsmatch match;
 
     if (std::regex_search(wstr, match, pattern)) {
-        std::u32string_view token = std::u32string_view(W2Ustring(match.str()).data(), match.str().size());
+        std::u32string temp = W2Ustring(match.str());
+        std::u32string_view token = std::u32string_view(temp.data(), match.str().size());
         m_text = std::u32string(match.suffix().first, match.suffix().second); // Update text to the remaining part after the match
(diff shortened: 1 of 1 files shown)


wenbingl (approved):

operators/tokenizer/bpe_utils.hpp
```diff
@@ -198,7 +198,8 @@ class TokenWithRegularExp {
     std::wsmatch match;
 
     if (std::regex_search(wstr, match, pattern)) {
-        std::u32string_view token = std::u32string_view(W2Ustring(match.str()).data(), match.str().size());
+        std::u32string temp = W2Ustring(match.str());
+        std::u32string_view token = std::u32string_view(temp.data(), match.str().size());
```
snnn, line 202: But it is still a temporary object. Why do you return it out at line 204?

sayanshaw24: so `W2Ustring(match.str())` creates a temporary std::u32string object right, and the std::u32string_view token was created from its data, so once the temporary std::u32string is destroyed, the std::u32string_view becomes invalid.

hence i add another step and create the temp variable to ensure that the std::u32string object outlives the std::u32string_view

wenbingl: @sayanshaw24, @snnn is right, there is another issue of returning a view object refereeing to a temporary object.
image. You didn't fix the issue completely.

PR #811, opened 2024-09-18
Support the Unigram tokenizer kind from sentencepiece library

wenbingl:
1. Add a new type of tokenizer to support the Unigram tokenizer kind.
2. Refine bpe tokenizer loading code.


operators/tokenizer/bpe_kernels.cc
```diff
@@ -660,95 +659,79 @@ struct VectorEqual {
...
+bool JsonFastTokenizer::CheckForSpmModel(const json& tok_json) {
+  auto decoder_node = tok_json.find("decoder");
+  if (decoder_node != tok_json.end()) {
+    auto decoders_node = decoder_node->find("decoders");
+    if (decoders_node != decoder_node->end() && decoders_node->is_array()) {
+      for (const auto& step : *decoders_node) {
+        std::string type = step.value("type", "");
+        if (type == "Replace") {
+          std::string target = step.value("/pattern/String"_json_pointer, "");
+          if (target == "\xe2\x96\x81") {
```
sayanshaw24, line 700: can we also parse this value from somewhere (like the module file) rather than hardcoding it?

operators/tokenizer/bpe_kernels.cc
```diff
@@ -760,99 +743,145 @@ OrtxStatus JsonFastTokenizer::Load(const ort_extensions::bpe::TokenJsonConfig& c
...
+}
+
+OrtxStatus JsonFastTokenizer::LoadTikTokenBase64(const ort_extensions::bpe::TokenJsonConfig& config) {
+  std::string voc_file = config.GetVocabDataFile();
+  std::ifstream ifs = path(voc_file).open();
+  if (!ifs.is_open()) {
+    return OrtxStatus(kOrtxErrorInvalidFile, "Failed to open json file: " + voc_file);
+  }
+
+  // Following vocab and merges only used for tiktoken case but accessed outside scope below
```
sayanshaw24, line 781: this comment can now be removed since we have separate methods for loading tiktoken and regular json tokenizer files.

shared/api/tokenizer_impl.cc
```diff
@@ -1,41 +1,61 @@
...
+  }
 
-    // load the tokenizer from a config
-    status = tokenizer_->Load(*tok_config_);
-  } else {
-    // load the tokenizer from a config
-    status = tokenizer_->Load(*tok_config_);
+  auto vocab_file_path = ortx::path(tok_config_->GetVocabDataFile());
+  auto tokenizer = std::make_unique<JsonFastTokenizer>();
+  auto fx_load = vocab_file_path.extension() == ".json"?
```
sayanshaw24, line 48: perhaps we can also have a check here for if it's neither - so if `vocab_file_path.extension()` is ".json" then we use `JsonFastTokenizer::Load`, else if it's ".tiktoken", we use `JsonFastTokenizer::LoadTikTokenBase64`, otherwise we throw and error message saying something like "Tokenizer file type not supported! Please use .json or .tiktoken files to load tokenizer."

that way, if the user or anyone has errors in the future we immediately know what the root cause is.

operators/tokenizer/bpe_kernels.cc
wenbingl: yes, define a constant variable and removed the duplicates.

shared/api/tokenizer_impl.cc
wenbingl: the vocab file path should be checked in JsonConfig load which is the center place to do all checks. Add a comment here for calarifiction.

sayanshaw24 (approved):

PR #814, opened 2024-09-23
Fix the Unicode code discrepency on CLIP model

wenbingl:
Also clean up some unused files and C++ code warnings.


sayanshaw24 (approved):
looks great, thanks for adding this

PR #815, opened 2024-09-23
Allow the header and lib dir to be set externally

baijumeswani:


cmake/ext_ortlib.cmake
```diff
@@ -2,8 +2,14 @@ if(_ONNXRUNTIME_EMBEDDED)
   set(ONNXRUNTIME_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/../include/onnxruntime/core/session)
   set(ONNXRUNTIME_LIB_DIR "")
 elseif(ONNXRUNTIME_PKG_DIR)
```
wenbingl, line 4: If DONNXRUNTIME_PKG_DIR already there, why don't we directly use it?

baijumeswani: Because for ort-genai, the lib and include are in different folders.

baijumeswani: More context: ort-genai [downloads the nuget from ort-nightly](https://github.com/microsoft/onnxruntime-genai/blob/d67eda0cb2c3e4d52b623b31a8fd904e18896f01/cmake/ortlib.cmake#L18).

The nuget package has a unique structure that does not put the include and lib dir at the same level.

wenbingl: in that case, we use if (exists ...) to support nuget package folder structure here to simplify setting the folder?

wenbingl: and lib variable can be skipped since missing it only trigger a warning on building unit test, which should be fine for ort-genai.

baijumeswani:
Sorry, this is not needed after all:

https://github.com/microsoft/onnxruntime-extensions/blob/176c1d013864044bcc0747b908bdd32048669401/CMakeLists.txt#L297

I read the code incorrectly.

PR #817, opened 2024-09-25
support the merges array in tokenizer.json

wenbingl:
HF tokenizer made the change due to the space issue.
https://github.com/huggingface/tokenizers/commit/6a5fce9fa094d1514a498419f86ac0916e98ef8a


natke:
Should this PR have tests?

yufenglee (approved):

wenbingl:
> Should this PR have tests?



> Should this PR have tests?

Verified offline, but didn't find a public tokenizer.json for that, may check in some tests in the future with more tokenizer data file with this format available in HF.

PR #822, opened 2024-10-09
Add general regex support

sayanshaw24:
Add `RegexMatchGeneral` parser designed to handle virtually any regex case. This is in addition to the recently added `RegexMatchGPT2` and `RegexMatchLlama` parsers which handle the GPT2 and Llama regex based tokenizers.

This implementation also adds support for regex categories (such as `"\p{L}"`) which the C++ Standard Library methods (std::regex) do not support.


operators/tokenizer/bpe_utils.hpp
```diff
@@ -198,353 +299,435 @@ class TokenWithRegularExp {
...
-      return RegexMatchSTD(ustring(regex_expr));
+      std::u32string res = RegexMatchGeneral(regex_expr);
+      if (res != U"")
+        return res;
+      else
+        return RegexMatchSTD(ustring(regex_expr));
     } catch (const std::exception& /* ex */) {
       std::string part1 = "Regex '";
       std::string part2 = "' not supported!";
       throw std::runtime_error(part1 + regex_expr + part2);
```
wenbingl, line 705: can you replace all of throw with ORTX_CXX_API_THROW?

operators/tokenizer/bpe_utils.hpp
```diff
@@ -198,353 +299,435 @@ class TokenWithRegularExp {
...
-          break;
-      }
-      std::u32string_view res = m_text.substr(0, i);
-      m_text = m_text.substr(i);
-      return res;
-    }
+  // Function to wrap standalone \p{...} categories
+  std::string WrapStandaloneCategories(const std::string &input) {
+      std::string modified_input = input;
+      std::regex standalone_category_regex(R"(\\p\{[A-Za-z]+\})");
```
wenbingl, line 533: it's already RawString ( R"(....)"), why is there still double \ ahead of p?

operators/tokenizer/bpe_utils.hpp
```diff
@@ -189,7 +290,7 @@ class TokenWithRegularExp {
   }
 
   // Use the C++ standard library to handle regex pattern matching for compatible strings
-  std::u32string_view RegexMatchSTD(const std::u32string& regex) {
+  std::u32string RegexMatchSTD(const std::u32string& regex) {
```
wenbingl, line 293: Is there any test case which cover this function?

wenbingl (commented):
Did compare the perf diff between the original GPT2 implementation and the changes in this PR?

operators/tokenizer/bpe_utils.hpp
sayanshaw24: i tried without double backslash, seems RawString does not interpret backslashes as escape characters

operators/tokenizer/bpe_utils.hpp
sayanshaw24: done!

wenbingl (dismissed):

sayanshaw24:
> Did compare the perf diff between the original GPT2 implementation and the changes in this PR?

Yes! Both run in roughly ~ 0.002-0.003 ms for current test cases for all our tokenizers.

wenbingl (approved):

PR #823, opened 2024-10-10
Add the MLlama Imaging Processing Support

wenbingl:
Also validated the CLIP image processor


test/data/processor/clip_image.json
```diff
@@ -11,20 +11,14 @@
...
       {
         "operation": {
           "name": "resize",
           "type": "Resize",
           "attrs": {
             "interpolation": "CUBIC",
-            "width": 256,
-            "height": 256
+            "width": 224,
+            "height": 224
```
sayanshaw24, line 21: Is there some HF page(s) for reference/docs for better understanding/motivation here (not just for this but in general regards to this imaging support change)?

my original thought process for this snippet specifically would be that the `array = (array * 255).astype(np.uint8)` line below in order to convert to [0, 255] range would then also need to be modified (for [0, 223] accordingly) but maybe my understanding is wrong.

wenbingl: This is the image size, not the pixel value. The second paragraph you mentioned is called 'Rescale'. These basic operation was defined in Hugging face library here: https://github.com/huggingface/transformers/blob/a65a6ce7fece0ec44970b2e142729f33a98ac801/src/transformers/image_processing_utils.py#L53

sayanshaw24: Perfect, thanks!

sayanshaw24 (approved):

PR #826, opened 2024-10-14
CoreAudio

skyline75489:


CMakeLists.txt
```diff
@@ -71,11 +71,12 @@ option(OCOS_ENABLE_BERT_TOKENIZER "Enable the BertTokenizer building" ON)
 option(OCOS_ENABLE_BLINGFIRE "Enable operators depending on the Blingfire library" ON)
 option(OCOS_ENABLE_MATH "Enable math tensor operators building" ON)
 option(OCOS_ENABLE_DLIB "Enable operators like Inverse depending on DLIB" ON)
-option(OCOS_ENABLE_VENDOR_IMAGE_CODECS "Enable and use vendor image codecs if supported over libpng & libjpeg" OFF)
-option(OCOS_ENABLE_OPENCV_CODECS "Enable cv2 and vision operators that require opencv imgcodecs." ON)
-option(OCOS_ENABLE_CV2 "Enable the operators in `operators/cv2`" ON)
-option(OCOS_ENABLE_VISION "Enable the operators in `operators/vision`" ON)
+option(OCOS_ENABLE_VENDOR_IMAGE_CODECS "Enable and use vendor image codecs if supported over libpng & libjpeg" ON)
```
wenbingl, line 74: We need to remove the OpenCV dependency for these Ops before we can turn these flags off

skyline75489:
The difference between `CoreAudio` produced result and `dr_libs` is too large. The audio sounds the same, but the actual binary differs vastly. I cannot ship it with this kind of difference. Will close it for now.

PR #832, opened 2024-10-23
Fix regex prefast warnings

sayanshaw24:


operators/tokenizer/bpe_utils.hpp
```diff
@@ -512,87 +512,87 @@ class TokenWithRegularExp {
...
+      } else if (category == "Z") {
+        return ufal::unilib::unicode::Z;
+      } else if (category == "Zl") {
+        return ufal::unilib::unicode::Zl;
+      } else if (category == "Zp") {
+        return ufal::unilib::unicode::Zp;
+      } else if (category == "Zs") {
+        return ufal::unilib::unicode::Zs;
+      }
+    } catch (const std::exception& /* ex */) {
```
wenbingl, line 593: why do you need a try-catch?

} else {
ORTX_CXX_API_THROW("Invalid category string provided!", ORT_INVALID_ARGUMENT);
}
should be good enough

operators/tokenizer/bpe_utils.hpp
```diff
@@ -512,87 +512,87 @@ class TokenWithRegularExp {
 
   // Determine ufal::unilib::unicode regex category given string code.
   static ufal::unilib::unicode::category_t StringToCategory(const std::string & category = ""){
-    ufal::unilib::unicode::category_t cat;
-
     // Since C++ is not an interpreted language, we cannot simply convert the category to an object by typing
     // part of code into a string, so we manually parse it. Note that C++ also does not have switch-case statements.
```
edgchen1, line 516: "Note that C++ also does not have switch-case statements."

nit: this is not true. see https://en.cppreference.com/w/cpp/language/switch. it just doesn't support string comparison.

sayanshaw24: oh ok good to know, cheers!

PR #835, opened 2024-10-29
Upgrade versions in CI matrix and fix CI issue

sayanshaw24:


wenbingl:
I don't think a lot of packages are available for Python 3.13, Can we just shift ORT version only?

wenbingl (commented):
We cannot drop python 3.8 support. and should only shift ort version from [1.13, 1.17] to [1.15, 1.19]

edgchen1:
> We cannot drop python 3.8 support. and should only shift ort version from [1.13, 1.17] to [1.15, 1.19]

FYI, Python 3.8 is EOL now. https://devguide.python.org/versions/

do we still need to support it?

wenbingl:
> > We cannot drop python 3.8 support. and should only shift ort version from [1.13, 1.17] to [1.15, 1.19]
> 
> FYI, Python 3.8 is EOL now. https://devguide.python.org/versions/
> 
> do we still need to support it?

I just happened to see the HF's transformers Python package 4.46 yanked because it works with Python 3.8 https://pypi.org/project/transformers/4.46.0/. At least, it should be dropped after Python 3.13 was fully supported.

wenbingl:
> > > We cannot drop python 3.8 support. and should only shift ort version from [1.13, 1.17] to [1.15, 1.19]
> > 
> > 
> > FYI, Python 3.8 is EOL now. https://devguide.python.org/versions/
> > do we still need to support it?
> 
> I just happened to see the HF's transformers Python package 4.46 yanked because it works with Python 3.8 https://pypi.org/project/transformers/4.46.0/. At least, it should be dropped after Python 3.13 was fully supported.

it seems ORT dropped, we can drop it too.

.pipelines/ci.yml
```diff
@@ -445,26 +445,26 @@ stages:
...
+          ort.version: '1.16.3'
+        py39-1151:
           python.version: '3.9'
-          torch.version: 'torch==1.13.1 torchvision==0.14.1 torchaudio==0.13.1 --index-url https://download.pytorch.org/whl/cpu'
-          ort.version: '1.14.1'
-        py38-1131:
-          python.version: '3.8'
-          torch.version: 'torch==1.12.1 torchvision==0.13.1 torchaudio==0.12.1 -f https://download.pytorch.org/whl/torch_stable.html'
-          ort.version: '1.13.1'
+          torch.version: 'torch torchvision torchaudio'
```
wenbingl, line 466: can you keep the torch version for some backward compatibility?

wenbingl (commented):
Can you also the build configuration for MacOS and Linux too?

.pipelines/ci.yml
sayanshaw24: done!

wenbingl (approved):

PR #843, opened 2024-11-13
fix the build for mobile packaging

wenbingl:


tools/gen_selectedops.py
```diff
@@ -74,7 +74,9 @@
     ],
     "OCOS_ENABLE_DLIB": [
         "Inverse",
-        "StftNorm"
+        "StftNorm",
+        "DecodeImage",
+        "EncodeImage"
```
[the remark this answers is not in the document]
wenbingl: Both OCOS_ENABLE_VISION and OCOS_ENABLE_DLIB are required because libjpeg and libpng C source files are in _dep/dlib folder.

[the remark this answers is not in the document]
wenbingl: BTW, enable_dlib doesn't introduce more footprint since we compile the required C++ files directly without any CMake dependencies. And if we could have native implementations for encode_image.cc, like decode_image did by @skyline75489, then libjpeg and etc can fully removed from iOS build.

.pipelines/ci.yml
```diff
@@ -851,10 +851,9 @@ stages:
         architecture: 'x64'
       displayName: "Use Python 3.12"
 
-    # iOS xcframework build doesn't work with CMake 3.25.1, pin to 3.25.0
     - script: |
-        python -m pip install cmake==3.25.0
-      displayName: "Install CMake 3.25.0"
```
wenbingl, line 857: I think it was caused by openCV stuff

CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 # Licensed under the MIT License.
 
 # Minimum CMake required
-cmake_minimum_required(VERSION 3.25)
+cmake_minimum_required(VERSION 3.28)
```
wenbingl, line 5: starting from 3.28, cmake supports EXCLUDE_FROM_ALL which already used in some project. A lower version cmake actually doesn't report any issue, which confused us to see some inconsistent build processes without any useful log.

.pipelines/ci.yml
edgchen1: sorry, I don't remember. if it works with a later version then that's good.

tools/gen_selectedops.py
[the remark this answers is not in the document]
skyline75489: If we have enough test case covering the encoding scenarios, I think it should be possible to add native encoding capability on Windows & Apple platforms, just as their decoding counterpart in #808

CMakeLists.txt
```diff
@@ -334,6 +334,20 @@ endif()
...
+  set(DLIB_ISO_CPP_ONLY ON CACHE INTERNAL "")
+  set(DLIB_NO_GUI_SUPPORT ON CACHE INTERNAL "")
+  set(DLIB_USE_CUDA OFF CACHE INTERNAL "")
+  set(DLIB_USE_LAPACK OFF CACHE INTERNAL "")
+  set(DLIB_USE_BLAS OFF CACHE INTERNAL "")
+  include(dlib)
+
+  # Ideally, dlib should be included as
+  # file(GLOB TARGET_SRC_DLIB "${dlib_SOURCE_DIR}/dlib/all/source.cpp")
+  # To avoid the unintentional using some unwanted component, only include
```
edgchen1, line 347: is this comment incomplete? "only include..."

.pipelines/ci.yml
```diff
@@ -851,10 +851,9 @@ stages:
...
+        python -m pip install cmake
```
edgchen1, line 855: is it ok to always get the latest? things could break unexpectedly. also, it might be hard to reproduce a build from an older commit if the cmake version used changes.

CMakeLists.txt
wenbingl: to include the file in the next line, which looks like a tricky. I'll update it to be more clear.

.pipelines/ci.yml
wenbingl: Agree the points you mentioned here. But ort-extensions doesn't depend on any complex libraries anymore, so ideally latest cmake should be totally OK, unless there is very uncommon situation. 
Also, the fixed cmake version may requires the version update in the future as well. Furthermore, if some new cmake version causes some trouble, we can know it at the earliest time so that we can fix it quickly to help ort-extensions users which build it from source.

So, as a trade-off, keeping it be latest may be more suitable for ort-extensions project.

.pipelines/ci.yml
wenbingl: validated it. I think openCV was the reason.

tools/gen_selectedops.py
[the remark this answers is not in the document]
wenbingl: > If we have enough test case covering the encoding scenarios, I think it should be possible to add native encoding capability on Windows & Apple platforms, just as their decoding counterpart in #808

There are some Python test cases covered this operator, like this one: https://github.com/microsoft/onnxruntime-extensions/blob/5104bb989745dfb12b384add07d6886b504c25cd/test/test_tools_add_pre_post_processing_to_model.py#L248. We can add some native C++ test to replace this one: https://github.com/microsoft/onnxruntime-extensions/blob/main/test/shared_test/test_ortops_vision.cc too.

edgchen1 (approved):

PR #844, opened 2024-11-14
fix the win32 macro usage

wenbingl:
_"WIN32 is a name that you could use and even define in your own code and so might clash with Microsoft's usage. _WIN32 is a name that is reserved for the implementor (in this case Microsoft) because it begins with an underscore and an uppercase letter - you are not allowed to define reserved names in your own code, so there can be no clash."_


snnn (approved):

snnn:
I often use the following website as the reference:

https://sourceforge.net/p/predef/wiki/OperatingSystems/

It is very comprehensive

PR #848, opened 2024-11-21
Added support for native image encoding

skyline75489:
Follow-up of #808


operators/vision/image_encoder_darwin.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+struct EncodeImage {
+  OrtxStatus OnInit() {
+    float compression = 0.95;
+
+    CFStringRef optionKeys[2];
+    CFTypeRef optionValues[2];
+    optionKeys[0] = kCGImagePropertyHasAlpha;
+    optionValues[0] = (CFTypeRef)kCFBooleanFalse;
+    optionKeys[1] = kCGImageDestinationLossyCompressionQuality;
```
skyline75489, line 22: `FilterOption` for PNG is not exposed on Darwin. Only `LossyCompressionQuality` is available.

operators/vision/image_encoder_win32.hpp
```diff
@@ -0,0 +1,225 @@
...
+                   size_t* outsize) const {
+     const uint8_t* encode_source_data = source_data;
+     winrt::com_ptr<IStream> pOutputStream;
+     winrt::com_ptr<IWICBitmapEncoder> pIEncoder;
+     winrt::com_ptr<IWICBitmapFrameEncode> pBitmapFrame;
+     winrt::com_ptr<IPropertyBag2> pPropertyBag;
+
+     HRESULT hr = ::CreateStreamOnHGlobal(NULL, FALSE, pOutputStream.put());
+     if (FAILED(hr)) {
+       ORTX_CXX_API_THROW(errorWithHr_("Failed when CreateStreamOnHGlobal.", hr).Message(), ORT_RUNTIME_EXCEPTION);
```
skyline75489, line 114: In the existing `libjpeg` and `libpng` encoder, `ORTX_CXX_API_THROW` was used. So I copied this behaviour. Should this be a `OrtxStatus` instead (like in decoders)? Should this use `ORT_RUNTIME_EXCEPTION`? I'm not sure

operators/vision/encode_image.hpp
```diff
@@ -18,12 +30,58 @@ struct KernelEncodeImage : BaseKernel {
     }
 
     extension_ = std::string(".") + format;
+    encoder_.OnInit();
   }
 
-  void Compute(const ortc::Tensor<uint8_t>& input_bgr,
-               ortc::Tensor<uint8_t>& output) const;
+  void Compute(const ortc::Tensor<uint8_t>& input_bgr, ortc::Tensor<uint8_t>& output) const{
```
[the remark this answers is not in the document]
wenbingl: Yes. but we can add an attribute for this class to support both.

operators/vision/image_encoder_win32.hpp
```diff
@@ -0,0 +1,225 @@
...
+    // Create the COM imaging factory
+    hr = CoCreateInstance(CLSID_WICImagingFactory, NULL, CLSCTX_INPROC_SERVER, IID_PPV_ARGS(&pIWICFactory_));
+    if (FAILED(hr)) {
+      return errorWithHr_("Failed to create pIWICFactory.", hr);
+    }
+
+    return {};
+  }
+
+  bool JpgSupportsBgr() const{ return true; }
```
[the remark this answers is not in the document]
skyline75489: Yes. But I'm trying to prevent unnecessary conversion as much as possible.

CMakeLists.txt
```diff
@@ -666,7 +666,7 @@ if(OCOS_ENABLE_VISION)
     elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin" OR CMAKE_SYSTEM_NAME STREQUAL "iOS")
       # Use ImageIO on Apple platforms
       set(_DEFAULT_CODEC_ENABLE OFF)
-      target_link_libraries(ocos_operators PRIVATE "-framework CoreFoundation" "-framework CoreGraphics" "-framework ImageIO")
+      target_link_libraries(ocos_operators PRIVATE "-framework CoreFoundation" "-framework CoreGraphics" "-framework CoreServices" "-framework ImageIO")
```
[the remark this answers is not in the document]
skyline75489: Yes. This is needed. Otherwise linker complains about symbol not founc.

operators/vision/image_encoder_win32.hpp
skyline75489: How can we get if the format is png or jpeg, without using `BaseKernel`? Is it still possible?

operators/vision/encode_image.hpp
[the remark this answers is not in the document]
skyline75489: I think an attribute is much needed for clarification. Because from the shape, we can't reliably determine if a bunch of bytes is RGB or BGR. They look exactly the same from the callers' point of view..

wenbingl:
@skyline75489 , is there any update for this PR?

operators/vision/encode_image.hpp
```diff
@@ -8,23 +8,148 @@
...
+      conversion_buf = std::make_unique<uint8_t[]>(height * width * color_space);
+      auto cvt_data = conversion_buf.get();
+      for (int32_t y = 0; y < height; ++y) {
+        for (int32_t x = 0; x < width; ++x) {
+          cvt_data[(y * width + x) * color_space + 0] = input_data_ptr[(y * width + x) * color_space + 2];
+          cvt_data[(y * width + x) * color_space + 1] = input_data_ptr[(y * width + x) * color_space + 1];
+          cvt_data[(y * width + x) * color_space + 2] = input_data_ptr[(y * width + x) * color_space + 0];
+        }
+      }
+      rgb_data = cvt_data;
```
skyline75489, line 128: Is it correct `rgb_data` is not really used? I think if `fx_supported_bgr` if false and `is_bgr_` is true, the `rgb_data` should be used instead of `input_data_ptr`, right?

wenbingl: you are right, there is a missing change on calling EncodeJPG/PNG function.

PR #849, opened 2024-11-21
Add OCOS_ENABLE_VENDOR_IMAGE_CODECS compile definition, even if VISION is disabled

stwykd:
- As per PR#808, I shouldn't need `libpng` and `libjpeg` on Windows and Apple, if `OCOS_ENABLE_VENDOR_IMAGE_CODECS` is enabled
- The CMakeLists.txt only adds the `OCOS_ENABLE_VENDOR_IMAGE_CODECS` compile definition when `OCOS_ENABLE_VISION` is set, assuming that decoders are not used if `VISION` is not enabled. That's not the case because `shared\api\image_processor.cc` is added for `OCOS_ENABLE_C_API` (`CMakeLists.txt:761`) and `OCOS_BUILD_SHARED_LIB=OFF` (`CMakeLists.txt:873`)

In this change:
- Handling for `-DOCOS_ENABLE_VENDOR_IMAGE_CODECS` is performed even if `VISION` is not enabled.
- I removed `_DEFAULT_CODEC_ENABLE`, and `if(_DEFAULT_CODEC_ENABLE)` as it is redundant given it's preceded by `set(_DEFAULT_CODEC_ENABLE ON)`

### DIFF
--- a/tools/ci_build/github/azure-pipeline/templates/windows-build-stage.yml
+++ b/tools/ci_build/github/azure-pipeline/templates/windows-build-stage.yml
@@ -9,6 +9,7 @@ stages:
   dependsOn: []
   jobs:
   - job: main
+    timeoutInMinutes: 120
     pool:
       name: onnxruntime-Win-CPU-2022
       os: windows


skyline75489:
Would this conflict with #848 ?

wenbingl:
> * As per PR#808, I shouldn't need `libpng` and `libjpeg` on Windows and Apple, if `OCOS_ENABLE_VENDOR_IMAGE_CODECS` is enabled
> * The CMakeLists.txt only adds the `OCOS_ENABLE_VENDOR_IMAGE_CODECS` compile definition when `OCOS_ENABLE_VISION` is set, assuming that decoders are not used if `VISION` is not enabled. That's not the case because `shared\api\image_processor.cc` is added for `OCOS_ENABLE_C_API` (`CMakeLists.txt:761`) and `OCOS_BUILD_SHARED_LIB=OFF` (`CMakeLists.txt:873`)
> 
> In this change:
> 
> * Handling for `-DOCOS_ENABLE_VENDOR_IMAGE_CODECS` is performed even if `VISION` is not enabled.
> * I removed `_DEFAULT_CODEC_ENABLE`, and `if(_DEFAULT_CODEC_ENABLE)` as it is redundant given it's preceded by `set(_DEFAULT_CODEC_ENABLE ON)`

_DEFAULT_CODEC_ENABLE was added for these platforms which doesn't have native APIs for image codecs so that we don't need explicitly turn this flag ON or OFF for pipelines on different platforms. If this PR can resolve issue. we can merge this PR to unblock you since this flag was OFF now so it might not break any pipeline.

Moreover, OCOS_ENABLE_VENDOR_IMAGE_CODECS cannot be ON by default is due to missing encoder support, with #848, we will turn on this flag.

skyline75489:
@stwykd #855 will be the attempt to turn on `OCOS_ENABLE_VENDOR_IMAGE_CODECS ` by default, in case you're interested.

PR #856, opened 2024-12-10
Fix image encoder status

skyline75489:
Follow-up of #848. This fixes a native C++ crash.


operators/vision/image_encoder_darwin.hpp
```diff
@@ -95,6 +95,7 @@ struct EncodeImage {
     CFDataGetBytes(result, CFRangeMake(0, size), *outbuffer);
     CFRelease(result);
     *outsize = size;
+    return {};
```
skyline75489, line 98: Missing the return results in `EXC_BREAKPOINT (code=1, subcode=0x10000e698)` on ARM64.

wenbingl (approved):

PR #857, opened 2024-12-10
Pipeline - Publish whl Files to ORT-Nightly Feed

idiskyle:
1 Publish whl Files to ORT-Nightly Feed:

2 For these 3 pipelines, official, linux wheel and macro wheel, upgrade Python version to 9,10,11,12,13 instead of 8,9,10,11,12.

### DIFF
--- a/tools/ci_build/github/azure-pipeline/templates/windows-build-stage.yml
+++ b/tools/ci_build/github/azure-pipeline/templates/windows-build-stage.yml
@@ -20,7 +20,7 @@ stages:
       os: windows
 
     variables:
-      CIBW_BUILD: cp3{9,10,11,12,13}-*amd64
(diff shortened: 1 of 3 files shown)


tools/ci_build/github/azure-pipeline/templates/windows-build-stage.yml
```diff
@@ -15,7 +20,7 @@ stages:
       os: windows
 
     variables:
-      CIBW_BUILD: cp3{8,9,10,11,12}-*amd64
+      CIBW_BUILD: cp3{9,10,11,12,13}-*amd64
```
snnn, line 23: You may remove 9 as well, because ORT has done the same.

snnn (approved):

PR #858, opened 2024-12-11
update python api for blob support

sayanshaw24:


onnxruntime_extensions/cvt.py
```diff
@@ -47,6 +204,9 @@ def gen_processing_models(processor: Union[str, object],
         Keyword arguments for generating the post-processing model
     opset: int
         the target opset version of the model
+    cached: bool
```
wenbingl, line 207: can a better name for this argument? 'cached' look like being irrelevant to what we are doing here.

wenbingl:
why are the CI pipelines skipped?

sayanshaw24:
> why are the CI pipelines skipped?

not sure - let me try to manually run it.

sayanshaw24:
> pipeline triggers exclude this branch

ok seems pipeline triggers exclude this branch so it cannot be run from it (we likely need a PR to main branch, if we check in your hfjst branch in soon in #859, I can pull the changes and make a separate PR to main)

PR #863, opened 2024-12-19
Add regex loading from tokenizer.json and code refinement

wenbingl:
1. using regex pattern if it exist in tokenizer.json
2. add tokenizer test case like AMD/OLMa through pp_api


operators/tokenizer/bpe_tokenizer_model.hpp
```diff
@@ -44,6 +44,58 @@ class BpeModel {
...
+      return {};
+    }
+
+    auto iter_type = node_pre_tokenizer->find("type");
+    if (iter_type == node_pre_tokenizer->end() || iter_type->is_null()) {
+      return {};
+    }
+
+    if (iter_type->get<std::string>() != "Sequence") {
+      return {kOrtxErrorNotImplemented, "Unsupported pretokenizer type!"};
```
sayanshaw24, line 59: something in the test seems off - i saw in `test_pp_api.py` you use the `"amd/AMD-OLMo-1B-SFT-DPO"` model, and if you look in the `tokenizer.json` for it, `pre_tokenizer` type is "ByteLevel", not "Sequence"; so it would fail here (but in the CI it is passing). So - we should add "ByteLevel" to the supported types as well here, but also first identify why it is not failing the test currently, perhaps the type is not being extracted right.

operators/tokenizer/bpe_tokenizer_model.hpp
```diff
@@ -44,6 +44,58 @@ class BpeModel {
...
+
+
+      auto pre_type = iter_type->get<std::string>();
+      if (pre_type == "Split") {
+        auto iter_pattern = node.find("pattern");
+        if (iter_pattern == node.end() || iter_pattern->is_null()) {
+          continue;
+        }
+
+        auto regex_str = iter_pattern->find("Regex");
```
sayanshaw24, line 82: "regex" should be lower case here no?

sayanshaw24:
Can we add an example to test the regex string option as well? HF "amd/AMD-OLMo-1B-SFT-DPO" has a "pre_tokenizer" value with type ""ByteLevel" and "use_regex": true, but maybe we can find one with a "regex" string that we can then use to demonstrate the custom regex functionality.

operators/tokenizer/bpe_tokenizer_model.hpp
wenbingl: it looks the latest schema is different from what OLMo was created. I supported both of them now.

operators/tokenizer/bpe_tokenizer_model.hpp
wenbingl: May you share a link to that example?

sayanshaw24: hmm can't seem to find the one i saw at the moment, but perhaps it was a custom tokenizer and not one of the popular tokenizers? nvm then we can keep as is for now

sayanshaw24 (approved):

PR #874, opened 2025-01-11
a general regex match algorithm to fix all related issues

wenbingl:


operators/tokenizer/bpe_utils.hpp
```diff
@@ -99,189 +93,169 @@ class SpecialTokenMap {
...
             ((m_text[1] == U'l') && (m_text[2] == U'l'))) {
-          std::u32string res = m_text.substr(0, 3);
+          std::u32string_view res = m_text.substr(0, 3);
           m_text = m_text.substr(3);
           return res;
         }
       }
     }
 
-    // Case 2: Language Character Class (8th Alternative)
```
sayanshaw24, line 163: can we keep comments explaining different regex cases for better understanding when looking at/editing this code in the future?

sayanshaw24 (approved):

operators/tokenizer/bpe_utils.hpp
wenbingl: This information can easily be found https://regex101.com/ and they are standard regex pattern, so I prefer not to elaborate them to every functions

PR #876, opened 2025-01-15
Support fast tokenizer type in JSON tokenizer

wenbingl:


test/test_pp_api.py
```diff
@@ -140,6 +140,15 @@ def test_Qwen_QVQ_tokenizer(self):
         ortx_inputs = tokenizer.tokenize(test_sentence)
         np.testing.assert_array_equal(ortx_inputs, inputs)
 
+    def test_Phi4_tokenizer(self):
+        model_id = "/g/phi-x-12202024"
```
wenbingl, line 144: it should be another PR but keep it here so you can use it verify the phi-4 tokenization new regex pattern. @sayanshaw24

sayanshaw24 (approved):

PR #877, opened 2025-01-16
Add phi-4 regex support

sayanshaw24:


operators/tokenizer/bpe_utils.hpp
```diff
@@ -97,6 +97,8 @@ class PreTokenizerWithRegEx {
       R"('s|'t|'re|'ve|'m|'ll|'d| ?\p{L}+| ?\p{N}+| ?[^\s\p{L}\p{N}]+|\s+(?!\S)|\s+)";
   static constexpr const char LLAMA_REGEX_PATTERN[] =
       R"((?i:'s|'t|'re|'ve|'m|'ll|'d)|[^\r\n\p{L}\p{N}]?\p{L}+|\p{N}{1,3}| ?[^\s\p{L}\p{N}]+[\r\n]*|\s*[\r\n]+|\s+(?!\S)|\s+)";
+  static constexpr const char PHI4_REGEX_PATTERN[] =
```
wenbingl, line 100: No need to put pattern here, which can be loaded from tokenizer.json

operators/tokenizer/bpe_tokenizer_model.hpp
```diff
@@ -417,6 +417,8 @@ class BpeModel {
 
     if (model_name == "Llama") {
       return bpe::PreTokenizerWithRegEx::LLAMA_REGEX_PATTERN;
+    } else if (model_name == "Phi") {
```
wenbingl: No need either, given new patterns are always stored in tokenizer.json file

operators/tokenizer/bpe_utils.hpp
```diff
@@ -359,6 +361,114 @@ class PreTokenizerWithRegEx {
...
+  std::u32string_view Match_PHI4_Pattern_1() {
+    size_t i = 0;
+
+    // [^\r\n\p{L}\p{N}]?
+    if (!IsRN(m_text[i]) && !IsN(m_text[i]) && !IsL(m_text[i])) {
+      i++;
+    }
+
+    // [\p{Lu}\p{Lt}\p{Lm}\p{Lo}\p{M}]*
+    std::vector<ufal::unilib::unicode::category_t> categories1 = {ufal::unilib::unicode::Lu,
```
wenbingl, line 372: std::set should be a little fast.

operators/tokenizer/bpe_utils.hpp
```diff
@@ -359,6 +361,114 @@ class PreTokenizerWithRegEx {
...
+                                                                 ufal::unilib::unicode::Lm,
+                                                                 ufal::unilib::unicode::Lo,
+                                                                 ufal::unilib::unicode::M};
+    while (std::find(categories1.begin(), categories1.end(), ufal::unilib::unicode::category(m_text[i])) != categories1.end()){
+      i++;
+    }
+
+    // [\p{Ll}\p{Lm}\p{Lo}\p{M}]+
+    size_t j = i;
+    std::vector<ufal::unilib::unicode::category_t> categories2 = {ufal::unilib::unicode::Ll,
```
wenbingl, line 383: std::set, and can we make this check to be a IsLl.. function?

sayanshaw24: yeah, figured we wanted to do simply hardcode it for perf since the previous `StringToCategory` and `IsCategory(char32_t ch, std::vector<std::string> categories)` was removed in your last PR; will add them back.

wenbingl: Can we just make if/else to compare the type here? the other approach looks like over-design to me.
And extracting them into a function is to dedup the same code since the same sub-pattern appears twice.

sayanshaw24: sounds good!

operators/tokenizer/bpe_utils.hpp
```diff
@@ -359,6 +359,112 @@ class PreTokenizerWithRegEx {
     return {};
   }
 
+  void CategoryMatch(size_t& index, std::set<ufal::unilib::unicode::category_t>& categories){
+    while (categories.find(ufal::unilib::unicode::category(m_text[index])) != categories.end()){
+      index++;
```
wenbingl, line 364: Need check if index is greater than m_text.size()

sayanshaw24: right, good catch

wenbingl (approved):

PR #878, opened 2025-01-27
Add preprocessing supports Phi-O model

wenbingl:


test/pp_api_test/test_processor.cc
```diff
@@ -88,36 +88,82 @@ TEST(ProcessorTest, TestMLlamaImageProcessing) {
...
+  ASSERT_EQ(err, kOrtxOK);
+  err = OrtxGetTensorData(tensor.get(), reinterpret_cast<const void**>(&data), &shape, &num_dims);
+  ASSERT_EQ(err, kOrtxOK);
+  ASSERT_EQ(std::vector<int64_t>(shape, shape + num_dims), std::vector<int64_t>({3, 10, 32, 32}));
+
+  err = OrtxTensorResultGetAt(result.get(), 3, tensor.ToBeAssigned());
+  ASSERT_EQ(err, kOrtxOK);
+  err = OrtxGetTensorData(tensor.get(), reinterpret_cast<const void**>(&data), &shape, &num_dims);
+  ASSERT_EQ(err, kOrtxOK);
+  ASSERT_EQ(std::vector<int64_t>(shape, shape + num_dims), std::vector<int64_t>({3, 1}));
```
sayanshaw24, line 168: it seems for image (like audio) we don't compare result values, shape and status - perhaps we can have some local tests (that CI does not automatically run and need to be run manually), that use preexisting test/data files and also compare values (for example like comparing difference with atol, rtol like we did for audio locally)

but this can be done in a future PR - just a thought

sayanshaw24 (approved):

test/pp_api_test/test_processor.cc
wenbingl: Sure, we should add a unit test for the output value check, which can be done in another PR.

PR #881, opened 2025-02-01
Add unit test for Phi-O audio output

sayanshaw24:


test/test_phi-4_audio.py
wenbingl, line 1: can you write a C++ unit test to load the expected_output.txt and compare it with API output?

test/pp_api_test/test_feature_extraction.cc
```diff
@@ -99,3 +99,78 @@ TEST(ExtractorTest, TestPhi4AudioFeatureExtraction8k) {
...
+  err = OrtxGetTensorData(tensor.get(), reinterpret_cast<const void**>(&data), &shape, &num_dims);
+  ASSERT_EQ(err, kOrtxOK);
+  ASSERT_EQ(std::vector<int64_t>(shape, shape + num_dims), std::vector<int64_t>({1, 1332, 80}));
+
+  // Dimensions
+  const size_t num_rows = shape[1];
+  const size_t num_columns = shape[2];
+
+  // Read the expected output from the file
+  std::filesystem::path expected_file_path = "data/models/phi-4/expected_output.txt";
```
wenbingl, line 131: "expect_audio_embed_ouptut"?

test/pp_api_test/test_feature_extraction.cc
```diff
@@ -99,3 +99,78 @@ TEST(ExtractorTest, TestPhi4AudioFeatureExtraction8k) {
...
+  std::ifstream infile(expected_file_path);
+
+  if (!infile.is_open()) {
```
wenbingl, line 134: this is a unit test, so ASSERT the error is a common way to handle the failure.

test/pp_api_test/test_feature_extraction.cc
```diff
@@ -99,3 +99,78 @@ TEST(ExtractorTest, TestPhi4AudioFeatureExtraction8k) {
...
+      row_idx++;
+  }
+
+  infile.close();
+
+  // Calculate the mismatch percentage
+  float mismatch_percentage = static_cast<float>(num_mismatched) / total_elements;
+
+  std::cout << "Mismatch percentage: " << mismatch_percentage * 100 << "%" << std::endl;
+  ASSERT_LT(mismatch_percentage, 0.02) << "Mismatch percentage exceeds 2% threshold!";
```
wenbingl, line 175: Where does this percentage come from?

sayanshaw24: Same as whisper python unit test here: https://github.com/microsoft/onnxruntime-extensions/blob/main/test/test_whisper.py#L61

HF has 1.5% mismatch; our mismatch percentage in this example is around 0.07%.

test/pp_api_test/test_feature_extraction.cc
wenbingl: :) I mean the file name, since there are some other output file.

wenbingl (approved):

PR #888, opened 2025-02-07
Update op_def_struct.h to fix memory leak

yunmengxie:
This commit removes the .release() calls, ensuring that CreateLiteCustomOpV2 now returns std::shared_ptr, maintaining proper smart pointer ownership and preventing memory corruption.

### DIFF
--- a/cmake/externals/json.cmake
+++ b/cmake/externals/json.cmake
@@ -1,6 +1,6 @@
 FetchContent_Declare(nlohmann_json
-  URL       https://codeload.github.com/nlohmann/json/zip/refs/tags/v3.10.5
-  URL_HASH  SHA1=f257f8dc27c5b8c085dc887b40cddd18ae1f725c
+  URL       https://codeload.github.com/nlohmann/json/zip/refs/tags/v3.11.3
+  URL_HASH  SHA1=5e88795165cc8590138d1f47ce94ee567b85b4d6
   SOURCE_SUBDIR not_set
   )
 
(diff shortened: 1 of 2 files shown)


include/op_def_struct.h
```diff
@@ -286,18 +286,18 @@ struct OrtLiteCustomStructV2 : public OrtLiteCustomOp {
 };
 
 template <typename RType, typename... Args>
-OrtLiteCustomOp* CreateLiteCustomOpV2(const char* op_name,
+std::shared_ptr<OrtLiteCustomOp> CreateLiteCustomOpV2(const char* op_name,
```
wenbingl, line 289: regarding to the memory leak, did you directly call this function, instead of this macro? https://github.com/microsoft/onnxruntime-extensions/blob/48176ac7dc9412296a3f23ff4ea52fe696b08fba/include/ocos.h#L46-L47

yunmengxie: No, I didn't call this function directly, the memory leak error (AddressSanitizer: new-delete-type-mismatch) was thrown when I tried to register this onnxruntime-extensions library in onnxruntime. Although the macro wraps the returned raw pointer using a shared ptr, the release() call in function CreateLiteCustomOpV2 might already cause the type information of the derived class getting lost.

wenbingl: There is no destructor these classes, so I don't think there could be some real issue on new/delete type mismatch. But this change makes sense and make the memory check tool be happy. We can merge the PR.

Thanks for the contribution!

wenbingl (approved):

PR #890, opened 2025-02-19
Add missing tokens from added_tokens_decoder into added_tokens

sayanshaw24:
Fixes tokenizer issue where special tokens added into `added_tokens_decoder` attribute in `tokenizer_config.json` but missing in `added_tokens` in `tokenizer.json` are still populated correctly into added tokens map such that they are not tokenized incorrectly. This ensures compatibility even with HuggingFace models whose files have inconsistent `tokenizer.json` and `tokenizer_config.json` files.

Validation
- [x] Ran C++ test locally with Phi tokenizer files and validated with different tokens
- [x] Tested correct population of `added_tokens` with missing and complete cases

### DIFF
--- a/operators/tokenizer/bpe_kernels.cc
+++ b/operators/tokenizer/bpe_kernels.cc
@@ -682,11 +682,11 @@ void JsonFastTokenizer::UpdateTokenizer(const TokenJsonConfig& config, const jso
     }
   }
 
-  const json added_tokens_decoder = config.added_tokens_decoder;
+  std::shared_ptr<json> added_tokens_decoder = config.added_tokens_decoder;
 
   // Add any tokens from the added_tokens_decoder that were missing in added_tokens_
-  if (!added_tokens_decoder.empty()) {
-    for (const auto& [id_str, token] : added_tokens_decoder.items()) {
+  if (added_tokens_decoder && !added_tokens_decoder->empty()) {
+    for (const auto& [id_str, token] : added_tokens_decoder->items()) {
       int id = std::stoi(id_str); // Convert key (ID) from string to integer
 
       // Check if this token is already in the added_tokens_
--- a/operators/tokenizer/bpe_kernels.cc
+++ b/operators/tokenizer/bpe_kernels.cc
@@ -682,6 +682,32 @@ void JsonFastTokenizer::UpdateTokenizer(const TokenJsonConfig& config, const jso
     }
   }
 
+  const json added_tokens_decoder = config.added_tokens_decoder;
+
+  // Add any tokens from the added_tokens_decoder that were missing in added_tokens_
+  if (!added_tokens_decoder.empty()) {
+    for (const auto& [id_str, token] : added_tokens_decoder.items()) {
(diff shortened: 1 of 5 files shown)


wenbingl:
> Fixes tokenizer issue where special tokens added into `added_tokens_decoder` attribute in `tokenizer_config.json` but missing in `added_tokens` in `tokenizer.json` are still populated correctly into added tokens map such that they are not tokenized incorrectly. This ensures compatibility even with HuggingFace models whose files have inconsistent `tokenizer.json` and `tokenizer_config.json` files.
> 
> Validation
> 
> * [x]  Ran C++ test locally with PhiO tokenizer files and validated on numerous cases
> * [x]  Tested correct population of `added_tokens` with missing and complete cases

Good finding, thanks!

operators/tokenizer/bpe_kernels.cc
```diff
@@ -682,6 +682,32 @@ void JsonFastTokenizer::UpdateTokenizer(const TokenJsonConfig& config, const jso
     }
   }
 
+  const json added_tokens_decoder = config.added_tokens_decoder;
```
wenbingl, line 685: Can you just merge the following implementation into config::LoadAddedToken. https://github.com/microsoft/onnxruntime-extensions/blob/30609b74a1175fb3d47ff8224f42d9640e7f563d/operators/tokenizer/tokenizer_jsconfig.hpp#L274?

operators/tokenizer/tokenizer_jsconfig.hpp
```diff
@@ -46,6 +46,7 @@ class TokenJsonConfig final {
   ~TokenJsonConfig() {}
   using json = nlohmann::json;
   using json_pointer = nlohmann::json_pointer<std::string>;
+  json added_tokens_decoder;
```
wenbingl, line 49: this json object may not be valid since most of json object are temporary.

wenbingl:
Add a sentence which contains one or more tokens from added_tokens_decoder in the current unit test?

operators/tokenizer/tokenizer_jsconfig.hpp
sayanshaw24: good catch! will change to use a shared pointer to avoid this issue.

operators/tokenizer/bpe_kernels.cc
sayanshaw24: we load `added_tokens_decoder` from the `TokenJsonConfig& config`, so we would need to pass the entire config to LoadAddedTokens as well in that case. i think we should keep it here as this essentially modifies the added_tokens field before passing it to https://github.com/microsoft/onnxruntime-extensions/blob/2893ff3afd915b756befaa8dbb44960f0527fa02/operators/tokenizer/bpe_tokenizer_model.hpp#L287

Note: this is different from https://github.com/microsoft/onnxruntime-extensions/blob/30609b74a1175fb3d47ff8224f42d9640e7f563d/operators/tokenizer/tokenizer_jsconfig.hpp#L274

baijumeswani (approved):

PR #898, opened 2025-02-27
Add check for audio sampling rate other than 8k or 16k Hz

sayanshaw24:
Fixes issue where user was getting vague error when using audio file with 48k Hz sampling rate: https://github.com/microsoft/onnxruntime-extensions/blob/0d5aefa9b2e3c0be57ba3ae2d872b920d6d494a6/include/ort_c_to_cpp.h#L109

### DIFF
--- a/shared/api/speech_features.hpp
+++ b/shared/api/speech_features.hpp
@@ -626,7 +626,7 @@ class Phi4AudioEmbed {
     
     // Currently we only support 8k and 16k Hz sampling rate.
     if (sr_val != 8000 && sr_val != 16000){
-      ORTX_CXX_API_THROW("ort-extensions internal error: currently only 8k and 16k Hz sampling rate is supported. Please resample your audio file to either 8k or 16k.", ORT_RUNTIME_EXCEPTION);
+      return OrtxStatus(kOrtxErrorNotImplemented, "Currently only 8k and 16k Hz sampling rate is supported. Please resample your audio file with unsupported audio sampling rate: " + sr_val);
     }
 
     logmel.Init(sr_val == 8000 ? logmel_8k_attrs_: logmel_attrs_);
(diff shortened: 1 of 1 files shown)


kunal-vaishnavi (approved):

shared/api/speech_features.hpp
```diff
@@ -623,7 +623,13 @@ class Phi4AudioEmbed {
 
     SpeechLibLogMel logmel;
     // already checked in Init
-    logmel.Init(sr_val == 8000? logmel_8k_attrs_: logmel_attrs_);
+    
+    // Currently we only support 8k and 16k Hz sampling rate.
+    if (sr_val != 8000 && sr_val != 16000){
+      ORTX_CXX_API_THROW("ort-extensions internal error: currently only 8k and 16k Hz sampling rate is supported. Please resample your audio file to either 8k or 16k.", ORT_RUNTIME_EXCEPTION);
```
wenbingl, line 629: to be more Mobile friendly and be consistent, please try to return the error status, instead of throwing an exception.

baijumeswani: Also, could we avoid adding library's name in the error message reported? To the end user, it may be confusing to see ort-extensions error if there are multiple libraries involved: win-ai-gallery -> aitk -> ort-genai -> ort-extensions ...

wenbingl: > Also, could we avoid adding library's name in the error message reported? To the end user, it may be confusing to see ort-extensions error if there are multiple libraries involved: win-ai-gallery -> aitk -> ort-genai -> ort-extensions ...

on calling ort-extensions from ort-genai, only error code and error message are expected. so, ort-genai can add whatever prefix for  the error message from ort-extensions.

wenbingl (approved):
Not sure if this change is related to the issue mentioned in the description, but the checker is good.

PR #899, opened 2025-02-28
Fix the string_view subscript overflow for Phi4 Tokenizer

wenbingl:


operators/tokenizer/bpe_utils.hpp
```diff
@@ -397,6 +397,11 @@ class PreTokenizerWithRegEx {
       return {};
     }
 
+    if (j >= m_text.size())  {
+      auto res = m_text.substr(0, j);
+      m_text = m_text.substr(j);
+      return res;
+    }
     i = j;
```
wenbingl, line 405: @sayanshaw24, any reason that why we need use both i and j here?

sayanshaw24: it's because j is a temp variable; before we copy i to it here: https://github.com/microsoft/onnxruntime-extensions/blob/main/operators/tokenizer/bpe_utils.hpp#L373 and so in this line we reset i before the last regex check (for (?i:'s|'t|'re|'ve|'m|'ll|'d)?)

sayanshaw24 (approved):

PR #906, opened 2025-03-06
apply_chat_template C API design in tokenizer

wenbingl:


include/ortx_tokenizer.h
```diff
@@ -171,6 +171,26 @@ extError_t ORTX_API_CALL OrtxTokenId2DArrayGetBatch(const OrtxTokenId2DArray* to
...
+ * if both tokenizer and template_str are provided, the template_str will supersede the tokenizer.
+ *
+ * @param tokenizer Pointer to an OrtxTokenizer used for template processing
+ * @param template_str Null-terminated string representing the chat template, can be null if tokenizer.json has one.
+ * @param input Null-terminated string containing the input to be processed.
+ * @param output Double pointer to an OrtxStringArray that will be populated with the output strings.
+ * @param add_generation_prompt Indicates whether to add a generation prompt to the output (defaults to true).
+ * @return extError_t Returns an error code indicating success or the type of failure.
+ */
+extError_t ORTX_API_CALL OrtxApplyChatTemplate(const OrtxTokenizer* tokenizer, const char* template_str,
```
wenbingl, line 190: a tokenizer object is needed because some bos_token and eos_token, and chat-template stored in json file only be only retrieved from the tokenizer object.

sayanshaw24 (approved):

shared/api/chat_template.cc
```diff
@@ -553,142 +544,148 @@ OrtxStatus TokenizerImpl::DeepSeekChatTemplate(std::string& output) {
...
+
+    // DeepSeek variants
+    { DEEPSEEK_CHAT_TEMPLATE, "deepseek-ai/DeepSeek-R1-Distill-Llama-70B" },
+    { DEEPSEEK_CHAT_TEMPLATE, "deepseek-ai/DeepSeek-R1-Distill-Qwen-32B" }
+};
+
 void TokenizerImpl::InitializeChatParameters(
-    bool add_prompt,
-    const std::string& eos,
-    const std::string& bos,
```
sayanshaw24, line 577: can we still keep `eos` and `bos` here? and below where you set them to `tok_config_->bos_token_` and `tok_config_->eos_token_`, we can check if the parameters exist, if so we set to them, otherwise if the params are empty then we use the tok_config values? this way, for our own testing, we can just pass these vals in and avoid having to include the tokenizer files for each of the tokenizers we want to test (i will have a extensive unit testing PR soon that tests all the core functions for the different models on a number of different cases)

wenbingl: trying the python unit test to do extensive tests. I cannot expect who can manually fill up all these arguments for many models.

And this InitializeChatParameters function breaks the thread-safe for tokenizer object. For example, when a tokenizer object was created but the template function was invoked in multiple thread, the messages and other member variables can be misused due to these functions not read-only (missing const function specifier).

sayanshaw24: sounds good! will move my unit testing to python too and adjust based on these changes.

PR #932, opened 2025-04-22
add USE_WINRT flag to allow other projects to build onnxruntime-extensions with CXX 20 and CppWinRT enabled

vortex-captain:
This PR adds option `USE_WINRT` to `CMakeLists.txt`. If ON, CMake will set `CMAKE_CXX_STANDARD` to `20` instead of `17`, and enable `CppWinRT` for target `ortcustomops`. This fixes a few build issues in other projects which use new versions of WinRT and depend on `onnxruntime-extensions`.


CMakeLists.txt
```diff
@@ -45,7 +45,11 @@ endif()
 # Needed for Java
 set(CMAKE_C_STANDARD 99)
 
-set(CMAKE_CXX_STANDARD 17)
+if(USE_WINRT)
```
snnn, line 48: How about:
```
if(NOT CMAKE_CXX_STANDARD )
  set(CMAKE_CXX_STANDARD 20)
endif()
```

wenbingl, line 48: please add a prefix as other option to avoid the potential conflicts with another project.

vortex-captain: fixed

wenbingl (approved):

PR #942, opened 2025-05-03
 Enable VCPKG

snnn:
Make it be possible to use this repo as a vcpkg port.


sayanshaw24:
@wenbingl had made this PR on Friday as well: https://github.com/microsoft/onnxruntime-extensions/pull/941; ort-extensions only depends on some external header files, no static/dynamic library for ort-genai scenario, so we can either go with that PR, or the vcpkg for the 3 dependencies.

wenbingl:
> @wenbingl had made this PR on Friday as well: #941; ort-extensions only depends on some external header files, no static/dynamic library for ort-genai scenario, so we can either go with that PR, or the vcpkg for the 3 dependencies.

Not sure if it is the same thing. I wasn't aware of this PR, but if this one can make ort-extensions meet the compliance requirement, my PR can be dropped.

PR #947, opened 2025-05-05
Support Chat Template Override for Unsupported Models

sayanshaw24:


shared/api/chat_template.cc
```diff
@@ -722,9 +722,9 @@ void TokenizerImpl::InitializeChatParameters(const char* template_str,
 // ApplyChatTemplate method to choose the template logic based on chat_template
 OrtxStatus TokenizerImpl::ApplyChatTemplate(const TokenizerImpl::MessageList& message_list, std::string& output,
                                             bool add_generation_prompt) const {
-  if (chat_template_parsing_status != "Success"){
-    return OrtxStatus(kOrtxErrorInvalidArgument, "Failed to parse chat template: " + chat_template_parsing_status);
-  }
+  // Note: chat_template_parsing_status holds the value for whether the official chat template from
```
baijumeswani, line 725: Should you throw if chat_template is empty and chat_template_parsing_status != "Success"?

baijumeswani: This PR makes chat_template_parsing_status redundant. Is it no longer needed?

sayanshaw24: it'll throw already then as `chat_template` won't be in `model_to_template_map`.

but yeah chat_template_parsing_status is not used after this PR; i kept it to store the model template parsing status for the future, i can remove it.

baijumeswani (approved):

PR #951, opened 2025-05-08
Add `tools` as chat template input

sayanshaw24:


test/pp_api_test/test_tokenizer_chat.cc
```diff
@@ -37,7 +37,7 @@ TEST(OrtxTokenizerTest, Phi4ChatTemplate) {
 
   auto err = OrtxApplyChatTemplate(
     tokenizer.get(), nullptr,
-    messages_json.c_str(), templated_text.ToBeAssigned(), true, false);
+    messages_json.c_str(), nullptr, templated_text.ToBeAssigned(), true, false);
```
natke, line 40: Can you add actual tests?

sayanshaw24: oh yes, that was simply to ensure that previous C++ test does not break, will add testing in Python.

snnn (approved):

PR #956, opened 2025-05-22
Update <think> tag logic for DeepSeek chat template

sayanshaw24:


shared/api/chat_template.cc
```diff
@@ -559,6 +559,13 @@ OrtxStatus TokenizerImpl::DeepSeekChatTemplate(std::string& output, bool add_gen
 
     // Handle assistant message without tool calls
     if (role == "assistant" && !content.empty()) {
+      
+      // Remove content before and including "</think>" if present
+      std::string::size_type think_pos = content.find("</think>");
+      if (think_pos != std::string::npos) {
+        content = content.substr(think_pos + std::string("</think>").length());
```
edgchen1, line 566: can we avoid constructing a string each time? e.g., with a constant like `constexpr std::string_view think_token = "</think>";` we can just call `think_token.size()`

sayanshaw24: good point, updated!

hanbitmyths (changes requested):
Do we have an existing test to validate this change?

apsonawane (approved):

hanbitmyths (approved):

PR #961, opened 2025-06-02
Fix DeepSeek Chat Template for Think Tag

sayanshaw24:
Another fix on top of #956.

### DIFF
--- a/shared/api/chat_template.cc
+++ b/shared/api/chat_template.cc
@@ -559,19 +559,19 @@ OrtxStatus TokenizerImpl::DeepSeekChatTemplate(std::string& output, bool add_gen
 
     // Handle assistant message without tool calls
     if (role == "assistant" && !content.empty()) {
-      constexpr std::string_view think_token = "</think>";
-      
-      // Remove content before and including "</think>" if present
-      std::size_t think_pos = content.find(think_token);
-      if (think_pos != std::string::npos) {
-        content = content.substr(think_pos + think_token.size());
-      }
-
       if (is_tool) {
         output += "<｜tool_outputs_end｜>" + content;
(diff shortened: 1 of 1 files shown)


shared/api/chat_template.cc
```diff
@@ -559,19 +559,19 @@ OrtxStatus TokenizerImpl::DeepSeekChatTemplate(std::string& output, bool add_gen
...
-      if (think_pos != std::string::npos) {
-        content = content.substr(think_pos + think_token.size());
-      }
-
       if (is_tool) {
         output += "<｜tool_outputs_end｜>" + content;
         output += eos_token;
         is_tool = false;
       } else {
+        constexpr std::string_view think_token = "</think>";
```
baijumeswani: optional: shall we call it `end_of_reasoning_tag` instead of `think_token`. 

```suggestion
        constexpr std::string_view end_of_reasoning_tag = "</think>";
```

shared/api/chat_template.cc
```diff
@@ -559,19 +559,19 @@ OrtxStatus TokenizerImpl::DeepSeekChatTemplate(std::string& output, bool add_gen
...
+      
+        // Remove content before and including "</think>" if present
+        std::size_t think_pos = content.find(think_token);
```
baijumeswani:
```suggestion
        const std::size_t think_pos = content.find(think_token);
```

shared/api/chat_template.cc
sayanshaw24: sure! it's more understandable that way - updated `think_pos` to `end_of_reasoning_pos` too.

baijumeswani (approved):

PR #965, opened 2025-06-17
Refactor SPM-based BPE Tokenizer and Update Chat Template Implementation

sayanshaw24:
### Updates
This PR introduces the following major changes:

1. Generalized chat template support: nothing is fixed anymore; we use a Minja (Jinja C++) chat template parsing engine.
2. Multimodal support, e.g.:
```
messages = 
        [
            {
                "role": "user",
                "content": [
                    {
                        "type": "image",
                        "url": "https://huggingface.co/spaces/big-vision/paligemma-hf/resolve/main/examples/password.jpg",
                    },
                    {"type": "text", "text": "What is the password?"},
                ],
            }
        ]
```
3. Support for content type, even for non-multimodal (i.e. text only) cases - so both OpenAI and HuggingFace standards are supported, therefore enabling full Foundry Local integration.

### Validation

- [x] Native C++ testing against HuggingFace output
- [x] Python unit testing on variety of test cases for new and old scenarios

### DIFF
--- a/test/test_pp_api.py
+++ b/test/test_pp_api.py
@@ -287,9 +287,10 @@ class TestPPAPI(unittest.TestCase):
         decoded_string = tokenizer.detokenize(ortx_input)
         self.assertEqual(decoded_string[0], self.tokenizer_test_sentence)
 
+    @unittest.skipIf(hf_token_id is None, "HF_TOKEN is not available")
     def test_gemma_3_chat_template(self):
         ckpt = "google/gemma-3-4b-it"
-        hf_tok = AutoTokenizer.from_pretrained(ckpt)
+        hf_tok = AutoTokenizer.from_pretrained(ckpt, token=hf_token_id)
 
         messages = [
             {
@@ -374,9 +375,10 @@ class TestPPAPI(unittest.TestCase):
         ortx_inputs = tokenizer.apply_chat_template(message_json, add_generation_prompt=True, tokenize=True)
         np.testing.assert_array_equal(ortx_inputs, inputs)
 
+    @unittest.skipIf(hf_token_id is None, "HF_TOKEN is not available")
     def test_phi3_vision_chat_template(self):
         ckpt = "microsoft/Phi-3-vision-128k-instruct"
-        hf_tok = AutoTokenizer.from_pretrained(ckpt)
+        hf_tok = AutoTokenizer.from_pretrained(ckpt, token=hf_token_id)
 
         for messages in self.messages_list:
             inputs = hf_tok.apply_chat_template(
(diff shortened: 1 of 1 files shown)


operators/tokenizer/bpe_kernels.cc
```diff
@@ -407,100 +417,169 @@ std::vector<int64_t> KernelBpeTokenizer::SpmTokenize(ustring& input, int64_t max
...
+
     std::u32string ustr(seg_id.first);
-    if (add_dummy_prefix) {
-      ustr.insert(ustr.begin(), 0x2581);  // UTF-8 string '\xe2\x96\x81'
-      add_dummy_prefix = false;           // only add dummy prefix once
+    if (add_dummy_prefix && !ustr.empty()) {
+      if (ustr.front() == U' '){
+        special = true;
+      }
+      ustr.insert(ustr.begin(), 0x2581);  // U+2581 = '▁'
```
apsonawane, line 427: Do we not need to make add_dummy_prefix false?

test/pp_api_test/test_tokenizer_chat.cc
```diff
@@ -80,3 +80,278 @@ TEST(OrtxTokenizerTest, Phi4ChatTemplate) {
 
   ASSERT_EQ(std::string(text), expected_decoder_output);
 }
+
+TEST(OrtxTokenizerTest, Phi3MediumChatTemplate) {
```
apsonawane, line 84: Phi3MediumChatTemplate and others are removed right? How does this work?

operators/tokenizer/bpe_kernels.cc
sayanshaw24: no, that was actually a bug from before - the dummy prefix needs to be added once for each segment in `special_token_split_res`, i.e. once each run of this loop if set to true (also tested this against HF in the tests).

the logic comes from the "Prepend" part of the "normalizer" in `tokenizer.json` files:
```
"normalizer": {
    "type": "Sequence",
    "normalizers": [
      {
        "type": "Prepend",
        "prepend": "▁"
      },
      {
        "type": "Replace",
        "pattern": {
          "String": " "
        },
        "content": "▁"
      }
    ]
  }
```

test/pp_api_test/test_tokenizer_chat.cc
sayanshaw24: So now we use our generic Minja parser (C++ based Jinja functionality) for everything :) before we only used it as fallback (and had local core functions like Phi4ChatTemplate, Phi3MediumChatTemplate, etc.), but with the changes in this PR turns out we can now use it for everything - it's quite cool, we now have model-agnostic chat template support!

apsonawane (approved):

PR #970, opened 2025-06-25
Remove farmhash dependency

sayanshaw24:
### Updates
The `farmhash` library is out of maintenance and has trouble with VCPKG and is hence being removed. We replace it with C++ standard library functionality with `std::hash`.

Note that this implies **random seed initialization**, so the same string hashed in one program run may produce a different hash value in the next run to prevent hash-flooding attacks in unordered containers. Therefore, in unit-testing we no longer check for specific expected hash values and instead **test general correctness (shape, range, determinism)**.

### Validation
- [x] Python unit testing in `test/test_string_ops.py`


snnn (dismissed):

snnn (approved):
Do not check in third-party source code to our repos. Soon or later it will be banned.

PR #986, opened 2025-08-06
Fix Compiler Warnings C4333 and C5038

sayanshaw24:
### Updates

Due to BinSkim security errors, we are turning on certain compiler options in our partner library, [ONNXRuntime GenAI](https://github.com/microsoft/onnxruntime-genai), which mean we need to fix certain critical compiler warnings in ONNXRuntime Extensions, which is consumed by ORT GenAI.

We also pin the `safetensors` version to an older, working version, as newer versions unconditionally reference `torch.uint64` (a non-existent attribute).

### Validation
- [x] Successful local build and CI builds

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -5,6 +5,7 @@ onnx             >=1.9.0
 protobuf         < 4.0.0
 # multiple versions of onnxruntime are supported, but only one can be installed at a time
 onnxruntime      >=1.12.0
+safetensors      < 0.4.3
 transformers     < 4.50.0
 tensorflow_text  >=2.5.0;python_version < '3.11'
 requests         >=2.26.0


requirements-dev.txt
```diff
@@ -5,6 +5,7 @@ onnx             >=1.9.0
 protobuf         < 4.0.0
 # multiple versions of onnxruntime are supported, but only one can be installed at a time
 onnxruntime      >=1.12.0
+safetensors      < 0.4.3
```
apsonawane, line 8: Is this dependency required?

sayanshaw24: yeah, CIs fail otherwise because of a safetensors change - Abhishek has noticed this as well.

apsonawane (approved):

PR #989, opened 2025-09-03
Add `skip_special_tokens` to Decoder API

sayanshaw24:
### Updates

Inside the foundry catalog for [Foundry Local](https://github.com/[microsoft/Foundry-Local](https://github.com/microsoft/Foundry-Local)), all tool call tokens are marked as special such that we can use them with guidance inside [GenAI](https://github.com/microsoft/onnxruntime-genai). Now, in order for decoding to work with these tokens, we need to expose using special tokens as an option in our Decoder API in Extensions. Hence, this API adds `skip_special_tokens` to `OrtxDetokenize1DWithOptions`, used in the GenAI `OgaDecoder`.

Note that we leave `OrtxDetokenize1D` as is for backward compatibility. In the future, we shall be adding an extensible options map API that will reduce the need for messy parameter additions to our Encoder/Decoder APIs, and can be easily consumed by GenAI and other partners.

### Validation
- [x] Native C++ testing against HuggingFace benchmark.

### DIFF
--- a/test/pp_api_test/test_tokenizer_chat.cc
+++ b/test/pp_api_test/test_tokenizer_chat.cc
@@ -603,6 +603,24 @@ TEST(OrtxTokenizerTest, Gemma3SpecialChatTemplate) {
   ASSERT_EQ(std::string(special_text), expected_special_decoder_output);
 }
 
+/*
+
+Neither OpenAI nor HuggingFace have explicit Whisper chat template functionality.
+
+OpenAI does not expose Whisper via a chat interface like ChatCompletion.
+Instead, their Whisper API uses raw audio uploads. For finetuning or embedding Whisper
+in pipelines, they rely on pre-tokenized sequences. They don’t use Jinja2 or chat templates
+for Whisper at all — it is purely sequence input with prepended tokens like
+<|startoftranscript|> manually inserted.
+
+In HuggingFace transformers, for Whisper, you are expected to pass in the template manually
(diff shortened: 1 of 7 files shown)


test/pp_api_test/test_tokenizer_chat.cc
```diff
@@ -517,26 +592,17 @@ TEST(OrtxTokenizerTest, Gemma3SpecialChatTemplate) {
...
 
-/*
-
-Neither OpenAI nor HuggingFace have explicit Whisper chat template functionality.
+  // Test detokenization with skip_special_tokens = false
+  OrtxDetokenize1DWithOptions(tokenizer.get(), &ids_vec.front(), ids_vec.size(), decoded_text.ToBeAssigned(), false);
+  const char* special_text = nullptr;
+  OrtxStringArrayGetItem(decoded_text.get(), 0, &special_text);
 
-OpenAI does not expose Whisper via a chat interface like ChatCompletion.
```
kunal-vaishnavi: Why remove these comments?

kunal-vaishnavi (approved):

test/pp_api_test/test_tokenizer_chat.cc
sayanshaw24: good catch, must've been from a previous local change i missed, will add them back!

kunal-vaishnavi (approved):

PR #990, opened 2025-09-04
Fix phi4 mm memory leak

baijumeswani:
There was a memory leak as a result of not freeing up intermediate tensors when processing images and audio for the phi4 multi-modal model.

This pull-request frees up that memory.


jiafatom (approved):

sayanshaw24 (approved):
Thanks for locating and fixing this!

ajindal1 (approved):

PR #992, opened 2025-09-05
Fix Tools Usage for Chat Templates

sayanshaw24:
### Updates
For Qwen, and some other models, `tools` should be a list of functions stored in JSON format, passed in separately, as an additional input to `apply_chat_template`. These tools were not handled correctly in the C++ backend in ApplyChatTemplate and were leading to extra newlines. This is now fixed by this PR.

### Validation
- [x] Python unit testing against HuggingFace.

### DIFF
--- a/test/test_pp_api.py
+++ b/test/test_pp_api.py
@@ -452,21 +452,44 @@ class TestPPAPI(unittest.TestCase):
             assert False, f"Error while trying to pass in tools to chat template: {e}"
 
     def test_qwen2_5_vl_chat_template(self):
-        model_id = "Qwen/Qwen2.5-VL-72B-Instruct"
+        model_id = "Qwen/Qwen2.5-1.5B-Instruct"
+        
+        # For Qwen, tools should be a list of functions stored in json format, passed in
+        # separately, as an additional input to apply_chat_template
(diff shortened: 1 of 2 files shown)


test/test_pp_api.py
```diff
@@ -452,21 +452,44 @@ def test_chat_tools_input(self):
...
+                }
+            },
+        ]
+
+
         messages = [
+            {
+                "role": "system",
+                "content": "You are a helpful assistant",
+                # "tools": json.dumps(tools), # pass the tools into system message using tools argument
```
kunal-vaishnavi, line 479: Should we remove this commented out line?

sayanshaw24: I think it's useful for understanding the qwen/phi-4 tools usage actually - I added another comment near the top of the test to give more context.

natke, line 479: Should the commented debug be removed?

sayanshaw24: I think we should keep it, especially in this Extensions test file; Kunal had the same comment - I think it's useful for understanding the qwen/phi-4 tools usage actually - I added another comment near the top of the test to give more context.

PR #994, opened 2025-09-06
Truncate the audio (30s+) into the first 30 secs

jiafatom:


sayanshaw24 (approved):
great! thanks for adding this

PR #997, opened 2025-09-12
Fix string indexing bug with Phi-4 mm tokenization

kunal-vaishnavi:
### Description

This PR adds two new out-of-bounds checks when running each step while matching the first Phi-4 multimodal pattern during tokenization.

### Motivation and Context

The checks are similar to the ones added [here](https://github.com/microsoft/onnxruntime-extensions/pull/899/). Without these checks, the following error can happen in some environments.

```
/opt/rh/gcc-toolset-12/root/usr/include/c++/12/string_view:239: constexpr const std::basic_string_view<_CharT, _Traits>::value_type& std::basic_string_view<_CharT, _Traits>::operator[](size_type) const [with _CharT = char32_t; _Traits = std::char_traits<char32_t>; const_reference = const char32_t&; size_type = long unsigned int]: Assertion '__pos < this->_M_len' failed.
```

### DIFF
--- a/operators/tokenizer/bpe_kernels.cc
+++ b/operators/tokenizer/bpe_kernels.cc
@@ -521,22 +521,6 @@ std::vector<int64_t> KernelBpeTokenizer::SpmTokenize(ustring& input, int64_t max
           }
         }
 
-        // Handle leading '▁' if in vocab
-        if (mutable_tok.front() == 0x2581) {
-          // Convert the full token to UTF-8 for vocab check
-          std::string full_utf8_token = std::string(ustring(mutable_tok));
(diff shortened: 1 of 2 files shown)


operators/tokenizer/bpe_utils.hpp
```diff
@@ -363,6 +363,12 @@ class PreTokenizerWithRegEx {
   std::u32string_view Match_PHI4_Pattern_1() {
     size_t i = 0;
 
+    if (i >= m_text.size())  {
```
hanbitmyths: Do you want to check m_text is empty?

kunal-vaishnavi: Fixed

PR #998, opened 2025-09-18
Add Tokenizer Options Support

sayanshaw24:


include/ortx_tokenizer.h
```diff
@@ -60,6 +60,18 @@ extern "C" {
...
+/** \brief Create a tokenizer object with the specified tokenizer path and options such as add_special_tokens for tokenizing, or skip_special_tokens for detokenizing.
+ *
+ * \param tokenizer Pointer to store the created tokenizer object
+ * \param tokenizer_path The path to the tokenizer directory, which is utf-8 encoded
+ * \param option_keys Array of option keys
+ * \param option_values Array of option values (same length as keys)
+ * \param num_options Number of options
+ * \return Error code indicating the success or failure of the operation
+ */
+extError_t ORTX_API_CALL OrtxCreateTokenizerWithOptions(OrtxTokenizer** tokenizer, const char* tokenizer_path, const char* option_keys[], const char* option_values[], size_t num_options);
```
baijumeswani, line 94: Could we list all the valid tokenizer options available. Please also specify the expected values for each option if possible.

shared/api/c_api_tokenizer.cc
```diff
@@ -373,7 +436,18 @@ extError_t ORTX_API_CALL OrtxDetokenizeCached(const OrtxTokenizer* tokenizer, Or
   }
 
   cache_ptr->last_text_.clear();
-  status = ReturnableStatus(token_ptr->Id2Token(next_id, cache_ptr->last_text_, cache_ptr->decoder_state_));
+
+  // If skip_special_tokens option exists, use its value, otherwise use default (true)
+  auto it = token_ptr->options_map.find("skip_special_tokens");
+  if (it != token_ptr->options_map.end()) {
+    bool skip_special_tokens = it->second == "true";
```
baijumeswani: I see that some places the option value can accept "false" as well as "0". Do we then also want to support "1". You could abstract this into a function of its own like 

```
bool skip_special_tokens = ParseOption<bool>(option_map, SkipSpecialTokensOptionName)
```

sayanshaw24: Great catch! So the default for `skip_special_tokens` is actually `true` so it suffices to check for "false" or "0" like I did above for `add_special_tokens` in [lines 168-173 in this file](https://github.com/microsoft/onnxruntime-extensions/pull/998/files/570f3130746d5e2f9c58527e25089b78c6ecb22c#diff-653e4ccf47f2ec04ce53e9eb92941686a5c922d5d5ddffbe9acb464ee8013bd5R171), but I missed doing that here - fixed in the latest commit.

As for `ParseOption`, in general I completely agree we should modularize logic like this as much as possible, but in this case, I think it is not needed given how straightforward it is - I'd like to keep the C API as clean as possible. We can do this in the future when we support many more options however.

include/ortx_tokenizer.h
sayanshaw24: Done!

shared/api/c_api_tokenizer.cc
```diff
@@ -47,6 +47,51 @@ extError_t ORTX_API_CALL OrtxCreateTokenizer(OrtxTokenizer** tokenizer, const ch
...
+    return kOrtxErrorInvalidArgument;
+  }
+
+  // Convert keys/values for tokenizer options to unordered_map
+  std::unordered_map<std::string, std::string> options;
+  for (size_t i = 0; i < num_options; ++i) {
+    if (keys[i] && values[i]) {
+      options[keys[i]] = values[i];
+    }
+  }
```
baijumeswani: Shall we check if an option was passed in that is not a valid option (in case of typos from caller side)?

shared/api/c_api_tokenizer.cc
```diff
@@ -65,38 +110,42 @@ extError_t ORTX_API_CALL OrtxCreateTokenizerFromBlob(OrtxTokenizer** tokenizer,
...
 
-  auto token_ptr = static_cast<const TokenizerImpl*>(tokenizer);
-  ReturnableStatus status = token_ptr->IsInstanceOf(extObjectKind_t::kOrtxKindTokenizer);
+  // Convert key/value arrays to unordered_map
+  std::unordered_map<std::string, std::string> options;
+  for (size_t i = 0; i < num_options; ++i) {
+    if (keys[i] && values[i]) {
+      options[keys[i]] = values[i];
+    }
+  }
```
baijumeswani: Move this to a function since it happens multiple times.

shared/api/c_api_tokenizer.cc
baijumeswani: Since we do these checks many times, I think it would make sense to move it into a function of its own so that we can avoid making parsing mistakes from one block of code to another.

shared/api/c_api_tokenizer.cc
```diff
@@ -113,7 +162,20 @@ extError_t ORTX_API_CALL OrtxTokenizeWithOptions(const OrtxTokenizer* tokenizer,
...
                  [](const char* str) { return std::string_view(str); });
 
-  status = token_ptr->Tokenize(input_view, t_ids, add_special_tokens);
+  // If add_special_tokens option exists, use its value, otherwise use default (true)
+  auto it = token_ptr->options_map.find("add_special_tokens");
+  if (it != token_ptr->options_map.end()) {
+    bool add_special_tokens = true; // default
+    std::string val = it->second;
+    std::transform(val.begin(), val.end(), val.begin(), ::tolower);
+    if (val == "false" || val == "0") {
```
kunal-vaishnavi: During input validation, we should convert the string used for the value to all lowercase. One scenario that could break this is if `val == "False"`.

sayanshaw24: yep exactly - that's already being done in the `std::transform` line. I actually added this for that exact reason, and there's a test case for it too in the unit tests (because I know users may parse "False" in Python).

kunal-vaishnavi: Can this transformation be done on the options before entering this method? The `std::transform` call is repeated across multiple functions and could be shared.

sayanshaw24: yep! i'm modularizing everything right now

shared/api/c_api_tokenizer.cc
sayanshaw24: done!

shared/api/c_api_tokenizer.cc
sayanshaw24: done!

kunal-vaishnavi (approved):

PR #1000, opened 2025-10-07
STFT energy-based segmentation for voice detection

nenad1002:


operators/math/dlib/stft_norm.hpp
```diff
@@ -50,3 +54,16 @@ struct StftNormal {
  private:
   int64_t onesided_{1};
 };
+
+  static std::vector<float> hann_window(int N) {
```
sayanshaw24, line 57: what's the incentive to move this here from `shared/api/speech_features.hpp`?

we need to be cognizant that ORT GenAI uses the shared API, so even if moving over the hann win calculation here does not cause issues in Extensions, it may do in GenAI (not sure about this but good to keep in mind)

operators/math/math.cc
```diff
@@ -17,8 +18,9 @@ FxLoadCustomOpFactory LoadCustomOpClasses_Math = []() -> CustomOpArray& {
                                CustomCudaFuncV2("NegPos", neg_pos_cuda),
 #endif
 #ifdef ENABLE_DLIB
-                               CustomCpuFuncV2("Inverse", inverse),
-                               CustomCpuStructV2("StftNorm", StftNormal),
+                               CustomCpuFuncV2("Inverse", inverse), CustomCpuStructV2("StftNorm", StftNormal),
```
sayanshaw24, line 21: can we move `CustomCpuStructV2("StftNorm", StftNormal),` to the next line for readability and consistency?

test/test_audio_signal.py
```diff
@@ -105,6 +107,28 @@ def test_stft_norm_np(self):
         actual = actual[0]
         np.testing.assert_allclose(expected[:, 1:], actual[:, 1:], rtol=1e-3, atol=1e-3)
 
+    def test_stft_energy_based_segmentation_and_merge(self):
```
sayanshaw24, line 110: can we add native C++ unit tests for `DetectEnergySegments` and `MergeAndFilterAudioSegments` as well? perhaps in `test/test_feature_extraction.cc`

in general, having native C++ tests in addition to python ones makes debugging simpler for Extensions.

operators/math/dlib/stft_norm.hpp
```diff
@@ -6,6 +6,10 @@
 #include "ocos.h"
 #include <dlib/matrix.h>
 
+#ifndef M_PI
+#define M_PI 3.14159265358979323846
```
kunal-vaishnavi, line 10: Can we use `M_PI` from the C math library with `#include "math.h"`?

kunal-vaishnavi: Can we use `M_PI` from the C math library with the following addition to handle older platforms?

```c
#define _USE_MATH_DEFINES
#include "math.h"
````

File: operators/math/dlib/stft_norm.hpp (hunk 1 above)
Comment (nenad1002) [reply operators/math/dlib/stft_norm.hpp:57]: hann win is closely correlated with stft calculation, and segment extraction code probably shouldn't include 'speech_features.hpp'. Additionally, I did find dependencies issues when trying to do so, but that's secondary.

If you think this shouldn't be here, I can move it into some other file/header, but I don't think including speech_features.hpp is a way to go..

This shouldnt cause any issue as I just moved the implementation.

File: operators/math/energy_stft_segmentation.cc [hunk 5]
```diff
@@ -0,0 +1,150 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "energy_stft_segmentation.hpp"
+#include "dlib/stft_norm.hpp"
+#include "../../shared/api/c_api_utils.hpp"
+
+OrtStatusPtr detect_energy_segments(const ortc::Tensor<float>& audio, const ortc::Tensor<int64_t>& sr_tensor,
```
kunal-vaishnavi: Can you add some code comments in each section regarding what these calculations are doing and why they need to be done? It will help future readers know more.

operators/math/energy_stft_segmentation.cc
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  float max_val = energy_db[0];
+  std::vector<float> tmp = energy_db;
+  std::nth_element(tmp.begin(), tmp.begin() + tmp.size() / 2, tmp.end());
+  float median_val = tmp[tmp.size() / 2];
+  for (float v : energy_db) max_val = std::max(max_val, v);
+  float threshold = std::max(max_val + energy_threshold_db, median_val);
+
+  std::vector<bool> mask(n_frames);
```
kunal-vaishnavi: Rather than store a vector of bools for the mask, can we compute the mask on-the-fly for each value of `t` and save memory? The subsequent for-loop goes through the number of frames and it can be computed then.

operators/math/energy_stft_segmentation.cc
```diff
@@ -0,0 +1,150 @@
...
+
+  // Merge overlapping or close segments directly
+  std::vector<std::pair<int64_t, int64_t>> merged;
+  merged.reserve(num_segments);
+
+  int64_t cur_start = seg_data[0];
+  int64_t cur_end = seg_data[1];
+
+  for (int64_t i = 1; i < num_segments; ++i) {
+    const int64_t s = seg_data[i * 2 + 0];
```
kunal-vaishnavi:
```suggestion
    const int64_t s = seg_data[i * 2];
```

operators/math/energy_stft_segmentation.cc
```diff
@@ -0,0 +1,150 @@
...
+    float end_s = static_cast<float>(n_frames * hop_length) / sr;
+    segments.emplace_back(start_s, end_s);
+  }
+
+  const int64_t num_segments = static_cast<int64_t>(segments.size());
+  std::vector<int64_t> out_shape = {num_segments, 2};
+  int64_t* out_data = output0.Allocate(out_shape);
+
+  for (int64_t i = 0; i < num_segments; ++i) {
+    out_data[i * 2 + 0] = static_cast<int64_t>(segments[i].first * 1000.0f);
```
kunal-vaishnavi:
```suggestion
    out_data[i * 2] = static_cast<int64_t>(segments[i].first * 1000.0f);
```

operators/math/energy_stft_segmentation.cc
nenad1002: Yeah, it is possible, I can skip the mask

operators/math/energy_stft_segmentation.cc
```diff
@@ -0,0 +1,150 @@
...
+      cur_start = s;
+      cur_end = e;
+    }
+  }
+  merged.emplace_back(cur_start, cur_end);
+
+  const int64_t m = static_cast<int64_t>(merged.size());
+  int64_t* out = output0.Allocate({m, 2});
+  for (int64_t i = 0; i < m; ++i) {
+    out[i * 2 + 0] = merged[i].first;
```
kunal-vaishnavi:
```suggestion
    out[i * 2] = merged[i].first;
```

operators/math/energy_stft_segmentation.hpp
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+
+#include "ocos.h"
+
+OrtStatusPtr detect_energy_segments(const ortc::Tensor<float>& audio, const ortc::Tensor<int64_t>& sr_tensor,
+                                 const ortc::Tensor<int64_t>& frame_ms_tensor,
+                                 const ortc::Tensor<int64_t>& hop_ms_tensor,
```
kunal-vaishnavi: Looks like there's some spacing issues

onnxruntime_extensions/_cuops.py
```diff
@@ -252,6 +252,38 @@ def get_outputs(cls):
...
+        ]
+
+    @classmethod
+    def get_outputs(cls):
+        return [
+            cls.io_def('value', onnx_proto.TensorProto.INT64, [None])
+        ]
+
+
+class MergeAndFilterAudioSegments(CustomOp):
```
kunal-vaishnavi: Can we keep the name simple like the above?

```suggestion
class MergeAudioSegments(CustomOp):
```

nenad1002: I actually think I need to remove the word Audio out altogether, it's not specific to Audio, could be any signal

operators/math/dlib/stft_norm.hpp
sayanshaw24: sounds good, let's leave it here for now then.

onnxruntime_extensions/_cuops.py
```diff
@@ -252,6 +252,38 @@ def get_outputs(cls):
         ]
 
 
+class SplitSignalEnergySegments(CustomOp):
```
kunal-vaishnavi:
```suggestion
class SplitSignalSegments(CustomOp):
```

nenad1002: Why? Adding energy suggests how is it split.

nenad1002: Maybe SplitSignalByEnergySegments ?

kunal-vaishnavi: We should keep the naming schema for the two methods being introduced to be the same. The other method is titled as `MergeSignalSegments` so this suggestion was to follow that pattern. The energy threshold being provided as a parameter can suggest how it is split. In the future, other ways to split the segments could be introduced as well.

nenad1002: Oh, right, we can do this, but we would have to rename it in future as extending the SplitSignalSegments method would meed messy and would break a single responsibility principle.

sayanshaw24:
Ignoring CI failure for merge since we need this PR in. FYI CI is currently failing on main, needs looking into.

PR #1003, opened 2025-10-15
Fix CI Pipeline Issues

sayanshaw24:
### Description

This PR updates the CI pipelines and development dependencies to align with current PyTorch and ONNX Runtime compatibility across all supported platforms.

- **CI Matrix Update:**
  Dropped **Python ≤ 3.10** to match ONNX Runtime’s supported versions.
  
- **Platform-Specific PyTorch Alignment:**
  Fixed torch version (2.4.x is the last version where all unit tests - GPT-2, Whisper, etc. - still align cleanly, e.g.`torch.onnx.export` usage is different in latest versions) and addresses macOS wheel availability issues for newer versions (2.4.x not yet published).
  
    Windows & Linux:
    ```
    torch==2.4.1
    torchvision==0.19.1
    torchaudio==2.4.1
    ```
    
    macOS:
    ```
    torch==2.2.2
    torchvision==0.17.2
    torchaudio==2.2.2
  ```
- **Missing `onnxscript` dependency:**
  Some tests rely on `onnxscript`, which was not previously listed in dev requirements, leading to `ModuleNotFoundError` failures in unrelated files during CI. Added `onnxscript==0.1.0` to `requirements-dev.txt` to fix this.

- **Pinned package versions:**
Pinned the following for transient dependency mismatches and stabilized CI installs.:
  ```
  onnx==1.17.0  
  protobuf==4.25.8  
  onnxscript==0.2.7
  ```
  
- **Old Gradle Version:**
Updated **Gradle wrapper validation** to use `v2` action.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -5,7 +5,7 @@ onnx             ==1.17.0
 protobuf         ==4.25.8
 # multiple versions of onnxruntime are supported, but only one can be installed at a time
 onnxruntime      >=1.18.0
-onnxscript       ==0.2.7
+onnxscript       ==0.1.0
 safetensors      < 0.4.3
 transformers     < 4.50.0
 tensorflow_text  >=2.5.0;python_version < '3.11'


nenad1002 (approved):

requirements-dev.txt
```diff
@@ -5,9 +5,10 @@ onnx             >=1.9.0
 protobuf         < 4.0.0
 # multiple versions of onnxruntime are supported, but only one can be installed at a time
 onnxruntime      >=1.12.0
+onnxscript       >=0.1.0
 safetensors      < 0.4.3
 transformers     < 4.50.0
-tensorflow_text  >=2.5.0;python_version < '3.11'
+tensorflow_text  >=2.5.0; python_version < '3.11' and python_version != '3.10'
```
snnn: You may consider dropping the support for python 3.10 and below. ORT already does so. Then it will be easier.

requirements-dev.txt
```diff
@@ -5,9 +5,10 @@ onnx             >=1.9.0
 protobuf         < 4.0.0
 # multiple versions of onnxruntime are supported, but only one can be installed at a time
 onnxruntime      >=1.12.0
```
snnn: better to pin these versions instead of using `==` .

sayanshaw24: just ort or you mean every requirement?

snnn: You may decide it. The current file looks a bit complicated, because of the various conditions.

nenad1002 (approved):

PR #1004, opened 2025-10-19
Expose C API for SplitSignalSegments and MergeSignalSegments

nenad1002:


test/pp_api_test/test_feature_extraction.cc
```diff
@@ -295,42 +295,48 @@ TEST(ExtractorTest, TestStftEnergySegmentationAndMerge) {
...
 
-  ortc::Tensor<int64_t> hop(alloc);
-  hop.Allocate({1})[0] = 10;
+  ortc::Tensor<int64_t> hop_ms(alloc);
+  hop_ms.Allocate({1})[0] = 10;
 
-  ortc::Tensor<float> thr(alloc);
-  thr.Allocate({1})[0] = -40.0f;
+  ortc::Tensor<float> energy_threshold_db(alloc);
+  energy_threshold_db.Allocate({1})[0] = -40.0f;
```
sayanshaw24, line 313: where do we get these numbers from? could we add some comments here to describe the standards for future reference/debugging (same for `out_shape` below, etc.)

nenad1002: It's just a test example, the diff of 40 decibels can be a reasonable diff between voice and silence (or background noise)

sayanshaw24:
```suggestion
  // Difference of 40 decibels can be a reasonable diff between voice and silence (or background noise)
  energy_threshold_db.Allocate({1})[0] = -40.0f;

sayanshaw24 (approved):

PR #1006, opened 2025-11-10
Add Phi-4 Tools Quote Normalization

sayanshaw24:
## Description

This PR improves the previously introduced [Tool Normalization features](https://github.com/microsoft/onnxruntime-extensions/pull/1002) by addressing the issue of inconsistent handling of tool quotes in Phi-4 function calling. It introduces the `normalize_tool_quotes` function to maintain consistency in JSON serialization and deserialization, particularly when tools are embedded within the message body as it is in the case of Phi-4, rather than as an additional input like with Qwen 2.5 and most models.

## Validation

- [x] Native C++ unit testing for all three tool formats for Phi-4-mini (Minja/Phi-4, OpenAI `tool`, OpenAI `function`)
- [x] E2E testing in Foundry Local with supported Phi-4 and Qwen 2.5 models
- [x] Verified that chat template rendering produces the expected system/user/assistant block
- [x] Confirmed that existing tests and APIs remain unaffected when tools are not provided/provided in previous standard

### DIFF
--- a/shared/api/chat_template.cc
+++ b/shared/api/chat_template.cc
@@ -304,6 +304,31 @@ static json NormalizeTools(const char* tools_str) {
   return normalized;
 }
 
+/*
+ * This function normalizes quotes in tool-related strings within the input message. This normalization is crucial for
+ * ensuring consistent JSON serialization and deserialization when working with tool calls. Specifically, the function
+ * helps avoid differences between `json::dump` and `json::parse` behavior, which can occur when single quotes are used
+ * incorrectly in tools. This also leads to missing initial tool tag tokens (<|tool_call|>) during inference.
+ *
+ * In the case of Phi-4, when tools are embedded within messages, Minja's handling of tools within the message body
+ * (instead of treating them as separate input keys/values) can lead to issues with the serialization process. For instance,
+ * `json::dump` might escape single quotes, but `json::parse` might not interpret them correctly, resulting in inconsistent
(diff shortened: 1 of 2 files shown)


shared/api/chat_template.cc
```diff
@@ -304,6 +304,59 @@ static json NormalizeTools(const char* tools_str) {
...
+ * (such as '{', ':', ' ', or ',') which typically indicate the presence of tool keys and values.
+ *
+ * Parameters:
+ * - input (const std::string&): The input string containing single quotes that may need to be normalized.
+ *
+ * Returns:
+ * - std::string: A new string where appropriate single quotes are replaced by double quotes.
+ *
+ * Example:
+ * - If the input string is "{'key': 'value'}", the function will return "{\"key\": \"value\"}".
```
edgchen1, line 330: could a single quote within a quoted string be replaced with a double quote? is that the intended behavior?

`"some key": "some value that has a single quote: ' "`

sayanshaw24: great point, Kunal and I brainstormed this fallacy - the motivation for this PR is to fix a missing tool call token that arises from single quotes; and these unfortunately cannot be replaced for just the keys/values in the JSON because parsing them to/fro JSON reintroduces the issue. of course this is still brittle, like for lone quotes and typos like you mentioned; also, long term, Phi-4 tools will be processed just like other models, but currently (because of how the model is written) they are processed as part of the messages.

PR #1007, opened 2025-11-11
Add Android 16KB page size support for ONNX Runtime Extensions

ssk18:
Add linker flags and build configuration to support Android devices
  using 16KB page sizes. This ensures compatibility with Android 15+
  devices while maintaining backward compatibility with 4KB page systems.

  Changes:
  - CMakeLists.txt: Add -Wl,-z,max-page-size=16384 linker flag for 
    Android shared and executable targets
  - cmake/ext_java.cmake: Configure JNI library with 16KB page alignment
  - tools/build.py: Enable ANDROID_SUPPORT_FLEXIBLE_PAGE_SIZES flag
  - java/build-android.gradle: Add NDK ABI filters and CMake arguments
  - java/src/test/android/app/build.gradle: Configure test app for 16KB

  Technical details:
  - All LOAD segments now align to 0x4000 (16384 bytes)
  - Applied to libonnxruntime_extensions4j_jni.so and libortextensions.so

  Testing:
  - Built AAR for arm64-v8a architecture
  - Verified LOAD segment alignment using NDK's llvm-readelf
  - Confirmed both native libraries use 16KB page alignment"

### DIFF
--- a/.github/workflows/android-16kb-check.yml
+++ b/.github/workflows/android-16kb-check.yml
@@ -94,8 +94,8 @@ jobs:
             # Display LOAD segments
             $READELF -l "$LIB" | grep -A 1 "LOAD" || true
 
-            # Check for 16KB alignment (0x4000)
-            if $READELF -l "$LIB" | grep -q "0x4000"; then
+            # Check for 16KB alignment (0x4000) in the Align field of LOAD segments
+            if $READELF -l "$LIB" | grep "LOAD" | awk '{print $NF}' | grep -q '^0x4000$'; then
               echo "PASS: Found 16KB alignment (0x4000) in $(basename $LIB)"
             else
               echo "FAIL: Did not find 16KB alignment (0x4000) in $(basename $LIB)"


CMakeLists.txt
```diff
@@ -218,6 +218,24 @@ if(OCOS_BUILD_ANDROID)
...
+
+  # Add 16KB page size compatibility for Android 15+ devices
+  # See: https://developer.android.com/guide/practices/page-sizes
+  # This flag ensures binaries work on both 4KB and 16KB page size systems
+  option(ANDROID_SUPPORT_FLEXIBLE_PAGE_SIZES "Enable 16KB page size support for Android" ON)
+
+  if(ANDROID_SUPPORT_FLEXIBLE_PAGE_SIZES)
+    # Android NDK uses Clang + lld which supports this flag
+    if(CMAKE_C_COMPILER_ID MATCHES "Clang|GNU")
+      set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-z,max-page-size=16384")
```
edgchen1: is it necessary to specify the link options in this variable and also with `target_link_options()`?

CMakeLists.txt
```diff
@@ -218,6 +218,24 @@ if(OCOS_BUILD_ANDROID)
   endif()
 
   set(OCOS_BUILD_JAVA ON CACHE INTERNAL "")
+
+  # Add 16KB page size compatibility for Android 15+ devices
+  # See: https://developer.android.com/guide/practices/page-sizes
+  # This flag ensures binaries work on both 4KB and 16KB page size systems
+  option(ANDROID_SUPPORT_FLEXIBLE_PAGE_SIZES "Enable 16KB page size support for Android" ON)
```
edgchen1: is there a good reason to not enable it or can it simply always be enabled?

CMakeLists.txt
ssk18: good point.  The global CMAKE_SHARED_LINKER_FLAGS already handles these targets, so the explicit target_link_options calls are redundant.

CMakeLists.txt
ssk18: it should always be enabled in production builds as Google Play will eventually require 16KB support and it is backward compatible.

edgchen1: in that case, can we just remove the option and always add these linker flags for Android builds?

.github/workflows/android-16kb-check.yml
```diff
@@ -0,0 +1,127 @@
...
+  pull_request:
+    paths:
+      - 'CMakeLists.txt'
+      - 'cmake/**'
+      - 'java/**'
+      - 'tools/build.py'
+      - '.github/workflows/android-16kb-check.yml'
+  push:
+    branches:
+      - 'enable-16KB-for-android'  # Your branch name
```
edgchen1: branch name should be updated

.github/workflows/android-16kb-check.yml
```diff
@@ -0,0 +1,127 @@
+name: Android 16KB Page Size Verification
```
edgchen1, line 1: @sayanshaw24 is it ok to add a new workflow for this check? or would it be preferable to add it to the existing CI build pipeline?

CMakeLists.txt
```diff
@@ -218,6 +218,24 @@ if(OCOS_BUILD_ANDROID)
...
+
+  if(ANDROID_SUPPORT_FLEXIBLE_PAGE_SIZES)
+    # Android NDK uses Clang + lld which supports this flag
+    if(CMAKE_C_COMPILER_ID MATCHES "Clang|GNU")
+      set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-z,max-page-size=16384")
+      set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-z,max-page-size=16384")
```
edgchen1: do we need to set `CMAKE_EXE_LINKER_FLAGS` or do we just care about the shared library targets?

CMakeLists.txt
ssk18: the only reason i kept it as an option is pre-r23 NDKs could cause build errors. if thats not the concern then, i can remove this option and always enable it

CMakeLists.txt
ssk18: lets keep it for completeness. It ensures test executables work on 16KB page devices if/when someone runs the Android tests.

CMakeLists.txt
edgchen1: currently, it says [here](https://developer.android.com/ndk/downloads/) that the oldest supported version is r27d. I think we don't need to worry about supporting <r23

CMakeLists.txt
edgchen1: for completeness, there's also `CMAKE_MODULE_LINKER_FLAGS`. would it be simpler to use [`add_link_options()`](https://cmake.org/cmake/help/latest/command/add_link_options.html)?

CMakeLists.txt
```diff
@@ -218,6 +218,13 @@ if(OCOS_BUILD_ANDROID)
   endif()
 
   set(OCOS_BUILD_JAVA ON CACHE INTERNAL "")
+
+  # Add 16KB page size compatibility for Android 15+ devices
+  # See: https://developer.android.com/guide/practices/page-sizes
+  # This flag ensures binaries work on both 4KB and 16KB page size systems.
+  # Requires NDK r23+ (project minimum is r27d).
```
edgchen1: nit: let's not mention an explicit project minimum here to avoid having to update the comment later. could just say that r23 is no longer supported.
```suggestion
  # Requires NDK r23+.
```

README.md
```diff
@@ -10,6 +10,9 @@ Introduction: ONNXRuntime-Extensions is a C/C++ library that extends the capabil
 ## Quickstart
 The library can be utilized as either a C/C++ library or other advance language packages like Python, Java, C#, etc. To build it as a shared library, you can use the `build.bat` or `build.sh` scripts located in the root folder. The CMake build definition is available in the `CMakeLists.txt` file and can be modified by appending options to `build.bat` or `build.sh`, such as `build.bat -DOCOS_BUILD_SHARED_LIB=OFF`. For more details, please refer to the [C API documentation](./docs/c_api.md).
 
+### Android 16KB Page Size Support
```
edgchen1: this is good information, but I wonder if it belongs in the top-level readme. maybe it would be better as a code comment.

.github/workflows/android-16kb-check.yml
edgchen1, line 55: also pass `--update` option to generate build files

```suggestion
            --config=Release \
            --update \
```

ssk18:
@edgchen1 these CI failures look unrelated to this change.

edgchen1:
> @edgchen1 these CI failures look unrelated to this change.

they are not related to this change. we'll need to look into it.

ssk18:
@edgchen1 Could you please get this change merged by fixing the build. It's needed for android 16KB compatibility by April.

.github/workflows/android-16kb-check.yml
sayanshaw24: it would be preferable to add it to an existing one - @ssk18 could you add this to AndroidBuilds?

edgchen1: > could you add this to AndroidBuilds?

that would be here:
https://github.com/microsoft/onnxruntime-extensions/blob/a5f922a8a9f4cc9be0c520a746198b8e0bfad14f/.pipelines/ci.yml#L716

operators/tokenizer/tokenizer_common.h
```diff
@@ -5,6 +5,7 @@
 
 #include <string>
 #include <string_view>
+#include <unordered_map>
```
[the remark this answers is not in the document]
edgchen1: it's fine. this change was made to fix a build error.

.github/workflows/android-16kb-check.yml
edgchen1: can move this in a follow up PR.

PR #1008, opened 2025-11-12
Add Initial Qwen-2.5VL Image Processor Support

sayanshaw24:
## Description

This PR introduces support for **Qwen2.5-VL image preprocessing** in ONNX Runtime Extensions. It implements the full resize (including smart resize) → rescale → normalize → patching pipeline required by `Qwen2.5-VL-7B-Instruct`.

Key updates:

- **New Qwen2.5-VL–compatible `PatchImage` op**
  - Converts RGB HWC to CHW channel ordering
  - Performs accurate temporal padding and 9-D reshape/transpose logic  
  - Produces Python-aligned patch embeddings

- **Smart Resize support in the `Resize` op**
  - Implements Qwen2.5-VL’s shortest-edge and longest-edge constraints  
  - Supports pixel-count–based resizing (min/max pixels)  
  - Matches `smart_resize` behavior from the HF transformers

- **Qwen2.5-VL Normalization handling**
  - Fixes channel-order expectations across resize → normalize → patch  
  - Ensures exact mean/std behavior equivalent to HF transformers
  - Includes C++ normalization adjustments when Qwen2.5-VL mode is enabled

- **Qwen2.5-VL processor JSON**
  - Includes Qwen-specific resize parameters  
  - Correct `mean`/`std` normalization  
  - Patch dimensions (`patch_size`, `merge_size`, `temporal_patch_size`)

- **RGB Correctness Guarantees**
  - Fixes the BGR→RGB mismatch observed in `PatchImage`
  - Ensures alignment with the Qwen2.5VL

This enables **ORT Extensions processing support for Qwen2.5-VL vision & multimodal models** across ORT Extensions, ready-to-use in ORT GenAI. Note that we currently only provide single-image and non-video support.

## Validation

- [x] **C++ unit tests** verifying `pixel_values` match reference output; **MSE comparison**: achieved ≤ `1e-3`
- [x] **Step-by-step parity checks** across:  
  - Resize  
  - Rescale  
  - Normalize
  - Patch extraction
- [x] **Regression-tested** with Phi-3-V, Phi-4, MLlama, Gemma, etc. — no changes in their outputs

### DIFF
--- a/shared/api/image_transforms.hpp
+++ b/shared/api/image_transforms.hpp
@@ -106,17 +106,17 @@ struct Resize {
 
   int64_t round_by_factor(int64_t number, int64_t factor) {
       // Returns the closest integer to 'number' that is divisible by 'factor'.
-      return static_cast<int>(std::round(static_cast<double>(number) / factor) * factor);
+      return static_cast<int64_t>(std::round(static_cast<double>(number) / factor) * factor);
   }
 
   int64_t ceil_by_factor(int64_t number, int64_t factor) {
       // Returns the smallest integer greater than or equal to 'number' that is divisible by 'factor'.
-      return static_cast<int>(std::ceil(static_cast<double>(number) / factor) * factor);
+      return static_cast<int64_t>(std::ceil(static_cast<double>(number) / factor) * factor);
   }
 
   int64_t floor_by_factor(int64_t number, int64_t factor) {
       // Returns the largest integer less than or equal to 'number' that is divisible by 'factor'.
-      return static_cast<int>(std::floor(static_cast<double>(number) / factor) * factor);
+      return static_cast<int64_t>(std::floor(static_cast<double>(number) / factor) * factor);
   }
 
   std::tuple<int64_t, int64_t> smart_resize(int64_t height, int64_t width) {
@@ -234,6 +234,16 @@ struct Resize {
   bool keep_aspect_ratio_{true};
   std::string interpolation_{"CUBIC"};  // LINEAR, NEAREST, CUBIC
 
+  /*
+  These constants below are defaults from Qwen2/Qwen2.5-VL image processing, and come from
+  the config: https://huggingface.co/Qwen/Qwen2.5-VL-7B-Instruct/blob/main/preprocessor_config.json#L2
+
+  Some have calculations behind them, for instance:
+  min_pixels_ = 4 * 28 * 28
+  max_pixels_ = 16384 * 28 * 28,
+  but they are passed in through a JSON config and hence passed as constants (same for defaults).
+  */
+
   bool smart_resize_{false};
   int64_t image_factor_{28};
   int64_t min_pixels_{3136};
@@ -349,6 +359,7 @@ struct Normalize {
   }
 
  private:
+  // Constants set to defaults
   std::vector<float> mean_{0.48145466f, 0.4578275f, 0.40821073f};
   std::vector<float> std_{0.26862954f, 0.26130258f, 0.27577711f};
   bool qwen2_5_vl_{false};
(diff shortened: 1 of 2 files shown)


shared/api/image_transforms_qwen2_5.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+    // Allocate a temporary non-const buffer for the output data
+    std::vector<float> temp_buffer(patches.size());
+
+    // Copy the data into the non-const buffer
+    std::copy(patches.begin(), patches.end(), temp_buffer.begin());
+
+    // Assuming output.Allocate returns a const pointer, we can't directly write to it.
+    // So, we will copy the temp_buffer into the const data pointer.
+    auto* p_output_image = const_cast<float*>(output.Data());  // Convert the const pointer to non-const pointer
```
apsonawane: PatchImage only returns image, but we need image_grid as well as the output. Since vision model has 2 inputs --> pixel_values and image_grid

sayanshaw24: yep, exactly, will add the `image_grid_thw` output values in a commit along with other changes as well once they're done.

shared/api/image_transforms.hpp
```diff
@@ -177,6 +233,14 @@ struct Resize {
...
   bool keep_aspect_ratio_{true};
   std::string interpolation_{"CUBIC"};  // LINEAR, NEAREST, CUBIC
+
+  bool smart_resize_{false};
+  int64_t image_factor_{28};
+  int64_t min_pixels_{3136};
+  int64_t max_pixels_{12845056};
+  int64_t patch_size_{14};
+  int64_t merge_size_{2};
+  double max_ratio_{200.0};
```
hanbitmyths, line 253: Where are all constant numbers coming from?

shared/api/image_transforms.hpp
```diff
@@ -104,6 +104,46 @@ struct Resize {
     return methods;
   }
 
+  int64_t round_by_factor(int64_t number, int64_t factor) {
+      // Returns the closest integer to 'number' that is divisible by 'factor'.
+      return static_cast<int>(std::round(static_cast<double>(number) / factor) * factor);
```
hanbitmyths: cast to int64_t?

shared/api/image_transforms.hpp
sayanshaw24: These are defaults from Qwen2/Qwen2.5-VL image processing, and come from the config: https://huggingface.co/Qwen/Qwen2.5-VL-7B-Instruct/blob/main/preprocessor_config.json#L2

Some have calculations behind them, for instance:
`min_pixels_` = 4 * 28 * 28
`max_pixels_` = 16384 * 28 * 28,
but they are passed in through the JSON config and hence passed as constants (same for defaults).

Will add a comment here explaining this.

shared/api/image_transforms.hpp
sayanshaw24: good catch, updated!

hanbitmyths (approved):

shared/api/image_transforms_qwen2_5.hpp
```diff
@@ -0,0 +1,180 @@
...
+        for (int64_t w = 0; w < W; ++w) {
+          size_t base = static_cast<size_t>((h * W + w) * C);
+          // Input is RGB so base+0=R, base+1=G, base+2=B
+          chw.push_back(data[base + ch]);
+        }
+      }
+    }
+
+    // Add batch dimension (frames = 1 initially) and prepare patches vector
+    std::vector<float> patches = chw;
```
tianleiwu, line 45: We shall avoid such vector copy. 

I think we can allocate with padded space. Fill them with padded value. Then apply HWC -> CHW (need to compute destination offset with padded height / width).

PR #1009, opened 2025-11-12
Fix byte fallback

apsonawane:
This pull request improves the robustness and correctness of the `SpmUgmTokenizer` implementation, particularly in its handling of byte-level tokenization, unknown token segmentation, and boundary conditions. The main focus is on better support for Marian-style tokenizers, more accurate byte fallback behavior, and several bug fixes.

**Tokenizer compatibility and byte coverage:**

* Ensures that when using `MarianTokenizer`, all 256 possible byte tokens are present in the vocabulary, assigning them the minimum score to guarantee full byte coverage as expected by Marian.

**Byte fallback and unknown token handling:**

* Implements explicit byte-level fallback segmentation for each byte of a UTF-8 codepoint when `byte_fallback_` is enabled, improving tokenization accuracy for unknown characters.
* Adjusts logic to allow consecutive unknown tokens in the output sequence when `byte_fallback_` is enabled, ensuring correct segmentation in edge cases.

**Bug fixes and boundary corrections:**

* Fixes an off-by-one error in the `GetNode` method, preventing out-of-bounds access in the XCDA array.
* Corrects the returned match length in user-defined token matching to ensure the normalized length and consumed input reflect the matched token length, not the total prefix length.

### DIFF
--- a/shared/api/chat_template.cc
+++ b/shared/api/chat_template.cc
@@ -304,6 +304,59 @@ static json NormalizeTools(const char* tools_str) {
   return normalized;
 }
 
+/*
+ * This function normalizes quotes in tool-related strings within the input message. This normalization is crucial for
+ * ensuring consistent JSON serialization and deserialization when working with tool calls. Specifically, the function
+ * helps avoid differences between `json::dump` and `json::parse` behavior, which can occur when single quotes are used
+ * incorrectly in tools. This also leads to missing initial tool tag tokens (<|tool_call|>) during inference.
+ *
+ * In the case of Phi-4, when tools are embedded within messages, Minja's handling of tools within the message body
+ * (instead of treating them as separate input keys/values) can lead to issues with the serialization process. For instance,
+ * `json::dump` might escape single quotes, but `json::parse` might not interpret them correctly, resulting in inconsistent
+ * output when processing tools as part of a message.
+ *
+ * To mitigate this, the function scans through the string and normalizes single quotes to double quotes in the context
+ * of tool calls. This ensures that tool keys and values are serialized consistently, and the tool call context remains
+ * intact. The function specifically looks for single quotes that are preceded or followed by specific characters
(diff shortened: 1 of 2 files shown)


operators/tokenizer/ugm_kernels.hpp
```diff
@@ -243,12 +243,25 @@ struct SpmUgmTokenizer {
...
+    // Add explicit byte tokens (<0xXX> converted earlier to single-byte) if Marian expects full byte coverage
     double min_score = DBL_MAX;
+    for (const auto& entry : vocab_) {
+      min_score = std::min<double>(min_score, std::get<1>(entry.second));
+    }
+    if (config.tokenizer_class_ == "MarianTokenizer") {
+      // Ensure all 256 byte values have entries; use min_score so they are not worse than existing minimum
+      for (int b = 0; b < 256; ++b) {
+        std::string btok(1, static_cast<unsigned char>(b));
+        if (vocab_.count(btok) == 0) {
```
sayanshaw24, line 255: should we do some unit perf testing to see how this affects perf since we're searching vocab for `btok` each time here?

operators/tokenizer/ugm_kernels.hpp
```diff
@@ -339,7 +364,7 @@ struct SpmUgmTokenizer {
     for (struct BestTokenization& tokenization = tokenization_results[input_len];;
          tokenization = tokenization_results[tokenization.input_offset]) {
       bool is_unknown = tokenization.token_id == special_unk_id_;
-      if (!(is_prev_unknown && is_unknown)) {
+      if (!( !byte_fallback_ && is_prev_unknown && is_unknown)) { // Fix #5: keep consecutive unknowns when byte_fallback_ is true
```
sayanshaw24: could you update the comments to have a one-line explanation or so for the main changes - e.g. _why_ we keep consecutive unknowns when `byte_fallback_` is true?

sayanshaw24:
thanks for adding this! left some comments but looks mostly good

PR #1010, opened 2025-11-19
[NuGet] Publish import-libs and public headers for native builds

amirkutcher:
This change includes missing import-libs and public header files for native (c/c++) build & linkage,
In addition, the change includes VS 2026 cmake generator support.


build.bat
```diff
@@ -25,6 +25,8 @@ set GEN_PLATFORM=-A x64
 set GENERATOR="Visual Studio 16 2019"
 IF "%VisualStudioVersion:~0,2%" == "16" GOTO :START_BUILD
 set GENERATOR="Visual Studio 17 2022"
+IF "%VisualStudioVersion:~0,2%" == "17" GOTO :START_BUILD
+set GENERATOR="Visual Studio 18 2026"
```
sayanshaw24: do we need to set this here? `build.bat` is our go-to build script, and we have machine pools on older versions that may need to be on VS 2022, and cause problems otherwise; i think a VS upgrade can be separated from this PR and needs more consideration.

nuget/WinOnlyNuget.nuspec
```diff
@@ -24,7 +24,9 @@
 		<file src="..\ThirdPartyNotices.txt" target="ThirdPartyNotices.txt" />
 		
 		<!-- Src paths for local Windows build via netstandard -->
-		<file src="..\out\Windows\bin\RelWithDebInfo\ortextensions.dll" target="runtimes\win-x64\native" />
+		<file src="..\build\RelWithDebInfo\bin\RelWithDebInfo\ortextensions.dll" target="runtimes\win-x64\native" />
```
sayanshaw24: why move this from `out\Windows`? i would prefer to keep it in the `out/{platform}` format for consistency, and also avoid the nested RelWithDebInfo directories (for both the .dll and .lib)

tools/build.py
```diff
@@ -261,7 +261,7 @@ def path_from_env_var(env_var: str):
 
     parser.add_argument(
         "--cmake_generator",
-        choices=["Visual Studio 16 2019", "Visual Studio 17 2022", "Ninja", "Unix Makefiles", "Xcode"],
+        choices=["Visual Studio 16 2019", "Visual Studio 17 2022", "Visual Studio 18 2026", "Ninja", "Unix Makefiles", "Xcode"],
```
sayanshaw24, line 264: could you remove the VS 2026 here as well please?

amirkutcher:
@microsoft-github-policy-service agree company="Microsoft"

PR #1014, opened 2025-12-05
Return grid_thw output

apsonawane:
Qwen pre-processor was missing image_grid_thw output, added that

### DIFF
--- a/shared/api/image_transforms_qwen2_5.hpp
+++ b/shared/api/image_transforms_qwen2_5.hpp
@@ -14,7 +14,9 @@ class PatchImage {
  public:
   PatchImage() = default;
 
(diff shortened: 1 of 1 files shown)


shared/api/image_transforms_qwen2_5.hpp
```diff
@@ -59,6 +61,13 @@ class PatchImage {
     int64_t grid_h = H / patch_size_;
     int64_t grid_w = W / patch_size_;
 
+    // Populate grid_thw output tensor
+    grid_thw_output.Allocate({1, 3});
```
nenad1002, line 65: @sayanshaw24 Once you are back, lets discuss this allocation in C++ code, I see stuff like this over extensions, I don't really like this and then leaving the owner hanging to fix these, stuff like this causes memory corruption.

PR #1019, opened 2026-01-28
Fix win-arm64 python wheel suffix

chemwolf6922:
Fix win-arm64 python wheel suffix with similar logic to the win-amd64 fix. This will produce the correct win-arm64 wheels on woa machines with the following command:
```sh
set CIBW_ARCHS=ARM64
python -m cibuildwheel --platform windows --output-dir out .
```


chemwolf6922:
Hi @hanbitmyths, The CI seems to fail on a Linux build because of a build docker image preparation issue. Should I try retrigger the pipeline? Thanks.

apsonawane (approved):

PR #1027, opened 2026-02-20
Add Qwen3-VL support to Normalize image transform

hanbitmyths:
Qwen3-VL uses the same BGR→RGB channel swap during image normalization as Qwen2.5-VL. This PR adds qwen3_vl as a recognized key in the Normalize::Init attribute parser, enabling onnxruntime-genai to process Qwen3-VL models natively without aliasing to qwen2_5_vl.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -777,6 +777,12 @@ target_compile_definitions(ocos_operators PRIVATE ${OCOS_COMPILE_DEFINITIONS})
 target_link_libraries(ocos_operators PRIVATE ${ocos_libraries})
 
 file(GLOB _TARGET_LIB_SRC "shared/lib/*.cc")
+
+if(OCOS_ENABLE_AUDIO)
+  file(GLOB nemo_mel_SRC "shared/api/nemo_mel_*")
+  list(APPEND _TARGET_LIB_SRC ${nemo_mel_SRC})
+endif()
+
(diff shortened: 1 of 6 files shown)


test/pp_api_test/test_processor.cc
```diff
@@ -239,3 +239,42 @@ TEST(ProcessorTest, TestQwen2_5VLImageProcessing) {
...
+  err = OrtxGetTensorData(tensor.get(), reinterpret_cast<const void**>(&cpp_data), &shape, &num_dims);
+  ASSERT_EQ(err, kOrtxOK);
+
+  // Expect 3 Dimensions: [num_patches, patch_dim]
+  ASSERT_EQ(num_dims, 3ULL);
+
+  // Qwen3-VL uses patch_size=16, so patch_dim = 3 * 2 * 16 * 16 = 1536
+  // (vs Qwen2.5-VL's 3 * 2 * 14 * 14 = 1176)
+  int64_t patch_dim = shape[2];
+  ASSERT_EQ(patch_dim, 1536);
```
sayanshaw24, line 278: looks good! in the long term, we can also test MSE with HF reference like we do for Qwen2.5-VL: https://github.com/microsoft/onnxruntime-extensions/blob/main/test/pp_api_test/test_processor.cc#L230 to ensure pixel values are correct.

PR #1045, opened 2026-04-14
Fix ChatGLM3 BPE tokenization

sayanshaw24:
### Problem

ChatGLM3-6b's `tokenizer_class` is `"ChatGLMTokenizer"`, which the tokenizer type dictionary maps to `TokenType::kUnigram`. However, the model's `tokenizer.json` actually contains a **BPE** model (with vocab + merges). This mismatch caused the Unigram code path to be used, resulting in character-level tokenization — e.g., `"hello"` → `[h, e, l, l, o]` (5 tokens) instead of proper BPE sub-word tokens like `[▁hel, lo]` (2 tokens).

### Root Cause

Three issues were identified:

1. **Type mismatch**: `ChatGLMTokenizer` → `kUnigram` in `kTokenizerDict`, but `tokenizer.json` has `"model": {"type": "BPE"}`.
2. **No-op pre-tokenizer not handled**: ChatGLM3's pre-tokenizer is `Split` on a literal string `"<!dummy-prefix!>"` that never appears in normal text — effectively a no-op. The existing `LoadPreTokenizer` only parsed `pretokenizers` arrays (Sequence wrappers), not top-level Split objects, so this was silently ignored and the fallback LLAMA regex was used instead, incorrectly splitting input into word-level chunks.
3. **Incorrect `▁` prefix**: The Metaspace decoder (with `add_prefix_space: true`) was setting `add_dummy_prefix_ = true`, which prepends `▁` to the input during encoding. But for ChatGLM3, the Metaspace is only relevant for *decoding* — the no-op pre-tokenizer means no prefix should be added during encoding.

### Fix

- **`tokenizer_jsconfig.hpp`**: Added `PeekModelType()` to read `model.type` from `tokenizer.json` without full parsing.
- **`tokenizer_impl.cc`**: When `tokenizer_class` maps to `kUnigram` but `tokenizer.json` declares `"BPE"`, override the type to `kBPE`.
- **`bpe_tokenizer_model.hpp`**: Handle top-level `Split` pre-tokenizers with `String` patterns (no-op) by setting a `no_op_pretokenizer_` flag. Added `IsNoOpPretokenizer()` accessor.
- **`bpe_kernels.cc`**:
  - Route no-op pre-tokenizer models through the character-by-character BPE path (Gemma path) instead of the LLAMA regex splitter.
  - Detect `Metaspace` decoder type in `LoadSpmModelParams` to set `spm_model_` and `add_dummy_prefix_`.
  - Override `add_dummy_prefix_ = false` when the pre-tokenizer is a no-op, since the `▁` prefix should not be added during encoding.

### Validation

Added `ChatGLM3Tokenizer` test with exact parity against HuggingFace `tokenizers` library output:

| Input | HF Token IDs | Our Token IDs |
|---|---|---|
| `"hello"` | `[15616, 30914]` | ✅ Match |
| `"I like walking my cute dog"` | `[30936, 659, 5902, 552, 11527, 3246]` | ✅ Match |
| `"This is a test."` | `[3919, 323, 260, 1429, 30930]` | ✅ Match |

### Files Changed

| File | Change |
|---|---|
| `operators/tokenizer/tokenizer_jsconfig.hpp` | Added `PeekModelType()` method |
| `shared/api/tokenizer_impl.cc` | Type override from Unigram → BPE when tokenizer.json says BPE |
| `operators/tokenizer/bpe_tokenizer_model.hpp` | Handle no-op Split pre-tokenizer, added `IsNoOpPretokenizer()` |
| `operators/tokenizer/bpe_kernels.cc` | Metaspace detection, no-op routing, disable dummy prefix for no-op |
| `test/pp_api_test/test_tokenizer_impl.cc` | ChatGLM3 tokenizer test with HF parity checks |
| `test/data/chatglm/` | Test data (tokenizer.json, tokenizer_config.json) |


kunal-vaishnavi:
When I was working to get this fixed, these were the changes I had made to get `tokenizer.json` to load.

image

Why are these changes not needed now?

test/data/chatglm/tokenizer_config.json
```diff
@@ -0,0 +1,52 @@
+{
```
kunal-vaishnavi, line 1: How was `tokenizer.json` generated? If I remember correctly, it does not exist and requires a special converter process to create.

sayanshaw24:
> When I was working to get this fixed, these were the changes I had made to get `tokenizer.json` to load.
> Why are these changes not needed now?

Our approach detects the mismatch at runtime rather than hardcoding the type mapping. In `tokenizer_impl.cc`, when `tokenizer_class` maps to `kUnigram`, we peek at the actual `model.type` field in `tokenizer.json` — if it says `"BPE"`, we override the type to `kBPE`. This means any tokenizer class that's mapped to Unigram but ships a BPE tokenizer.json will work automatically, without needing to update `kTokenizerDict` for each new model.

The remaining fixes handle chatglm3's unusual pre-tokenizer (a no-op Split on a literal string) and the incorrect ▁ prefix that was being inferred from the Metaspace decoder config.

test/data/chatglm/tokenizer_config.json
sayanshaw24: The tokenizer.json was downloaded from the [amd/chatglm3-6b-awq-g128-int4-asym-bf16-onnx-ryzen-strix](https://huggingface.co/amd/chatglm3-6b-awq-g128-int4-asym-bf16-onnx-ryzen-strix) repo on HuggingFace, which already has a pre-converted tokenizer.json with the BPE model (64,789 vocab entries + 90,061 merges). The tokenizer_config.json was downloaded from [zai-org/chatglm3-6b](https://huggingface.co/zai-org/chatglm3-6b).

kunal-vaishnavi (approved):

PR #1046, opened 2026-04-16
PowerPC: Fix build failure without libpng-devel by switching to OpenCV 3rdparty libpng

BODAPATIMAHESH:
PowerPC port: Fix build failure in the absence of the libpng-devel package.

Replaced upstream libpng integration with OpenCV’s bundled 3rdparty libpng for PowerPC (ppc64/ppc64le) builds.

The upstream libpng setup required generating pnglibconf.h and enforcing strict header include ordering, which caused build failures in ORT Extensions when libpng-devel was not installed. Errors were primarily due to missing or incorrectly resolved headers (e.g., pnglibconf.h, pngpriv.h).

OpenCV’s vendored libpng provides a preconfigured pnglibconf.h and a simpler, self-contained layout. This removes the need for configure-time header generation and avoids fragile include path dependencies.


BODAPATIMAHESH:
Please review it @sayanshaw24

skottmckay:
OpenCV is a massive dependency to pull in. Does this limit what is configured/built in OpenCV to the bare minimum?

That said, I would have expected solving the setup issues with direct usage of libpng to be a better option than pulling in OpenCV.

BODAPATIMAHESH:
> OpenCV is a massive dependency to pull in. Does this limit what is configured/built in OpenCV to the bare minimum?
> 
> That said, I would have expected solving the setup issues with direct usage of libpng to be a better option than pulling in OpenCV.

No, this does not limit what can be configured in OpenCV.

I will open a new PR that removes the manual compilation of source files.

BODAPATIMAHESH:
Please review it https://github.com/microsoft/onnxruntime-extensions/pull/1051
@sayanshaw24 @skottmckay

PR #1047, opened 2026-04-17
AIX: Fix build failure

KV2773:
While building ort-extension on AIX(a big endian platform), has faced a little problems related to the macros, some apis, and some big endian test case issue failures.
So here are some changes to make its build successful on AIX.


KV2773:
@microsoft-github-policy-service agree company="IBM"

PR #1051, opened 2026-04-20
PowerPC: Fix build failure without libpng-devel

BODAPATIMAHESH:
PowerPC: fix build without libpng-devel by using upstream libpng CMake

1. Fix build failure on PowerPC when libpng-devel is not available
2. Use libpng’s native CMake build instead of manual source compilation for PowerPC
3. Retain bundled/manual libpng build as fallback for other architectures

### DIFF
--- a/cmake/ext_imgcodecs.cmake
+++ b/cmake/ext_imgcodecs.cmake
@@ -18,7 +18,7 @@ if(CMAKE_SYSTEM_PROCESSOR MATCHES "ppc64|ppc64le|powerpc")
   FetchContent_Declare(
     libpng_external
     GIT_REPOSITORY https://github.com/pnggroup/libpng.git
-    GIT_TAG        libpng18   # stable branch
+    GIT_TAG        v1.6.58   # stable branch
   )
 
   FetchContent_MakeAvailable(libpng_external)
--- a/cmake/ext_imgcodecs.cmake
+++ b/cmake/ext_imgcodecs.cmake
@@ -7,107 +7,101 @@ set(_IMGCODEC_ROOT_DIR ${dlib_SOURCE_DIR}/dlib/external)
 #  project libpng
 #
 # ----------------------------------------------------------------------------
+# Default PNG target
 set (PNG_LIBRARY "libpng_static_c")
 
-if(CMAKE_SYSTEM_PROCESSOR MATCHES "ppc64|ppc64le")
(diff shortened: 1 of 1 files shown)


cmake/ext_imgcodecs.cmake
```diff
@@ -7,107 +7,101 @@ set(_IMGCODEC_ROOT_DIR ${dlib_SOURCE_DIR}/dlib/external)
...
   message(STATUS "Using upstream libpng for PowerPC")
+
   include(FetchContent)
+
   FetchContent_Declare(
     libpng_external
     GIT_REPOSITORY https://github.com/pnggroup/libpng.git
-    GIT_TAG        v1.6.56   # latest version
-    )
+    GIT_TAG        libpng18   # stable branch
```
skottmckay, line 21: For security reasons it's better to use a specific version rather than a branch.

BODAPATIMAHESH: Thanks, Updated.

BODAPATIMAHESH:
@sayanshaw24 @skottmckay Could you review

BODAPATIMAHESH:
The PR hasn’t been approved yet because it’s failing the checks? Please share any review comments—this will help me address the issues and speed up the approval process @skottmckay

skottmckay (dismissed):
:shipit:

cmake/ext_imgcodecs.cmake
```diff
@@ -7,107 +7,101 @@ set(_IMGCODEC_ROOT_DIR ${dlib_SOURCE_DIR}/dlib/external)
...
-
-# Add architecture-specific PNG optimizations
-set(lib_srcs)
-
-# PowerPC VSX optimizations
-if(CMAKE_SYSTEM_PROCESSOR MATCHES "ppc64|ppc64le")
-  list(APPEND lib_srcs
-    ${libPNG_SOURCE_DIR}/powerpc/powerpc_init.c
-    ${libPNG_SOURCE_DIR}/powerpc/filter_vsx_intrinsics.c
-  )
```
skottmckay, line 53: I assume these optimizations are not lost with the change to use libpng_external. Is that the case?

BODAPATIMAHESH: Thanks. Yes, these optimizations are already included in libpng_external

BODAPATIMAHESH:
I have addressed the Copilot review comments @sayanshaw24

sayanshaw24 (approved):

PR #1052, opened 2026-04-22
Add Gemma 4 Preprocessing Support (Image, Audio, Text)

sayanshaw24:
# Add Gemma 4 Preprocessing Support (Image, Audio, Text)

## Summary

Adds image, audio, and text preprocessing support for [Gemma 4](https://huggingface.co/google/gemma-4-E2B-it) to ORT Extensions. This enables end-to-end multimodal inference pipelines for Gemma 4 in ORT GenAI and Foundry Local.

### Image preprocessing (`Gemma4ImageTransform`)
- Variable aspect-ratio resize preserving original proportions, fitting within `max_soft_tokens × pooling_kernel_size × patch_size` grid
- Rescale pixels to [0, 1] — Gemma 4 does **not** use ImageNet normalization (unlike most vision models)
- Patchification into `(num_patches, patch_size² × 3)` in HWC order within each patch
- 2D position IDs `(x, y)` per patch with `(-1, -1)` padding for unused slots
- Per-image `num_soft_tokens` count
- Config: `patch_size=16`, `max_soft_tokens=280`, `pooling_kernel_size=3`

### Audio feature extraction (`Gemma4LogMel`)
- USM-style 128-dimensional log-mel spectrogram
- 16kHz sampling rate, 20ms frames (320 samples), 10ms hop (160 samples), 512-point FFT
- Periodic Hann window, HTK mel filterbank (0–8kHz), `mel_floor=0.001`
- Semicausal padding (prepend `frame_length/2` zeros)
- Outputs `(num_frames, 128)` log-mel features + `(num_frames,)` boolean attention mask

### Tokenizer & chat template
- Gemma 4 uses `GemmaTokenizer` (BPE, 262144 vocab) — same family as Gemma 2/3
- v5-era file layout: separate `chat_template.jinja`, no inline `add_bos_token`, `backend: "tokenizers"`
- Chat template uses `<|turn>role\n...<turn|>\n` format (different from Gemma 3's `<start_of_turn>...<end_of_turn>`)
- Multimodal tokens: `<|image|>`, `<|audio|>`, `<|video|>`
- All tokenizer files downloaded from [google/gemma-4-E2B-it](https://huggingface.co/google/gemma-4-E2B-it)
- No production code changes needed — existing v5 support handles everything

### Bug fix: BOS inference for no-op `post_processor`
Gemma 4's `tokenizer.json` has a `post_processor` that's a no-op `TemplateProcessing` (sequence placeholders only, no special tokens). The existing v5 BOS inference code only applied per-class defaults when `post_processor` was **absent**. When present but empty of BOS/EOS references, the Gemma per-class default (`add_bos_token=true`) was never reached. Fixed by tracking whether BOS/EOS was actually inferred from the `post_processor`, and falling through to per-class defaults if not. Fully backward-compatible.

### Production code
| File | Change |
|------|--------|
| `shared/api/image_transforms_gemma4.hpp` | **NEW** — `Gemma4ImageTransform` kernel |
| `shared/api/gemma4_audio_features.hpp` | **NEW** — `Gemma4LogMel` kernel |
| `shared/api/image_processor.cc` | Register `Gemma4ImageTransform` in kernel registry |
| `shared/api/speech_extractor.cc` | Register `Gemma4LogMel` in kernel registry |
| `operators/tokenizer/bpe_kernels.cc` | Fix BOS inference when `post_processor` exists but doesn't mention BOS/EOS |

### Test data
| File | Description |
|------|-------------|
| `test/data/models/gemma-4/tokenizer.json` | BPE vocabulary (262144 tokens) from HF |
| `test/data/models/gemma-4/tokenizer_config.json` | GemmaTokenizer config (v5-era layout) |
| `test/data/models/gemma-4/chat_template.jinja` | 344-line Jinja template with tool calling support |
| `test/data/models/gemma-4/processor_config.json` | Reference processor config from HF |
| `test/data/models/gemma-4/image_processor.json` | Image pipeline config for tests |
| `test/data/models/gemma-4/audio_feature_extraction.json` | Audio pipeline config for tests |

### Tests
| Test | File | What it verifies |
|------|------|-----------------|
| `TestGemma4ImageProcessing` | `test_processor.cc` | Single image: shape `(1, 2520, 768)`, values in [0,1], position IDs, num_soft_tokens |
| `TestGemma4ImageProcessingMultiImage` | `test_processor.cc` | Batch of 2 images: batch dimension = 2 |
| `TestGemma4AudioFeatureExtraction` | `test_feature_extraction.cc` | Single audio: shape `(1, N, 128)`, finite values, attention mask |
| `TestGemma4AudioFeatureExtractionMultiFile` | `test_feature_extraction.cc` | Batch of 2 audio files: batch dimension = 2 |
| `Gemma4Tokenizer` | `test_tokenizer_impl.cc` | Tokenize + round-trip detokenize, BOS = 2 |
| `Gemma4ChatTemplate` | `test_tokenizer_chat.cc` | Text-only, image+text, system+user chat templates |

## Backward compatibility

- All existing tests pass unchanged
- The BOS inference fix is a strict superset of the previous behavior — models whose `post_processor` already contains BOS/EOS references are unaffected
- No changes to public APIs

## Reference
- HuggingFace model: https://huggingface.co/google/gemma-4-E2B-it
- Image processing reference: [`image_processing_gemma4.py`](https://huggingface.co/google/gemma-4-E2B-it/blob/main/image_processing_gemma4.py)
- Audio processing reference: [`feature_extraction_gemma4.py`](https://huggingface.co/google/gemma-4-E2B-it/blob/main/feature_extraction_gemma4.py)

### DIFF
--- a/operators/tokenizer/bpe_kernels.cc
+++ b/operators/tokenizer/bpe_kernels.cc
@@ -812,11 +812,15 @@ void JsonFastTokenizer::UpdateTokenizer(const TokenJsonConfig& config, const jso
   // tokenizer_config.json. When these flags were not explicitly set, try to
   // infer BOS/EOS behavior from the post_processor in tokenizer.json, or fall
   // back to per-class defaults for known tokenizer families.
-  // Track BOS and EOS inference separately so that finding only one in the
-  // post_processor doesn't suppress the per-class default for the other.
+  //
+  // Note: EOS inference is intentionally scoped inside the BOS-not-explicit
+  // block rather than being fully independent. In pre-v5 tokenizers (e.g.
+  // Mistral), add_bos_token is explicit but add_eos_token is simply absent
+  // (meaning "default false", not "infer from post_processor"). Making EOS
+  // inference independent would incorrectly enable EOS for those models.
+  // In v5, both flags are absent, so this block correctly handles both.
   if (!config.add_bos_token_explicit_ && !config.bos_token_.empty()) {
     bool bos_inferred = false;
-    bool eos_inferred = false;
     auto post_processor = tok_json.find("post_processor");
     if (post_processor != tok_json.end()) {
       std::string text = post_processor->dump();
@@ -827,7 +831,6 @@ void JsonFastTokenizer::UpdateTokenizer(const TokenJsonConfig& config, const jso
       if (!config.add_eos_token_explicit_ &&
           text.find(config.eos_token_) != std::string::npos) {
         add_eos_token_ = true;
-        eos_inferred = true;
       }
     }
 
--- a/operators/tokenizer/bpe_kernels.cc
+++ b/operators/tokenizer/bpe_kernels.cc
@@ -813,17 +813,22 @@ void JsonFastTokenizer::UpdateTokenizer(const TokenJsonConfig& config, const jso
   // infer BOS/EOS behavior from the post_processor in tokenizer.json, or fall
   // back to per-class defaults for known tokenizer families.
   if (!config.add_bos_token_explicit_ && !config.bos_token_.empty()) {
+    bool inferred_from_post_processor = false;
     auto post_processor = tok_json.find("post_processor");
     if (post_processor != tok_json.end()) {
       std::string text = post_processor->dump();
       if (text.find(config.bos_token_) != std::string::npos) {
         add_bos_token_ = true;
+        inferred_from_post_processor = true;
       }
       if (text.find(config.eos_token_) != std::string::npos) {
         add_eos_token_ = true;
+        inferred_from_post_processor = true;
       }
-    } else {
-      // No post_processor in tokenizer.json and no explicit config flags:
+    }
+
+    if (!inferred_from_post_processor) {
+      // post_processor is absent or doesn't mention BOS/EOS:
       // apply per-class defaults for known tokenizer families.
       // Note: GPT2-family models (Phi-4, Qwen2, DeepSeek, etc.) correctly
       // default to add_bos_token_=false so they don't need to be listed here.
--- a/operators/tokenizer/bpe_kernels.cc
+++ b/operators/tokenizer/bpe_kernels.cc
@@ -812,23 +812,27 @@ void JsonFastTokenizer::UpdateTokenizer(const TokenJsonConfig& config, const jso
   // tokenizer_config.json. When these flags were not explicitly set, try to
   // infer BOS/EOS behavior from the post_processor in tokenizer.json, or fall
   // back to per-class defaults for known tokenizer families.
+  // Track BOS and EOS inference separately so that finding only one in the
+  // post_processor doesn't suppress the per-class default for the other.
   if (!config.add_bos_token_explicit_ && !config.bos_token_.empty()) {
-    bool inferred_from_post_processor = false;
+    bool bos_inferred = false;
+    bool eos_inferred = false;
     auto post_processor = tok_json.find("post_processor");
     if (post_processor != tok_json.end()) {
       std::string text = post_processor->dump();
       if (text.find(config.bos_token_) != std::string::npos) {
         add_bos_token_ = true;
-        inferred_from_post_processor = true;
+        bos_inferred = true;
       }
-      if (text.find(config.eos_token_) != std::string::npos) {
+      if (!config.add_eos_token_explicit_ &&
+          text.find(config.eos_token_) != std::string::npos) {
         add_eos_token_ = true;
-        inferred_from_post_processor = true;
+        eos_inferred = true;
       }
     }
 
-    if (!inferred_from_post_processor) {
-      // post_processor is absent or doesn't mention BOS/EOS:
+    if (!bos_inferred) {
+      // post_processor is absent or doesn't mention BOS:
       // apply per-class defaults for known tokenizer families.
       // Note: GPT2-family models (Phi-4, Qwen2, DeepSeek, etc.) correctly
       // default to add_bos_token_=false so they don't need to be listed here.
--- a/shared/api/gemma4_audio_features.hpp
+++ b/shared/api/gemma4_audio_features.hpp
@@ -259,7 +259,9 @@ class Gemma4LogMel {
       for (int64_t fi = 0; fi < num_frames; ++fi) {
         float* mel_row = logmel_data.data() + fi * feature_size_;
         for (int64_t m = 0; m < feature_size_; ++m) {
-          mel_row[m] /= per_bin_stddev_[m];
+          if (per_bin_stddev_[m] != 0.0f) {
+            mel_row[m] /= per_bin_stddev_[m];
+          }
         }
       }
     }
--- a/shared/api/gemma4_audio_features.hpp
+++ b/shared/api/gemma4_audio_features.hpp
@@ -1,0 +1,346 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+
+#include <cmath>
+#include <vector>
+#include <complex>
+#include <algorithm>
+#include <numeric>
+
+#include "ext_status.h"
+#include "op_def_struct.h"
+
+#ifndef M_PI
+#define M_PI 3.14159265358979323846
+#endif
+
+namespace ort_extensions {
+
+// Gemma 4 audio feature extraction: USM-style log-mel spectrogram that matches
+// the HuggingFace Gemma4AudioFeatureExtractor exactly.
+//
+// Pipeline:  AudioDecoder  ->  Gemma4LogMel
+//
+// Inputs:   float (1, num_samples)  — mono PCM at `sampling_rate` Hz
+// Outputs:  float (num_frames, feature_size) — log-mel features
+//           bool  (num_frames,)              — frame-level attention mask
+class Gemma4LogMel {
+ public:
+  Gemma4LogMel() = default;
+
+  // ---------- HTK mel-frequency conversions --------------------------------
+  static double HzToMel(double hz) { return 2595.0 * std::log10(1.0 + hz / 700.0); }
+  static double MelToHz(double mel) { return 700.0 * (std::pow(10.0, mel / 2595.0) - 1.0); }
+
+  // Build an HTK-scale mel filterbank  (num_freq_bins x num_mel_filters).
+  static std::vector<float> MelFilterBank(int num_freq_bins, int num_mel_filters,
+                                          int sampling_rate,
+                                          double min_freq, double max_freq) {
+    const double mel_min = HzToMel(min_freq);
+    const double mel_max = HzToMel(max_freq);
+    const int n_points = num_mel_filters + 2;
+    std::vector<double> mel_points(n_points);
+    for (int i = 0; i < n_points; ++i) {
+      mel_points[i] = mel_min + (mel_max - mel_min) * i / (n_points - 1);
+    }
+
+    // Convert mel points back to Hz, then to FFT bin indices.
+    std::vector<double> freq_points(n_points);
+    for (int i = 0; i < n_points; ++i) {
+      freq_points[i] = MelToHz(mel_points[i]);
+    }
+
+    // Filterbank: shape (num_freq_bins, num_mel_filters), stored row-major.
+    std::vector<float> fb(static_cast<size_t>(num_freq_bins) * num_mel_filters, 0.0f);
+
+    auto hz_to_bin = [&](double hz) -> double {
+      return hz * (num_freq_bins - 1) * 2.0 / sampling_rate;
+    };
+
+    for (int m = 0; m < num_mel_filters; ++m) {
+      const double left = hz_to_bin(freq_points[m]);
+      const double center = hz_to_bin(freq_points[m + 1]);
+      const double right = hz_to_bin(freq_points[m + 2]);
+
+      for (int k = 0; k < num_freq_bins; ++k) {
+        double kd = static_cast<double>(k);
+        float w = 0.0f;
+        if (kd > left && kd <= center && center > left) {
+          w = static_cast<float>((kd - left) / (center - left));
+        } else if (kd > center && kd < right && right > center) {
+          w = static_cast<float>((right - kd) / (right - center));
+        }
+        fb[static_cast<size_t>(k) * num_mel_filters + m] = w;
+      }
+    }
+    return fb;
+  }
+
+  // ---------- Periodic Hann window -----------------------------------------
+  static std::vector<float> PeriodicHannWindow(int N) {
+    // w[n] = 0.5 - 0.5 * cos(2*pi*n / N)   for n = 0 .. N-1
+    std::vector<float> w(N);
+    for (int n = 0; n < N; ++n) {
+      double s = std::sin(M_PI * n / N);
+      w[n] = static_cast<float>(s * s);
+    }
+    return w;
+  }
+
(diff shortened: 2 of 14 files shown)


operators/tokenizer/bpe_kernels.cc
```diff
@@ -813,17 +813,22 @@ void JsonFastTokenizer::UpdateTokenizer(const TokenJsonConfig& config, const jso
...
     auto post_processor = tok_json.find("post_processor");
     if (post_processor != tok_json.end()) {
       std::string text = post_processor->dump();
       if (text.find(config.bos_token_) != std::string::npos) {
         add_bos_token_ = true;
+        inferred_from_post_processor = true;
       }
       if (text.find(config.eos_token_) != std::string::npos) {
         add_eos_token_ = true;
+        inferred_from_post_processor = true;
```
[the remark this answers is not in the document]
sayanshaw24: Addressed - gated EOS inference on `!config.add_eos_token_explicit_` so explicit config always takes precedence.

operators/tokenizer/bpe_kernels.cc
```diff
@@ -813,17 +813,22 @@ void JsonFastTokenizer::UpdateTokenizer(const TokenJsonConfig& config, const jso
...
       }
-    } else {
-      // No post_processor in tokenizer.json and no explicit config flags:
+    }
+
+    if (!inferred_from_post_processor) {
+      // post_processor is absent or doesn't mention BOS/EOS:
       // apply per-class defaults for known tokenizer families.
       // Note: GPT2-family models (Phi-4, Qwen2, DeepSeek, etc.) correctly
       // default to add_bos_token_=false so they don't need to be listed here.
```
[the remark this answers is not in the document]
sayanshaw24: Split into `bos_inferred`/`eos_inferred` so each is tracked independently.

test/pp_api_test/test_processor.cc
```diff
@@ -279,6 +279,128 @@ TEST(ProcessorTest, TestQwen3VLImageProcessing) {
...
+      found_padding = true;
+      // All subsequent positions should also be padding.
+      for (int64_t j = i + 1; j < kMaxPatches; ++j) {
+        EXPECT_EQ(pos_data[j * 2], -1);
+        EXPECT_EQ(pos_data[j * 2 + 1], -1);
+      }
+      break;
+    }
+  }
+  EXPECT_TRUE(found_padding) << "Image should not fill all patches exactly (expect some padding)";
```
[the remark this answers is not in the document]
sayanshaw24: Replaced with a deterministic check using `num_soft_tokens` * `pooling_kernel_size²` to compute the exact boundary.

shared/api/gemma4_audio_features.hpp
```diff
@@ -0,0 +1,346 @@
...
+      double ang = -2.0 * M_PI / len;
+      std::complex<double> wlen(std::cos(ang), std::sin(ang));
+      for (int i = 0; i < n_fft; i += len) {
+        std::complex<double> w(1.0, 0.0);
+        for (int j = 0; j < len / 2; ++j) {
+          auto u = x[i + j];
+          auto v = x[i + j + len / 2] * w;
+          x[i + j] = u + v;
+          x[i + j + len / 2] = u - v;
+          w *= wlen;
```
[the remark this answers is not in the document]
sayanshaw24: The FFT size is fixed at 512 so the per-frame cost is minimal for now. Will track as a follow-up if profiling shows it's a bottleneck.

operators/tokenizer/bpe_kernels.cc
```diff
@@ -812,18 +812,27 @@ void JsonFastTokenizer::UpdateTokenizer(const TokenJsonConfig& config, const jso
   // tokenizer_config.json. When these flags were not explicitly set, try to
   // infer BOS/EOS behavior from the post_processor in tokenizer.json, or fall
   // back to per-class defaults for known tokenizer families.
+  // Track BOS and EOS inference separately so that finding only one in the
+  // post_processor doesn't suppress the per-class default for the other.
   if (!config.add_bos_token_explicit_ && !config.bos_token_.empty()) {
```
apsonawane, line 822: This separation is gated inside this add_bos_token_explicit condition. BOS and EOS logic sits inside this, if this condition is false none of the code runs.

Shouldn't we have 2 separate conditions?
```
  if (!config.add_bos_token_explicit_ && !config.bos_token_.empty()){
}
```
```
 if (!config.add_eos_token_explicit_ &&
          text.find(config.eos_token_) != std::string::npos)
```

operators/tokenizer/bpe_kernels.cc
```diff
@@ -812,18 +812,27 @@ void JsonFastTokenizer::UpdateTokenizer(const TokenJsonConfig& config, const jso
...
       std::string text = post_processor->dump();
       if (text.find(config.bos_token_) != std::string::npos) {
         add_bos_token_ = true;
+        bos_inferred = true;
       }
-      if (text.find(config.eos_token_) != std::string::npos) {
+      if (!config.add_eos_token_explicit_ &&
+          text.find(config.eos_token_) != std::string::npos) {
         add_eos_token_ = true;
+        eos_inferred = true;
```
apsonawane: Is this variable getting used anywhere?

shared/api/gemma4_audio_features.hpp
```diff
@@ -0,0 +1,346 @@
...
+        for (int64_t m = 0; m < feature_size_; ++m) {
+          mel_row[m] -= per_bin_mean_[m];
+        }
+      }
+    }
+    if (!per_bin_stddev_.empty()) {
+      for (int64_t fi = 0; fi < num_frames; ++fi) {
+        float* mel_row = logmel_data.data() + fi * feature_size_;
+        for (int64_t m = 0; m < feature_size_; ++m) {
+          mel_row[m] /= per_bin_stddev_[m];
```
apsonawane: Division by zero is possible here?

shared/api/image_transforms_gemma4.hpp
```diff
@@ -0,0 +1,193 @@
...
+    const double factor = std::sqrt(target_px / total_px);
+
+    int64_t tgt_h = static_cast<int64_t>(std::floor(factor * src_h / side_mult)) * side_mult;
+    int64_t tgt_w = static_cast<int64_t>(std::floor(factor * src_w / side_mult)) * side_mult;
+
+    const int64_t max_side = (max_patches / (pooling_kernel_size * pooling_kernel_size)) * side_mult;
+
+    if (tgt_h == 0 && tgt_w == 0) {
+      // Both rounded to zero — fall back to the smallest possible square.
+      tgt_h = side_mult;
```
apsonawane: In hf code they are throwing an error: https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma4/image_processing_pil_gemma4.py#L61
Why are we falling back to smallest square?

shared/api/image_transforms_gemma4.hpp
```diff
@@ -0,0 +1,193 @@
...
+        // Copy patch pixels in raster order: for each (dy, dx) in patch, for
+        // each channel.  Layout: C contiguous blocks of (patch_size * patch_size).
+        for (int64_t ch = 0; ch < C; ++ch) {
+          for (int64_t dy = 0; dy < patch_size_; ++dy) {
+            const int64_t row = py * patch_size_ + dy;
+            for (int64_t dx = 0; dx < patch_size_; ++dx) {
+              const int64_t col = px * patch_size_ + dx;
+              uint8_t* pixel = reinterpret_cast<uint8_t*>(
+                  rgb_dst->image[row] + col * 4);
+              dst[ch * patch_size_ * patch_size_ + dy * patch_size_ + dx] =
```
apsonawane: Is this the correct shape of patch? In HF it looks like it is [height, width, channel] --> https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma4/image_processing_pil_gemma4.py#L94

But here it returns [channel, height, width]

operators/tokenizer/bpe_kernels.cc
sayanshaw24: Good catch. I initially split them into fully independent blocks, but that regressed `Nvidia_Mistral_Tokenizer` — Mistral has `add_bos_token: true` (explicit) but `add_eos_token` is simply absent (pre-transformers v5 convention = default false). An independent EOS block incorrectly inferred EOS from the post_processor for those models. The fix is to keep EOS inference scoped inside the BOS-not-explicit block, which is effectively the v5-detection boundary: in v5, both flags are absent so this block fires; in pre-v5, BOS is typically explicit so it's skipped. Added a detailed comment explaining the rationale. EOS inference is still gated on `!config.add_eos_token_explicit_` so explicit EOS config always wins (first Copilot comment).

shared/api/image_transforms_gemma4.hpp
sayanshaw24: Great catch. HF's `convert_image_to_patches` does `.permute(1, 3, 2, 4, 0)` which yields `(patch_h, patch_w, channels)` - HWC order within each patch. My code had channel as the outermost loop (CHW). Fixed to match HF: inner loop is now for dy → for dx → for ch, writing to offset (dy * patch_size + dx) * 3 + ch.

shared/api/gemma4_audio_features.hpp
sayanshaw24: Good catch. Added a zero-guard: `if (per_bin_stddev_[m] != 0.0f)` before the division. In practice Gemma 4's config has `per_bin_stddev: null` so this path isn't exercised today, but it's the right defensive thing to do.

operators/tokenizer/bpe_kernels.cc
sayanshaw24: Nope, good catch, dead code - removed. It was leftover from an intermediate refactor. Not needed since EOS doesn't have per-class defaults to fall through to (only BOS has the Gemma/Llama family list).

shared/api/image_transforms_gemma4.hpp
sayanshaw24: You're right — HF raises a ValueError rather than falling back. Changed to match: `GetAspectRatioPreservingSize` now returns {0, 0} when either dimension rounds to zero, and `Compute()` returns `kOrtxErrorInvalidArgument` with a descriptive message. Removed all the lenient fallback branches.

shared/api/image_transforms_gemma4.hpp
```diff
@@ -43,22 +43,10 @@ class Gemma4ImageTransform {
...
-    } else if (tgt_w == 0) {
-      tgt_w = side_mult;
-      tgt_h = std::min(
-          static_cast<int64_t>(std::floor(static_cast<double>(src_h) / src_w)) * side_mult,
-          max_side);
-      if (tgt_h == 0) tgt_h = side_mult;
+    // Match HuggingFace behavior: reject extreme aspect ratios that round
+    // a dimension to zero.  See image_processing_gemma4.py L61.
+    if (tgt_h == 0 || tgt_w == 0) {
+      return {0, 0};  // caller checks and returns an error
```
apsonawane, line 49: Why did we remove the other conditions those were correct. We just throw an error for ```(tgt_h == 0 || tgt_w == 0)```

sayanshaw24: Ah right, restored single-zero fallbacks!

apsonawane (dismissed):

shared/api/image_transforms_gemma4.hpp
```diff
@@ -0,0 +1,201 @@
...
+    float* pv = pixel_values_out.Allocate({max_patches, patch_dim});
+    std::memset(pv, 0, static_cast<size_t>(max_patches * patch_dim) * sizeof(float));
+
+    for (int64_t py = 0; py < ph; ++py) {
+      for (int64_t px = 0; px < pw; ++px) {
+        const int64_t patch_idx = py * pw + px;
+        float* dst = pv + patch_idx * patch_dim;
+        // Copy patch pixels in HWC raster order to match HuggingFace:
+        // permute(1, 3, 2, 4, 0) in convert_image_to_patches() yields
+        // (patch_height, patch_width, num_channels) within each patch.
```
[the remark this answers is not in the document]
sayanshaw24: Fixed.

shared/api/gemma4_audio_features.hpp
```diff
@@ -0,0 +1,348 @@
...
+    for (int64_t fi = 0; fi < num_frames; ++fi) {
+      const int64_t offset = fi * hop_length_;
+      const float* frame_start = padded.data() + offset;
+
+      // --- pre-emphasis (HTK flavour) or simple truncation -----------------
+      std::vector<float> processed(static_cast<size_t>(frame_length_));
+      if (preemphasis_ > 0.0f) {
+        if (preemphasis_htk_flavor_) {
+          processed[0] = frame_start[0] * (1.0f - preemphasis_);
+          for (int64_t i = 1; i < frame_length_; ++i) {
```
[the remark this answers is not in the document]
sayanshaw24: Fixed.

shared/api/gemma4_audio_features.hpp
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+    const int64_t num_samples = pcm_shape[1];
+    const float* pcm = pcm_input.Data();
+
+    // --- build window & filterbank on first call ---------------------------
+    if (window_.empty()) {
+      window_ = PeriodicHannWindow(static_cast<int>(frame_length_));
+      int fft_len = 1;
+      while (fft_len < frame_length_) fft_len <<= 1;
```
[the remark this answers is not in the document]
sayanshaw24: Fixed.

shared/api/image_transforms_gemma4.hpp
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    // --- patchify ---------------------------------------------------------
+    const int64_t ph = tgt_h / patch_size_;   // patches along height
+    const int64_t pw = tgt_w / patch_size_;   // patches along width
+    const int64_t num_patches = ph * pw;
+    const int64_t patch_dim = patch_size_ * patch_size_ * C;   // 16*16*3 = 768
+    const float rescale = 1.0f / 255.0f;
+
+    // Output 1: pixel_values  (max_patches, patch_dim), zero-padded
```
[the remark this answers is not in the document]
sayanshaw24: Fixed.

test/pp_api_test/test_processor.cc
```diff
@@ -279,6 +279,127 @@ TEST(ProcessorTest, TestQwen3VLImageProcessing) {
...
+  // Derive the expected number of real patches from num_soft_tokens.
+  // Each soft token maps to pooling_kernel_size^2 = 9 patches.
+  // Read num_soft_tokens early (output 2) to compute expected real patch count.
+  OrtxObjectPtr<OrtxTensor> nst_tensor;
+  err = OrtxTensorResultGetAt(result.get(), 2, nst_tensor.ToBeAssigned());
+  ASSERT_EQ(err, kOrtxOK);
+  const int64_t* nst_peek{};
+  const int64_t* nst_shape_peek{};
+  size_t nst_dims_peek{};
+  OrtxGetTensorData(nst_tensor.get(), reinterpret_cast<const void**>(&nst_peek), &nst_shape_peek, &nst_dims_peek);
```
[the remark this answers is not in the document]
sayanshaw24: Fixed.

apsonawane (approved):

justinchuby:
@copilot does Gemma4LogMel support batch inputs? Would it be more efficient if the second output is the audio sequence length instead of a boolean mask?

@sayanshaw24

PR #1054, opened 2026-04-24
Adding dlib's fft dependency

KV2773:
Using the build.sh as 

```
./build.sh -DCMAKE_POLICY_VERSION_MINIMUM=3.5   -DONNXRUNTIME_LIB_DIR=/usr/local/lib -DONNXRUNTIME_INCLUDE_DIR=/usr/local/include/onnxruntime  -DCMAKE_INSTALL_PREFIX=/usr/local -DOCOS_ENABLE_OPENCV_CODECS=OFF -DOCOS_ENABLE_VISION=OFF -DOCOS_ENABLE_AUDIO=OFF -DOCOS_ENABLE_CV2=OFF
```

Problem:
```
[ 96%] Linking CXX shared library lib/libortextensions.so
ld: 0711-317 ERROR: Undefined symbol: ._ZN4dlib3fftIfEEvRKNS_8fft_sizeEPKNSt3__17complexIT_EEPS7_b
ld: 0711-345 Use the -bloadmap or -bnoquiet option to obtain more information.
.ibm-clang: error: linker command failed with exit code 8 (use -v to see invocation)
gmake[2]: *** [CMakeFiles/extensions_shared.dir/build.make:109: lib/libortextensions.so] Error 8
gmake[1]: *** [CMakeFiles/Makefile2:639: CMakeFiles/extensions_shared.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....
/home/Kushal/onnxruntime-1.20.0/aixoss/onnxruntime-extensions/test/static_test/test_ustring.cc:32:29: warning: 'codecvt_utf8<char32_t>' is deprecated [-Wdeprecated-declarations]

```

After my changes.
```
[ 92%] Linking CXX executable bin/ocos_test
[100%] Built target ocos_test
[100%] Built target extensions_test
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -410,7 +410,9 @@ if(OCOS_ENABLE_MATH)
     file(GLOB TARGET_SRC_MATH_CUDA "operators/math/cuda/*.*")
     list(APPEND TARGET_SRC_MATH ${TARGET_SRC_MATH_CUDA})
   endif()
-  list(APPEND TARGET_SRC ${dlib_SOURCE_DIR}/dlib/fft/fft.cpp) # For the OCOS_ENABLE_AUDIO=OFF so that unknown symbol error shouldn't come.
+  if(NOT OCOS_ENABLE_AUDIO)
+    list(APPEND TARGET_SRC ${dlib_SOURCE_DIR}/dlib/fft/fft.cpp) # For the OCOS_ENABLE_AUDIO=OFF so that unknown symbol error shouldn't come.
+  endif()
   list(APPEND TARGET_SRC ${TARGET_SRC_MATH} ${TARGET_SRC_DLIB} ${TARGET_SRC_INVERSE})
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -377,7 +377,6 @@ if(OCOS_ENABLE_DLIB)
   # To avoid the unintentional using some unwanted component, we only include the test_for_odr_violations.cpp
   # to check if there is any violation in build configuration to ensure compiling some dlib source files correctly.
   file(GLOB TARGET_SRC_DLIB "${dlib_SOURCE_DIR}/dlib/test_for_odr_violations.cpp")
-  list(APPEND TARGET_SRC_DLIB ${dlib_SOURCE_DIR}/dlib/fft/fft.cpp) # For the OCOS_ENABLE_AUDIO=OFF so that unknown symbol error shouldn't come.
 endif()
 
 if(OCOS_ENABLE_TF_STRING)
@@ -411,7 +410,7 @@ if(OCOS_ENABLE_MATH)
     file(GLOB TARGET_SRC_MATH_CUDA "operators/math/cuda/*.*")
     list(APPEND TARGET_SRC_MATH ${TARGET_SRC_MATH_CUDA})
   endif()
-
+  list(APPEND TARGET_SRC ${dlib_SOURCE_DIR}/dlib/fft/fft.cpp) # For the OCOS_ENABLE_AUDIO=OFF so that unknown symbol error shouldn't come.
   list(APPEND TARGET_SRC ${TARGET_SRC_MATH} ${TARGET_SRC_DLIB} ${TARGET_SRC_INVERSE})
 endif()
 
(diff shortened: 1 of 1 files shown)


KV2773:
Hii  @sayanshaw24 @skottmckay  may you review it once.

skottmckay:
Does it need the functions in fft.cpp or just the definition of `fft_size`?

What code requires an `fft_size` definition and how is that dependency resolved with a default build?

skottmckay:
I would suggest adding something like this in the OCOS_ENABLE_MATH section before line 414 so that we only include the file when needed:

```cmake
  # operators\math\energy_stft_segmentation.cc needs the dlib fft implementation
  list(APPEND TARGET_SRC ${TARGET_SRC_DLIB} ${dlib_SOURCE_DIR}/dlib/fft/fft.cpp)
```
https://github.com/microsoft/onnxruntime-extensions/blob/a9ec8481735486d900af34dc2b78b7a226cb52da/CMakeLists.txt#L414

And now as multiple places could include fft.cpp we should remove duplicates at the end of configuration just before the `add_library` around line 5196 That's simpler than multiple places checking whether a file has been added previously or not.

https://github.com/microsoft/onnxruntime-extensions/blob/a9ec8481735486d900af34dc2b78b7a226cb52da/CMakeLists.txt#L516-L517

```
list(REMOVE_DUPLICATES TARGET_SRC_NOEXCEPTION)
list(REMOVE_DUPLICATES TARGET_SRC)
```

KV2773:
> Does it need the functions in fft.cpp or just the definition of `fft_size`?
> 
> What code requires an `fft_size` definition and how is that dependency resolved with a default build?

So it needs the definition of fft_size only as per my observation.

This function is used by these files
```grep_64: out/AIX/RelWithDebInfo/CMakeFiles/ocos_operators.dir/operators/math/energy_stft_segmentation.cc.o: binary file matches
grep_64: out/AIX/RelWithDebInfo/CMakeFiles/ocos_operators.dir/operators/math/math.cc.o: binary file matches
```

So in a default build we are making the OCOS_ENABLE_AUDIO by default on and in which we are adding this file fft.cpp which will give this definition.

```
if(OCOS_ENABLE_AUDIO)
  if (NOT OCOS_ENABLE_DLIB)
    message(FATAL_ERROR "Audio operators require DLIB to be enabled.")
  endif()
  include(dr_libs)
  file(GLOB TARGET_SRC_AUDIO "operators/audio/*.*")
  list(APPEND TARGET_SRC_AUDIO ${dlib_SOURCE_DIR}/dlib/fft/fft.cpp)
  list(APPEND TARGET_SRC_NOEXCEPTION ${TARGET_SRC_AUDIO})
endif()
```

KV2773:
> I would suggest adding something like this in the OCOS_ENABLE_MATH section before line 414 so that we only include the file when needed:
> 
> ```cmake
>   # operators\math\energy_stft_segmentation.cc needs the dlib fft implementation
>   list(APPEND TARGET_SRC ${TARGET_SRC_DLIB} ${dlib_SOURCE_DIR}/dlib/fft/fft.cpp)
> ```
> 
> https://github.com/microsoft/onnxruntime-extensions/blob/a9ec8481735486d900af34dc2b78b7a226cb52da/CMakeLists.txt#L414
> 
> And now as multiple places could include fft.cpp we should remove duplicates at the end of configuration just before the `add_library` around line 5196 That's simpler than multiple places checking whether a file has been added previously or not.
> 
> https://github.com/microsoft/onnxruntime-extensions/blob/a9ec8481735486d900af34dc2b78b7a226cb52da/CMakeLists.txt#L516-L517
> 
> ```
> list(REMOVE_DUPLICATES TARGET_SRC_NOEXCEPTION)
> list(REMOVE_DUPLICATES TARGET_SRC)
> ```

Yep @skottmckay that way, also we can do, and your approach is better in that way.
Should I commit the changes or you will commit them.

KV2773:
Hii @sayanshaw24  may u review it once.

KV2773:
Hii  @skottmckay @sayanshaw24 may u review it once.

KV2773:
Hii @skottmckay @sayanshaw24 , added the duplicity handling logic .
Can u review it once

KV2773:
Hii @skottmckay @sayanshaw24 , can u review it once, to check if everything is right and then can merge it so that it doesn't fail on other systems.

KV2773:
Heyy @skottmckay may you merge this PR if everything works.

PR #1055, opened 2026-04-27
OrtxDecodeAudio Support for PCM extraction

nenad1002:


shared/api/c_api_feature_extraction.cc
```diff
@@ -165,3 +171,64 @@ extError_t ORTX_API_CALL OrtxFeatureExtraction(OrtxFeatureExtractor* extractor,
...
+
+extError_t ORTX_API_CALL OrtxDecodeAudio(OrtxRawAudios* raw_audios, size_t index, int64_t target_sample_rate,
+                                         OrtxTensorResult** result) {
+  if (raw_audios == nullptr || result == nullptr) {
+    ReturnableStatus::last_error_message_ = "Invalid argument";
+    return kOrtxErrorInvalidArgument;
+  }
+
+  auto* audios_obj = static_cast<RawAudiosObject*>(raw_audios);
+  if (index >= audios_obj->num_audios_) {
```
sayanshaw24: is there a reason we only decode single raw audio (batch size 1) but pass in multiple? should we also support the scenario of decoding multiple, and thereby passing in multiple indices, and storing outputs in a result array? if not, why pass in multiple audios, would it not just cause unnecessary overhead?

**TLDR**, API design brainstorm - should we add batch decode support for audio?

This API takes an `OrtxRawAudios*` (which can hold N audios) but only decodes a single index. From a genai standpoint, if we need to decode every audio in a container, we'll cross the FFI boundary N times with repeated decoder init + memcpy each time.

`OrtxFeatureExtraction` already batch-processes all audios in one call using `Preprocess` over the full span. Could we add a batch variant (e.g., `OrtxDecodeAudios`) that returns a result per audio? The single-index version is fine for flexibility, but the batch variant would let genai decode all audios in one call: single decoder init, single boundary crossing, all decoding in C++.

If batch support isn't needed now, the single-index API would read more cleanly taking `const void* data, size_t size` directly rather than requiring the caller to construct an OrtxRawAudios container just to index into it.

shared/api/c_api_feature_extraction.cc
```diff
@@ -165,3 +171,64 @@ extError_t ORTX_API_CALL OrtxFeatureExtraction(OrtxFeatureExtractor* extractor,
...
+    attrs["target_sample_rate"] = target_sample_rate;
+  } else if (target_sample_rate == 0) {
+    // 0 = keep native rate: pass empty downsample list to disable resampling
+    attrs["target_sample_rates"] = std::vector<std::int64_t>{};
+  } else {
+    ReturnableStatus::last_error_message_ = "target_sample_rate must be >= 0";
+    *result = nullptr;
+    return kOrtxErrorInvalidArgument;
+  }
+  attrs["stereo_to_mono"] = int64_t{1};
```
sayanshaw24: can we avoid the hardcoded `stereo_to_mono` with no caller control?

this forces mono output with no way for the caller to opt out. if the intent is always-mono for the genai use case, the header doc should state the output is always mono. otherwise, let's consider exposing it as a parameter (default mono is fine).

shared/api/c_api_feature_extraction.cc
```diff
@@ -165,3 +171,64 @@ extError_t ORTX_API_CALL OrtxFeatureExtraction(OrtxFeatureExtractor* extractor,
...
+
+  // Pack results: [0] = pcm float32, [1] = sample rate int64
+  auto ts_result = std::make_unique<TensorResult>();
+  std::vector<std::unique_ptr<ortc::TensorBase>> tensors;
+  tensors.push_back(std::move(pcm_tensor));
+  tensors.push_back(std::move(sr_tensor));
+  ts_result->SetTensors(std::move(tensors));
+  *result = static_cast<OrtxTensorResult*>(ts_result.release());
+
+  return extError_t();
```
sayanshaw24: nit: for consistency, return the named success constant (`kOrtxOK`) explicitly.

shared/api/c_api_feature_extraction.cc
nenad1002: Stereo to mono 1 will almost certainly always be needed, but I can abstract it to make it more general

test/pp_api_test/test_decode_audio.cc
```diff
@@ -0,0 +1,184 @@
...
+  int16_t* dst = reinterpret_cast<int16_t*>(p);
+  const float two_pi = 2.0f * 3.14159265358979323846f;
+  for (int i = 0; i < n; ++i) {
+    float v = 0.5f * std::sin(two_pi * freq_hz * i / sample_rate);
+    dst[i] = static_cast<int16_t>(v * 32767.0f);
+  }
+  return wav;
+}
+
+TEST(DecodeAudioTest, BasicDecode) {
```
sayanshaw24, line 50: for test coverage, can we also add:
- stereo input (verifies the hardcoded mono mixing actually works)
- corrupt/truncated data (validates decoder-level error handling, not just argument validation)
- non-WAV formats (MP3/FLAC) if those are intended to be supported through this API

shared/api/c_api_feature_extraction.cc
```diff
@@ -165,3 +171,64 @@ extError_t ORTX_API_CALL OrtxFeatureExtraction(OrtxFeatureExtractor* extractor,
...
+  }
+
+  auto* audios_obj = static_cast<RawAudiosObject*>(raw_audios);
+  if (index >= audios_obj->num_audios_) {
+    ReturnableStatus::last_error_message_ = "Audio index out of range";
+    return kOrtxErrorInvalidArgument;
+  }
+
+  // Configure decoder
+  AudioDecoder decoder;
```
sayanshaw24: here we have a decoder re-initialized every call so every invocation creates a fresh `AudioDecoder` and calls `Init`. can we instead do something like `OrtxFeatureExtraction` which uses a pre-created extractor handle? if this is called in a loop from genai, the repeated init is avoidable overhead. we could either accept a pre-configured decoder handle or document that this is designed for one-shot use.

nenad1002: This is not a problem now as we added batch decode

sayanshaw24 (approved):

PR #1056, opened 2026-05-06
Fix Gemma4 image resize to use float-domain bicubic for HF parity

justinchuby:
## Problem

The `Gemma4ImageTransform` used Pillow-style RGB (uint8) resize with `ImagingResample`, which clamps bicubic interpolation results to [0,255] via `clip8()`. HuggingFace uses torchvision `F.resize` on float32 tensors where bicubic can produce values outside [0,1] (overshoot).

This caused:
- Max pixel diff of 0.03–0.13 between ort-extensions and HF
- Token divergence at position 15 in Gemma4 multimodal generation
- 53% of resized pixels differed by >0.001

## Fix

Resize each RGB channel independently using Pillow `"F"` (float32) mode instead of `"RGB"` (uint8). The float32 path uses the exact same bicubic kernel and antialias scaling as the uint8 path, but skips the `clip8()` clamping — allowing bicubic overshoot values outside [0,1] to match torchvision behavior.

**Empirical verification** (Python):
| Comparison | Max pixel diff |
|---|---|
| Torchvision vs PIL float (proposed) | **0.00001** ✅ |
| Torchvision vs PIL uint8 (current) | 0.134 ❌ |

## Changes

- `shared/api/image_transforms_gemma4.hpp`: Convert uint8 HWC input to 3 separate `"F"` mode float images, resize each with `ImagingResample` in float domain, patchify directly from float channel data
- `test/pp_api_test/test_processor.cc`: Update range assertion to allow bicubic overshoot; widen reference tolerance (1e-3 → 2e-3)

All 6 Gemma4 tests pass.


justinchuby:
@sayanshaw24 @apsonawane

shared/api/image_transforms_gemma4.hpp
```diff
@@ -9,18 +9,220 @@
 
 #include "ext_status.h"
 #include "op_def_struct.h"
-#include "image_resample.h"
 
 namespace ort_extensions {
 
+// ---------------------------------------------------------------------------
+// Float-domain 3-channel bicubic resize with antialias.
```
sayanshaw24: should these all be a shared utility?

other image processors (phi3, phi4, mllama) still use `ImagingResample`. if any of them hit the same uint8 clamping issue (e.g., for models that also compare against `torchvision` float behavior), they'd need to duplicate this code. we can maybe put `BicubicResizeU8ToFloatRGB` and the `detail::` helpers into a separate header like `image_resample_float.h` so other transforms can reuse them.

test/pp_api_test/test_processor.cc
```diff
@@ -430,7 +434,60 @@ TEST(ProcessorTest, TestGemma4ImageProcessingMultiImage) {
   ASSERT_EQ(shape[0], 2);  // batch of 2 images
 }
 
-// Security regression test: CMYK JPEG must be rejected at decode time (CWE-122).
+TEST(ProcessorTest, TestGemma4ImageResizeBenchmark) {
```
sayanshaw24, line 440: can we use slicing here to test just a sample of the data (similar to other vision/audio tests)?

this runs 20 iterations at up to 3024x4032 on every CI test run. that's potentially slow on resource-constrained CI machines and the output (`std::cout`) is informational only - there's no pass/fail assertion on timing.

test/pp_api_test/test_processor.cc
```diff
@@ -430,7 +434,60 @@ TEST(ProcessorTest, TestGemma4ImageProcessingMultiImage) {
...
+  std::cout << "[Benchmark] Gemma4 BicubicResizeU8ToFloatRGB (resize only, no decode)\n";
+  for (const auto& tc : cases) {
+    // Compute Gemma4 target size
+    auto [tgt_h, tgt_w] = ort_extensions::Gemma4ImageTransform::GetAspectRatioPreservingSize(
+        tc.h, tc.w, patch_size, max_patches, pooling_kernel_size);
+    if (tgt_h == 0 || tgt_w == 0) continue;
+
+    // Create random uint8 source
+    std::vector<uint8_t> src(static_cast<size_t>(tc.h) * tc.w * 3);
+    for (auto& v : src) v = static_cast<uint8_t>(rand() % 256);
```
sayanshaw24, line 472: nit: non-deterministic benchmark data

`rand()` without a fixed seed means results aren't reproducible across runs. for a benchmark this is fine (random pixel content is realistic), but seeding with a constant (e.g., `srand(42)`) would make results comparable across machines and runs if you ever want to track regression.

test/pp_api_test/test_processor.cc
```diff
@@ -317,26 +319,28 @@ TEST(ProcessorTest, TestGemma4ImageProcessing) {
...
+  // Patch 0 is in a uniform region, so uint8 and float-domain values match closely.
   // Patch 0, first 10 values (HWC order within each patch):
   const float kPatch0Expected[] = {
       0.18823531f, 0.05490196f, 0.01960784f,
       0.18823531f, 0.05490196f, 0.01960784f,
       0.18823531f, 0.05490196f, 0.01960784f,
       0.18823531f};
   for (int i = 0; i < 10; ++i) {
-    EXPECT_NEAR(pv_data[i], kPatch0Expected[i], 1e-3f)
+    EXPECT_NEAR(pv_data[i], kPatch0Expected[i], 2e-3f)
```
sayanshaw24, line 346: why does the tolerance widen if the fix improves parity?

the PR description states max diff vs HF dropped from 0.03–0.13 to 0.00001. the expected values in `kPatch0Expected` appear to be exact multiples of 1/255 (which makes sense since they're from a uniform region). if the new float-domain resize matches HF better overall, shouldn't these reference values be updated to the new HF-matching output rather than loosening tolerance on the old ones? or is this just because the reference values were captured from the uint8 path, and the float-domain path produces slightly different values even here? if it's the latter then can we document this for future reference?

shared/api/image_transforms_gemma4.hpp
```diff
@@ -9,18 +9,220 @@
...
+    }
+  }
+}
+
+}  // namespace detail
+
+// Bicubic resize from uint8 RGB to float RGB: (src_h, src_w, 3) uint8 → (tgt_h, tgt_w, 3) float.
+// Fuses the 1/255 rescale into the horizontal pass to avoid a full-resolution
+// float copy of the source image. Output may exceed [0,1] due to bicubic
+// overshoot (matching torchvision F.resize behavior).
```
sayanshaw24: the function comment says "Output may exceed [0,1] due to bicubic overshoot" - could you also document the expected output layout? i.e., caller must provide dst of size tgt_h * tgt_w * 3, output is HWC interleaved float RGB in [0,1] nominally with possible overshoot (or similar). this makes the contract clearer for future callers.

sayanshaw24:
thanks for the fix! looks mostly good, left some comments.

justinchuby:
Thanks for the thorough review @sayanshaw24! Addressed all comments in 03b7853:

1. **Shared utility** — Good idea. Left as follow-up since only Gemma4 uses float-domain resize currently. When other models need it, we can extract to `image_resample_float.h`.

2. **Benchmark iterations** — Reduced from 20→3 iterations, 3→1 warmup. Informational only, no assertions on timing.

3. **srand(42)** — Added fixed seed for reproducible benchmark data.

4. **Tolerance explanation** — Added comment explaining the ~2e-3 diff: reference values are exact 1/255 multiples from HF's uint8 path, while float-domain resize has slightly different coefficient precomputation vs Pillow's integer rounding.

5. **Output layout docs** — Added parameter documentation: `dst` buffer size requirement, HWC-interleaved float RGB layout, overshoot caveat.

justinchuby:
Let me push copilot to extract the utility

justinchuby:
Extracted the shared utility as suggested in d4b97e6:

**New file:** `shared/api/image_resample_float.h`
- Contains `BicubicResizeU8ToFloatRGB` and all `detail::` helpers
- Reusable by phi3, phi4, mllama, etc. if they hit the same uint8 clamping issue

**Updated:** `image_transforms_gemma4.hpp` now `#include "image_resample_float.h"` instead of inlining the code. Gemma4-specific class (`Gemma4ImageTransform`) remains in place. No functional change.

justinchuby:
Renamed all abbreviated variables to descriptive names per Google C++ style guide in d646950:

| Before | After |
|--------|-------|
| `src`, `dst` | `source`, `destination` |
| `kk`, `kk_h`, `kk_v` | `coefficients`, `horizontal_coefficients`, `vertical_coefficients` |
| `s0`, `s1`, `s2` | `sum_r`, `sum_g`, `sum_b` |
| `ww`, `w`, `k` | `weight_sum`, `weight`, `coeffs` |
| `p` | `pixel` |
| `in_h`, `in_w`, `out_w` | `input_height`, `input_width`, `output_width` |
| `src_h`, `src_w`, `tgt_h`, `tgt_w` | `source_height`, `source_width`, `target_height`, `target_width` |
| `ksize` | `kernel_size` |
| `xmin`, `xmax` | `x_start`, `x_count` |
| `xx`, `yy` | `output_x`, `output_y` |

Applied to both `image_resample_float.h` and `image_transforms_gemma4.hpp`. Doc comments updated to match.

sayanshaw24 (approved):

PR #1059, opened 2026-05-08
Support Sequence pre-tokenizer with sequential Split steps

tanzeel-amd:
# Support Sequence pre-tokenizer with sequential Split steps

## Problem

HuggingFace `tokenizer.json` files may specify a **Sequence** pre-tokenizer containing multiple `Split` steps followed by `ByteLevel` (e.g. [Hunyuan](https://huggingface.co/tencent/Hunyuan-1.8B-Instruct), GPT-OSS, Qwen 2.5). The existing code on `main` handles the Sequence loop by overwriting `pre_tokenizer_regex_` with each Split entry — only the **last** Split regex survives, silently discarding earlier steps.

For single-Split models (like GPT-OSS) this happened to work. For multi-Split models (like Hunyuan with 3 Split steps), earlier patterns were lost and tokenization produced incorrect results.

## Solution

Implement true sequential pre-tokenization matching the [HuggingFace tokenizers Rust library](https://github.com/huggingface/tokenizers):

- **[`Sequence::pre_tokenize`](https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/pre_tokenizers/sequence.rs#L43-L48)** — the `for` loop that runs each step sequentially on the same `PreTokenizedString`
- **[`Split::pre_tokenize`](https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/pre_tokenizers/split.rs#L82-L88)** — splits text using a regex with configurable behavior
- **[`Isolated` behavior](https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/tokenizer/normalizer.rs#L74-L88)** — keeps both matched and unmatched regions as separate chunks
- **[`Pattern::find_matches` for Regex](https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/tokenizer/pattern.rs#L56-L70)** — returns contiguous coverage of the entire input, nothing is dropped
- **[`ByteLevel` with `use_regex: false`](https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/pre_tokenizers/byte_level.rs#L112-L122)** — only does byte-to-unicode mapping, no GPT-2 regex splitting

## Changes

| File | What |
|------|------|
| `bpe_utils.hpp` | Add `SplitIsolated()` — splits text preserving both matched and unmatched regions. Add CJK/Hunyuan hardcoded matchers + pattern table entries. |
| `bpe_tokenizer_model.hpp` | Store Sequence Split steps in `sequence_steps_` vector instead of overwriting a single string. Add `NormalizeJsonRegexEscapes()` for CR/LF handling from JSON parsing. |
| `bpe_kernels.cc` | In `Tokenize()`, branch on `HasSequencePreTokenizer()`: Sequence models run the sequential pipeline; non-Sequence models use the existing single-regex path unchanged. `SpmTokenize()` is untouched. |
| `test_fast_tokenizer.py` | Add `test_hunyuan_split_sequence_tokenization` — builds a synthetic Hunyuan-style tokenizer (3 Split steps + ByteLevel) and verifies ORT matches HuggingFace across CJK, numbers, tabs, newlines, and mixed content. |

Non-Sequence tokenizers (GPT-2, Llama, Phi, CLIP, Falcon, etc.) are completely unaffected — `HasSequencePreTokenizer()` returns false and the existing code path runs.

## Testing

| Test | Result |
|------|--------|
| `test_gpt_oss_chat_template` (was failing) | PASS |
| `test_hunyuan_split_sequence_tokenization` (new) | PASS |
| `tencent/Hunyuan-1.8B-Instruct` (11 strings) | All match HuggingFace |
| Full test suite (local, Windows) | 155 passed, 35 skipped — zero regressions vs `main` |
| Models verified | GPT-2, GPT-OSS, Llama, Mistral, Phi-3/4, CLIP, BERT, RoBERTa, T5, Falcon, Qwen, DeepSeek, Whisper, OLMa, Hunyuan |

### DIFF
--- a/operators/tokenizer/bpe_kernels.cc
+++ b/operators/tokenizer/bpe_kernels.cc
@@ -258,11 +258,97 @@ std::vector<int64_t> KernelBpeTokenizer::Tokenize(ustring& input, int64_t max_le
 
   // Parse input
   auto special_token_split_res = bbpe_tokenizer_->SplitByAddedAndSpecial(input, added_tokens_);
+
+  // Helper lambda: byte-encode a pre-tokenized chunk, perform BPE, and append IDs to res.
+  auto process_bpe_chunk = [&](std::string utf8_token, size_t& offset, OffsetMappingType& offset_mapping) {
+    if (utf8_token.empty()) return;
+
+    int64_t space_dif = 0;
+    if (compute_offset_mapping) {
+      if (utf8_token[0] == ' ') {
+        offset++;
+        space_dif = -1;
+      }
+    }
+
+    std::list<std::pair<uint32_t, uint32_t>> byte_list;
+    std::string token_bytes;
+    token_bytes.reserve(utf8_token.size() * 2);
+    size_t token_len = utf8_token.length();
+    size_t end_diff = 0;
+    if (clean_up_spaces) {
+      utf8_token.erase(std::remove(utf8_token.begin(), utf8_token.end(), U' '), utf8_token.end());
+      token_len = utf8_token.empty() ? 0 : utf8_token.length() - 1;
+      if (utf8_token.empty()) return;
+    }
+
+    for (size_t i = 0; i < token_len; i++) {
+      token_bytes += unicode_byte_encoder_[static_cast<unsigned char>(utf8_token[i])];
+    }
+
+    if (clean_up_spaces) {
+      end_diff = token_bytes.length();
+      if (!utf8_token.empty()) {
+        token_bytes += unicode_byte_encoder_[static_cast<unsigned char>(utf8_token.back())];
+        token_bytes += "</w>";
+      }
+      end_diff = token_bytes.length() - end_diff;
+    }
+
+    auto id = bbpe_tokenizer_->GetTokenId(token_bytes);
+    if (id != bpe::kInvalidTokenId) {
+      byte_list.push_back(std::make_pair(id, ort_extensions::narrow<uint32_t>(utf8_token.size())));
+    } else {
+      token_len = token_bytes.length();
+      for (size_t i = 0; i < token_len - end_diff; /* i++ */) {
+        size_t j = ustring::UTF8Len(token_bytes[i]);
+        byte_list.push_back(std::make_pair(bbpe_tokenizer_->GetTokenId(token_bytes.substr(i, j)),
+                                           ort_extensions::narrow<uint32_t>(j)));
+        i += j;
+      }
+      if (end_diff > 0) {
+        byte_list.push_back(
+            std::make_pair(bbpe_tokenizer_->GetTokenId(token_bytes.substr(token_len - end_diff, end_diff)),
+                           ort_extensions::narrow<uint32_t>(end_diff)));
+      }
+    }
+
+    bbpe_tokenizer_->PerformBPE(byte_list);
+
+    for (auto p : byte_list) {
+      if (static_cast<int64_t>(res.size()) >= max_length) break;
+      res.push_back(p.first);
+      if (compute_offset_mapping) {
+        if (clean_up_spaces) {
+          offset_mapping.emplace_back(std::make_pair(offset, ort_extensions::narrow<size_t>(offset + p.second)));
+          offset += p.second;
+        } else {
+          auto adjusted = static_cast<size_t>(static_cast<int64_t>(p.second) + space_dif);
+          offset_mapping.emplace_back(std::make_pair(offset, ort_extensions::narrow<size_t>(offset + adjusted)));
+          offset += adjusted;
+        }
+      }
+    }
+  };
+
+  // Prepare pre-tokenizer: either single-regex or sequential pipeline
+  const bool use_sequence = bbpe_tokenizer_->HasSequencePreTokenizer();
+
   bpe::PreTokenizerWithRegEx reg_splitter;
-  // NOTE: the pattern was already validated on loading json file.
-  // safe to ingore the return value here.
-  auto status = reg_splitter.Compile(bbpe_tokenizer_->GetPreTokenizerRegex(ModelName(), bpe_conf_.get().spm_model_));
-  assert(status.IsOk());
+  std::vector<bpe::PreTokenizerWithRegEx> seq_splitters;
+
+  if (use_sequence) {
+    auto& steps = bbpe_tokenizer_->GetSequenceSteps();
+    seq_splitters.resize(steps.size());
+    for (size_t i = 0; i < steps.size(); ++i) {
+      auto status = seq_splitters[i].Compile(steps[i].regex);
+      assert(status.IsOk());
+    }
+  } else {
+    auto status = reg_splitter.Compile(
+        bbpe_tokenizer_->GetPreTokenizerRegex(ModelName(), bpe_conf_.get().spm_model_));
+    assert(status.IsOk());
+  }
 
   for (auto& seg_id : special_token_split_res) {
     if (static_cast<int64_t>(res.size()) >= max_length) break;
@@ -272,9 +358,7 @@ std::vector<int64_t> KernelBpeTokenizer::Tokenize(ustring& input, int64_t max_le
       continue;
     }
 
-    // Note: keep ptr to make sure the string_view is valid in the following process
     std::u32string str(seg_id.first);
-    reg_splitter.Set(str.c_str());
 
     size_t offset = 0;
     OffsetMappingType offset_mapping;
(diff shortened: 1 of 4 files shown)


sayanshaw24:
This PR's branch predates #1045, which landed on main and added a top-level Split pre-tokenizer handler, `no_op_pretokenizer_`, and the `spm_model` parameter to `GetPreTokenizerRegex`. Please rebase on main — the Sequence loop changes here conflict with that handler, and `GetPreTokenizerRegex` needs the `spm_model` parameter restored so SPM models continue to get the Llama regex.

operators/tokenizer/bpe_tokenizer_model.hpp
```diff
@@ -473,19 +465,28 @@ class BpeModel {
...
+    std::string base_regex;
+    if (model_name == "Llama") {
+      base_regex = bpe::PreTokenizerWithRegEx::LLAMA_REGEX_PATTERN;
+    } else {
+      base_regex = bpe::PreTokenizerWithRegEx::GPT2_REGEX_PATTERN;
     }
 
-    if (model_name == "Llama" || spm_model) {
-      return bpe::PreTokenizerWithRegEx::LLAMA_REGEX_PATTERN;
+    if (split_regexes_.empty()) {
```
sayanshaw24: In addition to the Copilot comment here, when `split_regexes_` is non-empty, the fused result always appends the base GPT-2 or Llama regex after the split patterns. Is that intentional for Hunyuan? If the split regexes fully define the pre-tokenization, the base regex may cause unexpected matches. Worth a brief comment explaining the rationale for always including the base regex as a fallback.

operators/tokenizer/bpe_tokenizer_model.hpp
```diff
@@ -99,18 +78,31 @@ class BpeModel {
       if (pre_type == "Split") {
         ORTX_JSON_RETURN_IF_NULL(&node, "pattern", iter_pattern);
         ORTX_JSON_RETURN_IF_NULL(iter_pattern, "Regex", regex_str);
-        pre_tokenizer_regex_ = regex_str->get<std::string>();
-        // Validate the regex pattern
+        auto regex = regex_str->get<std::string>();
+        // JSON decodes \r and \n into literal CR/LF bytes, but the Compile()
```
sayanshaw24: The CR/LF normalization only handles \r and \n, but JSON also decodes \t into a literal tab (and potentially other escape sequences). If a future model's Split pattern includes \t, the same substring-matching failure would occur. Consider also normalizing \t → \\t here, or using a more general approach.

operators/tokenizer/bpe_utils.hpp
```diff
@@ -642,6 +644,119 @@ class PreTokenizerWithRegEx {
...
+    if (!IsRN(m_text[0]) && !IsL(m_text[0]) && !IsP(m_text[0]) && !IsS(m_text[0])) {
+      i = 1;
+    }
+
+    if (i >= m_text.size() || !IsLM(m_text[i])) return {};
+
+    while (i < m_text.size() && IsLM(m_text[i])) {
+      i++;
+    }
+    if (i == 0) return {};
```
sayanshaw24: This if (i == 0) return {}; is unreachable. If the optional leading character is consumed, i >= 1 entering the while loop. If it isn't consumed, i == 0 and the guard on line 695 (if (i >= m_text.size() || !IsLM(m_text[i])) return {};) either returns early or ensures the while loop runs at least once (making i >= 1). We can remove this or replace with assert(i > 0).

operators/tokenizer/bpe_utils.hpp
```diff
@@ -642,6 +644,119 @@ class PreTokenizerWithRegEx {
     return res;
   }
 
+  static bool IsCJK(char32_t ch) {
+    return (ch >= 0x4E00 && ch <= 0x9FFF) ||   // CJK Unified Ideographs (一-龥 approx)
```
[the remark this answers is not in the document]
tanzeel-amd: Fixed. `IsCJK()` now uses `0x9FA5` as the upper bound for the CJK unified range, matching the registered regex branch [一-龥぀-ゟ゠-ヿ]+. This avoids accepting characters outside the regex range.

tanzeel-amd:
> This PR's branch predates #1045, which landed on main and added a top-level Split pre-tokenizer handler, `no_op_pretokenizer_`, and the `spm_model` parameter to `GetPreTokenizerRegex`. Please rebase on main — the Sequence loop changes here conflict with that handler, and `GetPreTokenizerRegex` needs the `spm_model` parameter restored so SPM models continue to get the Llama regex.

Rebased this branch on latest microsoft/main. The conflict with #1045 is resolved, and the newer top-level Split pre-tokenizer handling, no_op_pretokenizer_, and spm_model support are preserved.

operators/tokenizer/bpe_utils.hpp
tanzeel-amd: Fixed. Removed the unreachable guard.

operators/tokenizer/bpe_tokenizer_model.hpp
tanzeel-amd: Fixed with a general control-character normalization helper. JSON-decoded regex control bytes are converted back to escaped spellings before `Compile()` does substring matching:

_\r, \n, \t, \f,_ and _\v_ are normalized to _\\r, \\n, \\t, \\f,_ and _\\v_.

This keeps regex escapes like _\p{L}_ and _\s+_ untouched.

operators/tokenizer/bpe_tokenizer_model.hpp
tanzeel-amd: Yes, this is intentional. Hugging Face Sequence pre-tokenization with Split(..., behavior="Isolated") does not mean the Split regexes fully define the final token boundaries for all input. The Split steps isolate high-priority spans, then ByteLevel/BPE processing still has to handle the remaining text.

In ORT Extensions this path is implemented as one compiled regex-driven pre-tokenizer. **Prepending the Split regex branches gives them priority**, which is required for Hunyuan cases like \p{N}{1,3} and the CJK/punctuation branches. **Appending the model’s base GPT-2/Llama regex keeps coverage for all text** not isolated by those Split branches, including normal words, contractions, whitespace, and fallback punctuation behavior.

Without the base regex fallback, unmatched text would either be skipped until another Split branch matches or get fragmented incorrectly by the single-pass matcher, which can corrupt token IDs for ordinary text. The ordering is the important part: Split branches come first, so Hunyuan-specific boundaries win; the base regex only handles residual text.

VishalX:
Similar PR: https://github.com/microsoft/onnxruntime-extensions/pull/1064

operators/tokenizer/bpe_tokenizer_model.hpp
```diff
@@ -76,18 +78,31 @@ class BpeModel {
...
+        }
         bpe::PreTokenizerWithRegEx pre_tokenizer;
-        auto status = pre_tokenizer.Compile(pre_tokenizer_regex_);
+        auto status = pre_tokenizer.Compile(regex);
         if (!status.IsOk()) {
           return status;
         }
+        split_regexes_.push_back(std::move(regex));
+      } else if (pre_type == "ByteLevel") {
+        has_byte_level_in_sequence_ = true;
```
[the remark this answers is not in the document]
tanzeel-amd: Fixed. Removed the unused has_byte_level_in_sequence_ member/state.

VishalX:
@kunal-vaishnavi pls review.

operators/tokenizer/bpe_tokenizer_model.hpp
```diff
@@ -451,16 +466,27 @@ class BpeModel {
...
-    // by default, use the GPT2 pretokenizer regex
-    return bpe::PreTokenizerWithRegEx::GPT2_REGEX_PATTERN;
+    // Fuse Split pre-tokenizer patterns into the base regex by prepending them
+    // as higher-priority alternation branches. This ensures patterns like
+    // \p{N}{1,3} match before the base regex's \p{N}+ branch.
+    std::string fused;
+    for (const auto& sr : split_regexes_) {
+      if (!fused.empty()) fused += "|";
+      fused += sr;
+    }
```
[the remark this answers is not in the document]
tanzeel-amd: Added test based on [HunYuan model](https://huggingface.co/tencent/Hunyuan-1.8B-Instruct)'s tokeniser in `test/test_fast_tokenizer.py`.

sayanshaw24:
@tanzeel-amd @VishalX these changes introduce a tokenizer error in `test_gpt_oss_chat_template` inside `test_pp_api.py`, which is causing the CIs to fail:

```
test_pp_api.py:436: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = (<built-in function eq>, [200006, 17360, 200008, 3575, 553, 17554, ...], [200006, 17360, 200008, 3575, 553, 17554, ...])
kwds = {'err_msg': '', 'header': 'Arrays are not equal', 'strict': False, 'verbose': True}

    @wraps(func)
    def inner(*args, **kwds):
        with self._recreate_cm():
>           return func(*args, **kwds)
                   ^^^^^^^^^^^^^^^^^^^
E           AssertionError: 
E           Arrays are not equal
```

Please try to repro and fix locally with an agent that has context on this session, we can jump in if needed.

VishalX:
> @tanzeel-amd @VishalX these changes introduce a tokenizer error in `test_gpt_oss_chat_template` inside `test_pp_api.py`, which is causing the CIs to fail:
> 
> ```
> test_pp_api.py:436: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> 
> args = (<built-in function eq>, [200006, 17360, 200008, 3575, 553, 17554, ...], [200006, 17360, 200008, 3575, 553, 17554, ...])
> kwds = {'err_msg': '', 'header': 'Arrays are not equal', 'strict': False, 'verbose': True}
> 
>     @wraps(func)
>     def inner(*args, **kwds):
>         with self._recreate_cm():
> >           return func(*args, **kwds)
>                    ^^^^^^^^^^^^^^^^^^^
> E           AssertionError: 
> E           Arrays are not equal
> ```
> 
> Please try to repro and fix locally with an agent that has context on this session, we can jump in if needed.

Thanks @sayanshaw24. We'll test and fix this.

VishalX:
@sayanshaw24 could you pls help us with the instructions to run all the tests locally? If they are already in the repo, pls point us to it.

sayanshaw24:
> @sayanshaw24 could you pls help us with the instructions to run all the tests locally? If they are already in the repo, pls point us to it.

Here's how to build and run the tests locally:

```bash
# 1. Install the package in development/editable mode (from the repo root)
pip install -e .

# 2. Install dev/test dependencies
pip install -r requirements-dev.txt

# 3. Run the specific failing test
pytest test/test_pp_api.py::test_gpt_oss_chat_template -v

# Or run the full test suite
pytest test
```

More details are in [`docs/development.md`](https://github.com/microsoft/onnxruntime-extensions/blob/main/docs/development.md).

The editable install (`pip install -e .`) builds the C++ extensions in-place so Python changes take effect immediately. If you need debug symbols, set `debug=1` in [`setup.cfg`](https://github.com/microsoft/onnxruntime-extensions/blob/main/setup.cfg) before building.

The failing test is in [`test/test_pp_api.py`](https://github.com/microsoft/onnxruntime-extensions/blob/main/test/test_pp_api.py).

tanzeel-amd:
@sayanshaw24 This CI failure revealed a core flaw in our logic. Rewrote the core logic to follow HuggingFace Tokenizer.
Updated the PR description for details on solution.

operators/tokenizer/bpe_kernels.cc
```diff
@@ -286,85 +367,33 @@ std::vector<int64_t> KernelBpeTokenizer::Tokenize(ustring& input, int64_t max_le
...
-        if (!utf8_token.empty()) {
-          token_bytes += unicode_byte_encoder_[static_cast<unsigned char>(utf8_token.back())];
-          token_bytes += "</w>";
+    if (use_sequence) {
+      // Sequential pre-tokenization: run each Split step on the accumulated chunks
+      std::vector<std::u32string> chunks = {std::move(str)};
+      for (auto& splitter : seq_splitters) {
+        std::vector<std::u32string> new_chunks;
+        for (auto& chunk : chunks) {
+          auto sub = splitter.SplitIsolated(chunk);
```
sayanshaw24, line 379: `SplitIsolated` calls `Set(text)` which resets `m_utf8_text` for the STL regex fallback path. However, if a Split step's regex contains patterns that weren't matched by hardcoded matchers and fell into `fallback_patterns_`, and the input chunk is a `std::u32string` that was moved from a previous step's output — the object is re-initialized correctly per call, so this is fine. Just want to confirm: have you tested with a Split regex that triggers the STL regex fallback (i.e., a pattern not in the hardcoded pattern table)?

sayanshaw24:
the new CI failures are from a `ClipTokenizer` segfault - just looked into it and it's the same `size_t` underflow Copilot flagged. `SplitIsolated` can produce space-only chunks that become empty after `clean_up_spaces` stripping, causing length() - 1 to wrap. Both the empty-guard and the underflow fix are needed.

operators/tokenizer/bpe_kernels.cc
```diff
@@ -258,11 +258,94 @@ std::vector<int64_t> KernelBpeTokenizer::Tokenize(ustring& input, int64_t max_le
...
+
+  // Helper lambda: byte-encode a pre-tokenized chunk, perform BPE, and append IDs to res.
+  auto process_bpe_chunk = [&](std::string utf8_token, size_t& offset, OffsetMappingType& offset_mapping) {
+    size_t space_dif = 0;
+    if (compute_offset_mapping) {
+      if (utf8_token.at(0) == ' ') {
+        offset++;
+        space_dif = -1;
+      }
+    }
```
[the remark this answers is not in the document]
tanzeel-amd: added if (utf8_token.empty()) return;

tanzeel-amd:
Added empty-token guard, safe token_len calculation, and changed space_dif to int64_t. Also fixed a bug in SplitIsolated() where MatchWithSTLRegEx() could silently skip characters - added a test with a non-hardcoded regex pattern ([aeiou]+) to cover this path.

sayanshaw24 (approved):

PR #1063, opened 2026-05-15
Add native \p{M} (Unicode Mark) regex support for Qwen3.5 tokenizer

apsonawane:
## Summary

Adds hand-coded regex matchers for Qwen3.5's tokenizer pre-tokenization patterns that use `\p{M}` (Unicode Mark category). Without this change, these patterns fall through to `std::regex`, which does not support Unicode property escapes and crashes at runtime.

## Problem

Qwen3.5 is the only model family whose tokenizer regex includes `\p{M}`. Its full pre-tokenizer regex:
```
(?i:'s|'t|'re|'ve|'m|'ll|'d)|[^\r\n\p{L}\p{N}]?[\p{L}\p{M}]+|\p{N}| ?[^\s\p{L}\p{M}\p{N}]+[\r\n]|\s[\r\n]+|\s+(?!\S)|\s+
```

Two sub-patterns contain `\p{M}`:

| Sub-pattern | Purpose |
|---|---|
| `[^\r\n\p{L}\p{N}]?[\p{L}\p{M}]+` | Word matching (letters + combining marks like diacritics) |
| ` ?[^\s\p{L}\p{M}\p{N}]+[\r\n]*` | Punctuation/symbol matching (excludes marks from punctuation class) |

The remaining 4 sub-patterns already have existing matchers (LLAMA3, GPT2).

## Changes

**`operators/tokenizer/bpe_utils.hpp`**
- Added `IsLM()` helper — matches `[\p{L}\p{M}]`
- Added `NotLMNZ()` helper — matches `[^\s\p{L}\p{M}\p{N}]`
- Added `Match_Qwen35_Pattern_1()` — implements `[^\r\n\p{L}\p{N}]?[\p{L}\p{M}]+`
- Added `Match_Qwen35_Pattern_2()` — implements ` ?[^\s\p{L}\p{M}\p{N}]+[\r\n]*`
- Registered both patterns in the `Compile()` lookup table (before shorter LLAMA3 patterns to avoid shadowing)

**`test/pp_api_test/test_tokenizer_impl.cc`**
- Added `Qwen35RegexTest` — compiles the full Qwen3.5 regex and verifies tokenization of text with combining marks (e.g., `café` using U+0301), punctuation, digits, and newlines

### DIFF
--- a/operators/tokenizer/bpe_utils.hpp
+++ b/operators/tokenizer/bpe_utils.hpp
@@ -642,6 +642,51 @@ class PreTokenizerWithRegEx {
     return res;
   }
 
+  // "[^\r\n\p{L}\p{N}]?[\p{L}\p{M}]+"
+  std::u32string_view Match_Qwen35_Pattern_1() {
+
+    if (m_text.empty()) return {};
+
+    if ((!IsRN(m_text[0]) && !IsN(m_text[0])) || IsLM(m_text[0])) {
+      if (IsLM(m_text[0]) || ((m_text.size() > 1) && IsLM(m_text[1]))) {
+        size_t i = 1;
+        for (; i < m_text.size(); ++i) {
+          if (!IsLM(m_text[i])) break;
+        }
+        std::u32string_view res = m_text.substr(0, i);
+        m_text = m_text.substr(i);
+        return res;
+      }
+    }
+
+    return {};
+  }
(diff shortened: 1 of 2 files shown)


sayanshaw24 (approved):
looks great, thanks for adding this!

PR #1064, opened 2026-05-19
BPE: honour every Split in a Sequence pre_tokenizer

justinchuby:
## Problem

When the HuggingFace `pre_tokenizer` is a `Sequence` of multiple `Split` nodes, the loader at `bpe_tokenizer_model.hpp:96-115` overwrites `pre_tokenizer_regex_` on every iteration and only the **last** Split survives. All earlier regexes are silently dropped.

This breaks tokenisers such as [`tencent/Hy-MT1.5-1.8B-2bit`](https://huggingface.co/tencent/Hy-MT1.5-1.8B-2bit), whose pre-tokenizer is:

```json
{"type": "Sequence", "pretokenizers": [
  {"type": "Split", "pattern": {"Regex": "\\\\p{N}{1,3}"}, "behavior": "Isolated"},
  {"type": "Split", "pattern": {"Regex": "[一-龥぀-ゟ゠-ヿ]+"}, "behavior": "Isolated"},
  {"type": "Split", "pattern": {"Regex": "[!\"#$%&'()*+,\\\\-./:;<=>?@\\\\[\\\\\\\\\\\\]^_\`{|}~][A-Za-z]+|..."}, "behavior": "Isolated"},
  {"type": "ByteLevel", ...}
]}
```

## Fix

When every Split has `behavior: "Isolated"`, applying them in sequence is equivalent to a single alternation `p1|p2|...|pN` — each chunk of the final segmentation matches exactly one of the patterns. `PreTokenizerWithRegEx::Compile` already accepts that union (fast-paths recognised sub-patterns, falls back to `std::regex` for the rest), so we simply concatenate the Splits with `|` instead of overwriting.

Unsupported behaviors (`Removed`, `MergedWith*`) are now rejected explicitly rather than silently producing wrong tokens.

## Test

Added a regression test `test_hy_mt_sequence_of_splits` in `test/test_fast_tokenizer.py` that loads the Hy-MT1.5 tokenizer and compares ORT-Extensions output against HuggingFace fast-tokenizer output for both English and Chinese inputs.

## Validation

End-to-end verified with a mobius-exported Hy-MT1.5-1.8B-2bit ONNX model running through ORT GenAI:
- Before: `"The cat is sleeping on the chair."` → `"<6c>"` (single byte token)
- After: `"The cat is sleeping on the chair."` → `"El gato está durmiendo en la silla."`

### DIFF
--- a/operators/tokenizer/bpe_tokenizer_model.hpp
+++ b/operators/tokenizer/bpe_tokenizer_model.hpp
@@ -93,24 +93,56 @@ class BpeModel {
 
     ORTX_JSON_RETURN_IF_NULL(node_pre_tokenizer, "pretokenizers", iter_node_list);
 
+    // A HuggingFace Sequence pre_tokenizer applies each child Split in order, with the
+    // output chunks of step N being fed into step N+1. When every Split uses
+    // behavior == "Isolated" (the only behavior we model), the chain is equivalent to a
+    // single alternation `p1|p2|...|pN` evaluated left-to-right — each chunk in the
+    // final segmentation matches exactly one of the patterns. We therefore concatenate
+    // the regexes with `|` and hand the union to PreTokenizerWithRegEx::Compile, which
+    // both fast-paths recognised sub-patterns and falls back to std::regex for the rest.
+    //
+    // Previously this loop overwrote pre_tokenizer_regex_ on every iteration, silently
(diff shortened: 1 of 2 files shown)


sayanshaw24:
Hunyuan model support is now added and merged in this similar PR: https://github.com/microsoft/onnxruntime-extensions/pull/1059

PR #1073, opened 2026-06-10
setup.py: respect onnxruntime env variables

AlekseiNikiforovIBM:
setup.py: respect OCOS_ONNXRUNTIME_PKG_URI and OCOS_ONNXRUNTIME_VERSION env variables

With this change it is possible to build onnxruntime-extensions using setup.py with custom-built onnxruntime.


AlekseiNikiforovIBM:
Any feedback?

PR #1075, opened 2026-06-10
Fix out-of-bounds error

apsonawane:
This pull request adds input validation to the `RaggedTensorToSparse` operator to ensure the row-splits tensor is well-formed. These checks help catch malformed inputs early and improve error messaging.

Input validation improvements:

* Added a check to ensure the row-splits tensor is not empty, starts at zero, and is monotonic non-decreasing, with clear error messages for each case in `op_ragged_tensor.cc`.

### DIFF
--- a/operators/text/op_ragged_tensor.cc
+++ b/operators/text/op_ragged_tensor.cc
@@ -17,6 +17,18 @@ OrtStatusPtr RaggedTensorToSparse(const ortc::Tensor<int64_t>& n_element,
                                   .c_str(),
                               ORT_INVALID_ARGUMENT);
   int64_t n_els = d_length[0] - 1;
+
+  if (n_els < 0)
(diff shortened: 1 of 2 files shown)


operators/text/op_ragged_tensor.cc
```diff
@@ -17,6 +17,18 @@ OrtStatusPtr RaggedTensorToSparse(const ortc::Tensor<int64_t>& n_element,
                                   .c_str(),
                               ORT_INVALID_ARGUMENT);
   int64_t n_els = d_length[0] - 1;
+
+  if (n_els < 0)
```
kunal-vaishnavi, line 21: Can you add unit tests to validate these out-of-bound checks?

apsonawane: Added unit tests

kunal-vaishnavi (approved):

PR #1078, opened 2026-06-12
Fix audio decoder overflow issue

apsonawane:
This pull request improves the robustness of the audio decoder by adding input validation to prevent buffer overflows when detecting audio formats, and introduces a regression test to ensure this behavior. The main changes are as follows:

**Audio Decoder Input Validation:**

* The `ReadStreamFormat` method in `AudioDecoder` now takes a `data_size` parameter and checks that the input audio data is at least 4 bytes long before attempting to detect the format, returning an error if it is too short. This prevents a potential heap-buffer-overflow. [[1]](diffhunk://#diff-08390b7f241260344c129ac4c6bacbf4787511d6655e6bbebc4ca0cded5e0769L37-R38) [[2]](diffhunk://#diff-08390b7f241260344c129ac4c6bacbf4787511d6655e6bbebc4ca0cded5e0769R57-R60) [[3]](diffhunk://#diff-390d650ec8953da3b49abdeb4d315b847272bbe2cd08780a0ac1428ec57f00e7L44-R44) [[4]](diffhunk://#diff-08390b7f241260344c129ac4c6bacbf4787511d6655e6bbebc4ca0cded5e0769L108-R113)

**Testing Improvements:**

* Added a new test case `TooShortForFormatDetection` in `test_decode_audio.cc` to verify that audio data shorter than 4 bytes is rejected gracefully, preventing regressions related to buffer overflows in format detection.

### DIFF
--- a/operators/audio/audio_decoder.cc
+++ b/operators/audio/audio_decoder.cc
@@ -34,7 +34,8 @@ OrtStatusPtr AudioDecoder::OnModelAttach(const OrtApi& api, const OrtKernelInfo&
   return status;
 }
 
-AudioDecoder::AudioStreamType AudioDecoder::ReadStreamFormat(const uint8_t* p_data, const std::string& str_format,
+AudioDecoder::AudioStreamType AudioDecoder::ReadStreamFormat(const uint8_t* p_data, size_t data_size,
+                                                             const std::string& str_format,
                                                              OrtxStatus& status) const {
   const std::map<std::string, AudioStreamType> format_mapping = {{"default", AudioStreamType::kDefault},
                                                                  {"wav", AudioStreamType::kWAV},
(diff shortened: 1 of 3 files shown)


operators/audio/audio_decoder.cc
```diff
@@ -53,6 +54,10 @@ AudioDecoder::AudioStreamType AudioDecoder::ReadStreamFormat(const uint8_t* p_da
   }
 
   if (stream_format == AudioStreamType::kDefault) {
+    if (data_size < 4) {
```
kunal-vaishnavi, line 57: Why 4? Will this break for audio chunks that have no speech but are passed in or will this still work for those as well?

apsonawane: It will not break valid audio with no speech.
4 comes from format detection logic below, it reads 4-byte magic marker

kunal-vaishnavi (approved):

PR #1087, opened 2026-06-30
Fix audio processing input validation issues

sayanshaw24:
### Summary

This PR adds missing input validation to the audio processing pipeline to prevent heap buffer overflows and crashes from malformed or oversized inputs.

### Changes

**`shared/api/speech_features.hpp`**
- Fix heap OOB write in `LogMel::Compute()`: when audio input is longer than the model's configured chunk size, `mag_time` exceeds `expected_time` and the copy loop wrote past the output buffer. Added `std::min` clamp to copy at most `expected_time` frames per row, matching Python Whisper's `pad_or_trim` behavior.
- Fix NULL pointer dereference: moved the NULL check for `log_fbank.Allocate()` before the `std::memcpy` call that uses the returned pointer.

**`shared/api/nemo_mel_spectrogram.cc`**
- Replace `assert()` with a runtime bounds check in `NemoStreamingMelExtractor::Process()`. The assert validated that the output buffer was large enough but compiled to a no-op in release builds, allowing silent heap overflow. Now returns 0 frames if the buffer is insufficient.

**`operators/audio/audio_decoder.cc`**
- Add channel count validation in `DrReadFrames()`. The channel count parsed from audio file headers (WAV/FLAC/MP3) was used directly in a multiplication without bounds checking. An extreme channel count (e.g., 65535) caused OOM on 64-bit or integer overflow on 32-bit. Now rejects inputs with more than 16 channels.

**`test/pp_api_test/test_decode_audio.cc`**
- Fix pre-existing `MalformedAudioLogMelDoesNotCrash` test: used a WAV long enough to pass dlib's STFT minimum window requirement, and renamed to `ShortAudioLogMelDoesNotCrash` with corrected assertion (the pipeline pads short audio, so LogMel succeeds).
- Add `OversizedAudioLogMelTruncates` test: verifies 31-second audio is truncated to expected 3000 frames rather than overflowing.
- Add `ExtremeChannelCountRejected` test: crafts a WAV with 65535 channels and verifies graceful rejection.
- Fix MSVC `fopen` deprecation warning by using `fopen_s` on Windows.

### Testing

All 16 audio `pp_api_test` tests pass locally (Debug build, Windows).

### DIFF
--- a/operators/audio/audio_decoder.cc
+++ b/operators/audio/audio_decoder.cc
@@ -76,6 +76,12 @@ static size_t DrReadFrames(std::list<std::vector<float>>& frames, FX_DECODER fx,
   const size_t default_chunk_size = 1024 * 256;
   int64_t total_buf_size = 0;
 
+  // Reject unreasonable channel counts to prevent integer overflow in the
+  // multiplication below. Real audio never exceeds 8 channels (7.1 surround).
+  if (obj.channels == 0 || obj.channels > 16) {
+    return 0;
+  }
+
   for (;;) {
     std::vector<float> buf;
     buf.resize(default_chunk_size * obj.channels);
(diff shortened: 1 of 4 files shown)


themason2011 (commented):
Looks great!!!!

themason2011 (approved):

PR #1091, opened 2026-07-09
Add gemma-4-12B unified (encoder-free) preprocessing support

justinchuby:
## What

Adds the onnxruntime-extensions preprocessing needed for the **gemma-4-12B "unified"** model, which is *encoder-free*: it consumes raw merged pixel patches and raw waveform frames directly, rather than the SigLIP image / USM log-mel audio contract of the standard `gemma-4` (E2B/E4B) processor.

Today onnxruntime-genai can only run the unified model by preprocessing inputs with the HuggingFace `AutoProcessor` and feeding tensors via `Generator.set_inputs`, because there is no native transform matching the unified contract. This PR provides the native building blocks so genai can preprocess without HuggingFace.

## Changes

**Audio — new op `Gemma4UnifiedAudioFrames`** (`shared/api/gemma4_audio_features.hpp`, registered in `speech_extractor.cc`).
Chunks a decoded 16 kHz waveform into fixed `audio_samples_per_token` (default 640) frames, reproducing HF `Gemma4UnifiedAudioFeatureExtractor._extract_waveform_features` exactly: zero-pad to a whole number of frames, reshape to `(num_tokens, 640)`. No mel / windowing / normalization.
Pipeline: `AudioDecoder -> Gemma4UnifiedAudioFrames`.

**Image — no new op required.**
HF produces the unified 48px merged patches (`patch_dim = 48*48*3 = 6912`) via *(16px patchify → 3×3 `patches_merge`)*. That is **provably identical** to a direct 48px patchify (verified: both pixel values and position IDs match, and the aspect-ratio resize budget is identical). So the existing `Gemma4ImageTransform` yields the unified contract when configured with `patch_size=48, pooling_kernel_size=1, max_soft_tokens=280`. A new test locks this equivalence in.

**Tests / data**
- `test/data/models/gemma-4-unified/audio_feature_extraction.json` and `image_processor.json`.
- `ExtractorTest.TestGemma4UnifiedAudioFrames` — verifies the `(num_tokens, 640)` framing contract.
- `ProcessorTest.TestGemma4UnifiedImageProcessing` — verifies the 6912-dim contract, merged position IDs / padding, and cross-checks that the reused op's top-left 48px patch matches the standard config's 16px teacher patch on the same image.

## Testing

Built with `./build.sh -DOCOS_ENABLE_C_API=ON -DOCOS_ENABLE_CTEST=ON`. All 9 `*Gemma4*` tests pass (2 new + 7 existing, no regressions).


shared/api/speech_extractor.cc
```diff
@@ -17,7 +17,8 @@ Operation::KernelRegistry SpeechFeatureExtractor::kernel_registry_ = {
     {"NemoLogMel", []() { return CreateKernelInstance(&NemoLogMel::Compute); }},
     {"PerFeatureNormalize", []() { return CreateKernelInstance(&PerFeatureNormalize::Compute); }},
     {"Phi4AudioEmbed", []() { return CreateKernelInstance(&Phi4AudioEmbed::Compute); }},
-    {"Gemma4LogMel", []() { return CreateKernelInstance(&Gemma4LogMel::Compute); }}};
+    {"Gemma4LogMel", []() { return CreateKernelInstance(&Gemma4LogMel::Compute); }},
+    {"Gemma4UnifiedAudioFrames", []() { return CreateKernelInstance(&Gemma4UnifiedAudioFrames::Compute); }}};
```
sayanshaw24: do you think we can have a more generic audio processor for gemma? so say instead of having both `Gemma4LogMel` and `Gemma4UnifiedAudioFrames` kernels, we maybe have a `Gemma4Audio` kernel (or similar), that can take JSON options in order to determine which pass to use?

for instance, with the Qwen-2.5VL image processor, instead of adding a new op for Smart Resize, which the model uses, we added smart resize parameters to finetune the pre-existing Resize op.

justinchuby: Great suggestion — done in `2f42800`. Consolidated into a single `Gemma4Audio` op that selects the front-end via a `type` attribute:

- `type="log_mel"` (default) → 128-dim USM log-mel spectrogram (E2B/E4B)
- `type="raw_frames"` → raw 640-sample waveform frames (12B unified)

Both branches now share the same `(features: float, mask: bool)` output signature (the raw-frames path emits an all-true frame mask, which also matches HF's `Gemma4UnifiedAudioFeatureExtractor` returning `input_features` + `input_features_mask`). `Gemma4Audio` internally delegates to the unchanged `Gemma4LogMel` DSP and the raw-frames impl, so there's no duplicated logic.

One back-compat note: `Gemma4LogMel` already exists on `main` and is referenced by shipped gemma-4 configs, so I kept it registered as a thin alias (identical code path) rather than removing it — existing/deployed configs keep loading. The repo's own gemma-4 and gemma-4-unified test configs now both use `Gemma4Audio`. All 9 Gemma4 tests pass (log-mel via `type=log_mel`, raw frames via `type=raw_frames` incl. mask assertions).

I also propagated the op rename to the companion onnxruntime-genai (microsoft/onnxruntime-genai#2286) and mobius PRs. Let me know if you'd prefer I drop the `Gemma4LogMel` alias entirely (would be a breaking change for existing gemma-4 configs) or rename the attribute (`type` → e.g. `mode`).

shared/api/gemma4_audio_features.hpp
```diff
@@ -344,4 +347,156 @@ class Gemma4LogMel {
...
+  }
+
+  template <typename DictT>
+  OrtxStatus Init(const DictT& attrs) {
+    // Select the front-end from the ``type`` attribute, then forward the full
+    // attribute dict to the chosen implementation (each ignores the ``type``
+    // key). Defaults to log-mel for backward compatibility.
+    for (const auto& [key, value] : attrs) {
+      if (key == "type") {
+        const std::string& type = std::get<std::string>(value);
```
sayanshaw24: nice consolidation into Gemma4Audio! one robustness concern: `std::get` here (and similar `std::get` calls in the audio sub-paths) can throw bad_variant_access on malformed JSON, which bypasses `OrtxStatus` error handling. could we switch to a checked extraction path and return `kOrtxErrorInvalidArgument` instead?

shared/api/gemma4_audio_features.hpp
```diff
@@ -344,4 +347,156 @@ class Gemma4LogMel {
...
+    std::copy(pcm_input.Data(), pcm_input.Data() + num_samples, out);
+    // Every frame of a single clip is valid (padding lives within the last frame).
+    std::fill(mask, mask + num_tokens, true);
+    return {};
+  }
+
+  template <typename DictT>
+  OrtxStatus Init(const DictT& attrs) {
+    for (const auto& [key, value] : attrs) {
+      if (key == "audio_samples_per_token" || key == "feature_size") {
```
sayanshaw24: for `audio_samples_per_token` and `feature_size` aliasing: if both keys are provided with different values, the current behavior silently takes whichever appears last. can we explicitly detect and reject conflicting values? silent override here could hide config mistakes and produce unexpected frame shapes.

test/pp_api_test/test_feature_extraction.cc
```diff
@@ -446,4 +446,58 @@ TEST(ExtractorTest, TestGemma4AudioFeatureExtractionMultiFile) {
   ASSERT_EQ(err, kOrtxOK);
   ASSERT_EQ(mask_dims, 2ULL);
   ASSERT_EQ(mask_shape[0], 2);
-}
\ No newline at end of file
+}
+
+TEST(ExtractorTest, TestGemma4UnifiedAudioFrames) {
```
sayanshaw24, line 451: could we add a multi-file `raw_frames` test (e.g. two clips with different lengths) for `Gemma4Audio` with `type=raw_frames`? this would lock in batch stacking + mask behavior for the unified path, similar to the existing multi-file coverage on the log-mel side.

shared/api/gemma4_audio_features.hpp
justinchuby: Fixed in `920f54b`. Added a checked `GetTypedAttr` helper (`std::get_if`) and routed all attribute reads in `Gemma4LogMel`, `Gemma4UnifiedAudioFrames`, and `Gemma4Audio` through it, so a wrong-typed config value now returns `kOrtxErrorInvalidArgument` instead of throwing `std::bad_variant_access` past the `OrtxStatus` boundary.

shared/api/gemma4_audio_features.hpp
justinchuby: Fixed in `920f54b`. `Gemma4UnifiedAudioFrames::Init` now tracks `audio_samples_per_token` and `feature_size` separately and, if both are provided with different values, returns `kOrtxErrorInvalidArgument` ("they are aliases and must match") instead of silently taking the last one.

test/pp_api_test/test_feature_extraction.cc
justinchuby: Added `TestGemma4UnifiedAudioFramesMultiFile` in `920f54b`: two clips of different lengths (jfk.flac + 1272-141231-0002.wav). It asserts batch stacking to `(2, max_tokens, 640)`, and that each row's mask is a contiguous true-prefix / false-suffix — i.e. the shorter clip's padded tail is marked invalid while its real frames stay valid. This locks in the batch + mask behavior for the raw-frames path, mirroring the log-mel multi-file coverage.

test/pp_api_test/test_processor.cc
```diff
@@ -407,6 +407,124 @@ TEST(ProcessorTest, TestGemma4ImageProcessing) {
...
+  ASSERT_EQ(shape[2], kMergedPatchDim);
+
+  // position_ids — merged 48-grid coordinates.
+  err = OrtxTensorResultGetAt(result.get(), 1, tensor.ToBeAssigned());
+  ASSERT_EQ(err, kOrtxOK);
+  const int64_t* pos_data{};
+  err = OrtxGetTensorData(tensor.get(), reinterpret_cast<const void**>(&pos_data), &shape, &num_dims);
+  ASSERT_EQ(err, kOrtxOK);
+  EXPECT_EQ(pos_data[0], 0);  // patch 0 x
+  EXPECT_EQ(pos_data[1], 0);  // patch 0 y
```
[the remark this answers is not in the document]
justinchuby: Fixed in `ff3d01b`. Added rank/shape asserts on the position_ids tensor (rank 3, shape (1, max_soft_tokens, 2)) before any pos_data indexing.

test/pp_api_test/test_processor.cc
```diff
@@ -407,6 +407,124 @@ TEST(ProcessorTest, TestGemma4ImageProcessing) {
...
+  err = OrtxTensorResultGetAt(result.get(), 2, nst_tensor.ToBeAssigned());
+  ASSERT_EQ(err, kOrtxOK);
+  const int64_t* nst{};
+  const int64_t* nst_shape{};
+  size_t nst_dims{};
+  err = OrtxGetTensorData(nst_tensor.get(), reinterpret_cast<const void**>(&nst), &nst_shape, &nst_dims);
+  ASSERT_EQ(err, kOrtxOK);
+  EXPECT_EQ(nst[0], 260) << "merged soft-token count for australia.jpg (HF reference)";
+  const int64_t num_merged = nst[0];
+  // Last real merged patch position: (20-1, 13-1) = (19, 12).
```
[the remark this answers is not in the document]
justinchuby: Fixed in `ff3d01b`. The test now asserts the num_soft_tokens tensor rank/shape and that the value is within [1, max_soft_tokens] before it's used to index pos_data.

test/pp_api_test/test_feature_extraction.cc
```diff
@@ -446,4 +446,125 @@ TEST(ExtractorTest, TestGemma4AudioFeatureExtractionMultiFile) {
...
+  ASSERT_EQ(shape[0], 1);     // single audio
+  ASSERT_EQ(shape[2], 640);   // 640 raw samples per token
+  EXPECT_GT(shape[1], 0);     // at least one frame
+  const int64_t num_tokens = shape[1];
+
+  // All values finite and within the normalized PCM range.
+  for (int64_t i = 0; i < std::min<int64_t>(num_tokens * 640, 5000); ++i) {
+    ASSERT_TRUE(std::isfinite(data[i])) << "frame value at index " << i << " is not finite";
+    ASSERT_LE(std::abs(data[i]), 4.0f) << "frame value at index " << i << " out of range";
+  }
```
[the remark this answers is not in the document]
justinchuby: Fixed in `ff3d01b`. Tightened the bound to the actual normalized PCM range: decoded 16-bit flac/wav samples are in [-1, 1], so the assertion is now <= 1.0001 (small epsilon for full-scale float rounding) and the comment matches.

shared/api/gemma4_audio_features.hpp
```diff
@@ -18,10 +18,28 @@
...
+  return {};
+}
+
+}  // namespace gemma4_audio_detail
+
 // Gemma 4 audio feature extraction: USM-style log-mel spectrogram that matches
 // the HuggingFace Gemma4AudioFeatureExtractor exactly.
 //
-// Pipeline:  AudioDecoder  ->  Gemma4LogMel
+// Pipeline:  AudioDecoder  ->  Gemma4Audio (type="log_mel")
```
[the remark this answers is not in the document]
justinchuby: Fixed in `ff3d01b`. Reworded the doc block to say this is the log-mel implementation, kept registered under its own `Gemma4LogMel` name for backward compatibility and reused internally by `Gemma4Audio` with type="log_mel"; the pipeline line now reads `AudioDecoder -> Gemma4LogMel (or Gemma4Audio type="log_mel")`.

PR #1092, opened 2026-07-15
Fix OOB heap read in Gemma4LogMel per-bin normalization

sushraja-msft:
## Summary
Gemma4LogMel::Compute iterates m over [0, feature_size_) and reads per_bin_mean_[m] / per_bin_stddev_[m], but Init() populates `feature_size_` and the two vectors from independent JSON config attributes with no size cross-check. A config with e.g.  feature_size: 128 and per_bin_mean: [0.0] causes out-of-bounds std::vector<float>::operator[] reads past the allocation for every audio frame.).

## Fix
Validate the size invariant in Gemma4LogMel::Init immediately after the attribute loop: if per_bin_mean / per_bin_stddev is non-empty, its length must equal eature_size, otherwise return kOrtxErrorInvalidArgument. Rejecting the malformed config is preferable to a partial (semantically wrong) normalization.

## Details
- shared/api/gemma4_audio_features.hpp: added post-parse size validation for per_bin_mean_ and per_bin_stddev_.
- Added #include <string> for std::to_string.

### DIFF
--- a/shared/api/gemma4_audio_features.hpp
+++ b/shared/api/gemma4_audio_features.hpp
@@ -8,6 +8,7 @@
 #include <complex>
 #include <algorithm>
 #include <numeric>
+#include <string>
 
 #include "ext_status.h"
 #include "op_def_struct.h"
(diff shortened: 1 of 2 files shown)


sayanshaw24:
thanks for adding this! could you also add a regression test for malformed `Gemma4LogMel` configs where `per_bin_mean` or `per_bin_stddev` length do not match `feature_size` in this [test file](https://github.com/microsoft/onnxruntime-extensions/blob/main/test/pp_api_test/test_feature_extraction.cc). that would lock in the new validation and prevent the OOB read from coming back.

sayanshaw24 (approved):

PR #1096, opened 2026-07-25
Return an error for malformed JPEG input

### LINKED ISSUE #1094: DecodeImage is causing my program to exit
In `operators/vision/image_decoder.hpp` there is this line:
```c
# operators/vision/image_decoder.hpp : 137
cinfo.err = jpeg_std_error(&jerr); // sets what happens when libjpg has an error
```

The problem with this is that now `cinfo.err()`  will call `exit`:
```c
# libjpeg/src/jerror.c : 66
METHODDEF(void)
error_exit(j_common_ptr cinfo)
{
  /* Always display the message */
  (*cinfo->err->output_message) (cinfo);

  /* Let the memory manager delete any temp files before we die */
  jpeg_destroy(cinfo);

  exit(EXIT_FAILURE);
}
```

Now, I have data being fed to generic models, and in this case when the data is invalid and given to an ImageDecoder node, my entire program just exits, instead of having an OrtStatus being returned.

I am running this with onnxruntime C api and usually models will just return an OrtStatus when they have an error. So I'm wondering if there is a workaround for this s.t. invalid images don't crash my program.

If this were to be changed, it could be that some other function that is not `jpeg_std_error` is used, s.t. `cinfo.err` doesn't call `error_exit()`, but instead sets an OrtStatus.

ilia-sokolov:
## Summary

- replace libjpeg's process-terminating default fatal-error path with an `OrtxStatus`
- keep all libjpeg-mutated recovery state in heap storage so `setjmp`/`longjmp` does not leave modified automatic C++ state indeterminate
- clean up fully and partially initialized decompressors
- treat truncated input and marker skips beyond the supplied buffer as corrupt data
- prevent zero-scanline suspension returns from looping
- preserve useful bounded libjpeg diagnostics

## Root cause

The generic decoder installed `jpeg_std_error` without overriding `error_exit`. Fatal libjpeg errors therefore used the library's default handler, which can terminate the host process instead of returning an inference error.

## Validation

A Linux build with the portable libjpeg backend passed all six image-decoder tests:

- valid PNG and JPEG decoding
- malformed JPEG headers and oversized marker skips
- repeated truncation after JPEG output allocation
- successful valid JPEG decoding after repeated failures
- oversized PNG and JPEG rejection

`git diff --check` also passed.

Fixes #1094


sayanshaw24:
Hi, thanks for the contribution! Please address comments/CI failures.

PR #1097, opened 2026-07-27
Fix Llama transform heap overflow for non-square tile size

nenad1002:


shared/api/image_transforms_mllama.hpp
```diff
@@ -272,6 +272,12 @@ struct Llama3ImageTransform {
     auto [num_tiles_height, num_tiles_width] = aspect_ratio;
     auto padded_height = num_tiles_height * tile_size_.first;
     auto padded_width = num_tiles_width * tile_size_.second;
```
sayanshaw24: even with the new fit check, `padded_height` and `padded_width` are computed via signed multiplication and later used in allocation/memcpy sizing. a crafted config with very large values can overflow these products before the guard, which can re-open memory safety issues. could we add checked multiplication and total-size overflow validation before allocate/memset/memcpy?

sayanshaw24 (approved):

PR #1098, opened 2026-08-04
Fix OOB reads in tokenizer decoders, image transforms, and speech features

sayanshaw24:
### Summary

This PR adds missing bounds checks and input validation across multiple custom operators to prevent heap out-of-bounds reads from malformed inputs or crafted models.

### Changes

**`operators/tokenizer/bpe_decoder.hpp`**
- Fix wrong loop bound in `KernelBpeDecoder::Compute`: the inner loop iterated over the entire tensor element count instead of the per-row sequence length, causing OOB reads when processing batched inputs (batch >= 2).

**`operators/tokenizer/wordpiece_tokenizer.cc`**
- Fix misplaced bounds check in `KernelWordpieceTokenizer_Tokenizer`: the `row_index >= n_existing_rows` guard ran one line after the dereference of `existing_rows[row_index]`. Moved the check into the condition before the access.

**`operators/tokenizer/bert_tokenizer_decoder.cc`**
- Fix unbounded slice in `KernelBertTokenizerDecoder::Compute`: start/end indices from the attacker-controlled `positions` input were used to slice the `ids` buffer without any bounds validation. Added `0 <= start <= end <= ids.NumberOfElement()` check.

**`operators/tokenizer/trie_tokenizer.hpp`**
- Fix shape vector OOB read in `TrieDetokenizer::Compute`: the code assumed rank-2 input and accessed `ids_dim[1]` without verifying `ids_dim.size() >= 2`. Rank-1 inputs now treated as `[1, N]`.

**`shared/api/image_transforms.hpp`**
- Fix OOB read in `Resize::Compute`: added `c != 3` channel validation. The copy loop uses a hardcoded stride of 3, so non-3-channel inputs read past the buffer.
- Fix OOB read in `Normalize::Compute`: added `C == mean_.size()` validation. The channel loop indexed into fixed 3-element `mean_`/`std_` vectors using the input's channel dimension.
- Fix OOB read in `Permute3D::Init`: added validation that each `dims_[i]` is in `[0, 2]`. Out-of-range values would index past the 3-element shape vector.

**`shared/api/speech_features.hpp`**
- Fix OOB read in `SpeechLibSTFTNorm`: added validation that `fft_win_.size() >= frame_length_` after initialization. A caller-supplied `hann_win` shorter than `frame_length_` caused the STFT loop to read past the window buffer.

**`test/pp_api_test/test_tokenizer_impl.cc`**
- Add `BpeDecoderBatchedDetokenize` regression test: verifies batched detokenization (batch=2) produces correct output without OOB access.

### Testing

The `BpeDecoderBatchedDetokenize` test exercises the BPE decoder fix through the `TokenizerImpl::Detokenize` API. The remaining fixes are in custom op kernels or config-driven pipelines that require either a crafted ONNX model with an ORT session or attacker-controlled preprocessor config bundles, not worth adding into the repository.

### DIFF
--- a/operators/tokenizer/bpe_decoder.hpp
+++ b/operators/tokenizer/bpe_decoder.hpp
@@ -155,7 +155,7 @@ struct KernelBpeDecoder {
       std::string text;
       bool f_special_last = false;
       bool f_special = false;
-      auto count = static_cast<size_t>(ids.NumberOfElement());
+      auto count = static_cast<size_t>(seq_len);
 
       for (size_t tok_idx = 0; tok_idx < count; ++tok_idx) {
         const auto token = *(p_ids + tok_idx);
--- a/shared/api/image_transforms.hpp
+++ b/shared/api/image_transforms.hpp
@@ -340,7 +340,8 @@ struct Normalize {
     if (C != static_cast<int64_t>(mean_.size()) || C != static_cast<int64_t>(std_.size())) {
       return {kOrtxErrorInvalidArgument,
               "[Normalize]: channel count " + std::to_string(C) +
-              " does not match mean/std size " + std::to_string(mean_.size())};
+              " does not match mean/std size (mean=" + std::to_string(mean_.size()) +
+              ", std=" + std::to_string(std_.size()) + ")"};
     }
 
     float* out = output.Allocate({H, W, C});
@@ -479,7 +480,8 @@ struct Permute3D {
     }
 
     if (dims_.size() != 3 || dims_[0] < 0 || dims_[0] > 2 ||
-        dims_[1] < 0 || dims_[1] > 2 || dims_[2] < 0 || dims_[2] > 2) {
+        dims_[1] < 0 || dims_[1] > 2 || dims_[2] < 0 || dims_[2] > 2 ||
+        dims_[0] == dims_[1] || dims_[0] == dims_[2] || dims_[1] == dims_[2]) {
       return {kOrtxErrorInvalidArgument, "[Permute]: dims must be a permutation of {0, 1, 2}"};
     }
 
--- a/shared/api/image_transforms.hpp
+++ b/shared/api/image_transforms.hpp
@@ -155,6 +155,10 @@ struct Resize {
     int w = static_cast<int>(dimensions[1]);
     int c = static_cast<int>(dimensions[2]);
 
+    if (c != 3) {
+      return {kOrtxErrorInvalidArgument, "[Resize]: Expected 3-channel RGB input, got " + std::to_string(c) + " channels"};
+    }
+
     Imaging rgb_image = ImagingNew("RGB", w, h);
     for (int32_t i = 0; i < h; ++i) {
       for (int32_t j = 0; j < w; ++j) {
@@ -333,6 +337,12 @@ struct Normalize {
     int64_t W = dimensions[1];
     int64_t C = dimensions[2];
 
+    if (C != static_cast<int64_t>(mean_.size()) || C != static_cast<int64_t>(std_.size())) {
+      return {kOrtxErrorInvalidArgument,
+              "[Normalize]: channel count " + std::to_string(C) +
+              " does not match mean/std size " + std::to_string(mean_.size())};
+    }
+
     float* out = output.Allocate({H, W, C});
 
     if (!qwen2_5_vl_) {
@@ -466,6 +476,11 @@ struct Permute3D {
       } else {
         return {kOrtxErrorInvalidArgument, "[Permute]: Invalid argument"};
       }
+    }
+
+    if (dims_.size() != 3 || dims_[0] < 0 || dims_[0] > 2 ||
+        dims_[1] < 0 || dims_[1] > 2 || dims_[2] < 0 || dims_[2] > 2) {
+      return {kOrtxErrorInvalidArgument, "[Permute]: dims must be a permutation of {0, 1, 2}"};
     }
 
     return {};
--- a/shared/api/image_transforms.hpp
+++ b/shared/api/image_transforms.hpp
@@ -313,9 +313,15 @@ struct Normalize {
     for (const auto& [key, value] : attrs) {
       if (key == "mean") {
         auto mean = std::get<std::vector<double>>(value);
+        if (mean.size() < 3) {
+          return {kOrtxErrorInvalidArgument, "[Normalize]: mean must have at least 3 values"};
+        }
         mean_ = {static_cast<float>(mean[0]), static_cast<float>(mean[1]), static_cast<float>(mean[2])};
       } else if (key == "std") {
         auto std = std::get<std::vector<double>>(value);
+        if (std.size() < 3) {
+          return {kOrtxErrorInvalidArgument, "[Normalize]: std must have at least 3 values"};
+        }
         std_ = {static_cast<float>(std[0]), static_cast<float>(std[1]), static_cast<float>(std[2])};
       } else if (key == "qwen2_5_vl" || key == "qwen3_vl") {
         qwen2_5_vl_ = std::get<int64_t>(value) != 0;
--- a/shared/api/speech_features.hpp
+++ b/shared/api/speech_features.hpp
@@ -39,17 +39,17 @@ class SpeechFeatures {
       }
     }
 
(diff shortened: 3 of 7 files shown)


test/pp_api_test/test_tokenizer_impl.cc
```diff
@@ -944,4 +944,23 @@ TEST(OrtxTokenizerProfileTest, DISABLED_MemoryUsage) {
   fprintf(stderr, "Note: CPU-only heap memory (vocab + merges + lookup tables).\n");
   fprintf(stderr, "Allocated once at model load, freed when tokenizer is destroyed.\n");
   fprintf(stderr, "================================================================\n\n");
+}
+
+// Regression test: batched detokenization with batch >= 2 must not read past the
+// token ID buffer. Previously the inner loop used the total tensor element count
+// instead of the per-row sequence length as its bound.
+TEST(OrtxTokenizerTest, BpeDecoderBatchedDetokenize) {
```
apsonawane, line 952: Code changes are in KernelBpeDecoder but the test is with BpeDecoderBatchedDetokenize, so we should test that

operators/tokenizer/bpe_decoder.hpp
apsonawane, line 149: This can be 0 right for shape [2,0]?

test/pp_api_test/test_tokenizer_impl.cc
sayanshaw24: The Detokenize(batch) call exercises KernelBpeDecoder::Compute with a rank-2 [batch, seq_len] tensor internally — this is the code path that was fixed.

operators/tokenizer/bpe_decoder.hpp
sayanshaw24: Yes, seq_len=0 is safe - the inner loop simply runs 0 iterations and produces empty strings, which is correct for decoding zero tokens. No OOB or crash.

test/pp_api_test/test_tokenizer_impl.cc
```diff
@@ -944,4 +944,23 @@ TEST(OrtxTokenizerProfileTest, DISABLED_MemoryUsage) {
   fprintf(stderr, "Note: CPU-only heap memory (vocab + merges + lookup tables).\n");
   fprintf(stderr, "Allocated once at model load, freed when tokenizer is destroyed.\n");
   fprintf(stderr, "================================================================\n\n");
+}
+
+// Regression test: batched detokenization with batch >= 2 must not read past the
```
kunal-vaishnavi, line 949: From Copilot:

The BPE test exercises the intended batched code path through `TokenizerImpl::Detokenize`, which is good. However, it only checks that both output strings are equal. This would pass if both rows decode to the same incorrect result. Since the PR description says the test verifies correct output, add an explicit expected-value assertion such as `EXPECT_EQ(out_text[0], "This is a test.");` and check both rows. Also consider using different rows to prove that row boundaries are respected.

shared/api/image_transforms.hpp
```diff
@@ -333,6 +337,13 @@ struct Normalize {
     int64_t W = dimensions[1];
```
kunal-vaishnavi, line 343: From Copilot:

Could we validate the lengths of the `mean` and `std` vectors before indexing `[0]`, `[1]`, and `[2]` here? A malformed configuration with fewer than three values can cause an out-of-bounds read during `Init`, before the channel-count check in `Compute` runs. Please return an invalid-argument error for undersized vectors, while retaining the `C` check in `Compute` for tensor/config compatibility.

shared/api/speech_features.hpp
```diff
@@ -46,6 +46,13 @@ class SpeechFeatures {
         fft_win_ = hann_window(frame_length_);
       }
     }
+
+    if (static_cast<int64_t>(fft_win_.size()) < frame_length_) {
```
kunal-vaishnavi, line 55: From Copilot:

Could we validate the speech configuration more comprehensively here? Checking only that `fft_win_.size() >= frame_length_` still permits `frame_length_ == 1`, but `SpeechLibSTFTNorm` unconditionally accesses `y_frames_prev(r, 1)`, which would read out of bounds. Zero or negative frame lengths, as well as invalid `n_fft_` and `hop_length_` values, should also be rejected during initialization before computation begins. At minimum, please require `frame_length_ >= 2` and validate the related FFT and hop parameters.

test/pp_api_test/test_tokenizer_impl.cc
sayanshaw24: done!

shared/api/speech_features.hpp
sayanshaw24: done!

shared/api/image_transforms.hpp
sayanshaw24: done!

shared/api/speech_features.hpp
```diff
@@ -46,6 +46,18 @@ class SpeechFeatures {
         fft_win_ = hann_window(frame_length_);
```
kunal-vaishnavi, line 51: From Copilot:

Could we move the validation of `frame_length_`, `hop_length_`, and `n_fft_` before this block?

With a malformed configuration, a negative `frame_length_` reaches `hann_window()` or `hamming_window()` first. The resulting negative `int` can be converted to a very large `size_t` when constructing the vector, potentially causing a huge allocation or `std::bad_alloc` before the later validation runs. Please reject these values before generating the default window, then keep the existing `fft_win_` length check afterward.

shared/api/speech_features.hpp
```diff
@@ -46,6 +46,18 @@ class SpeechFeatures {
...
+    }
+
+    if (static_cast<int64_t>(fft_win_.size()) < frame_length_) {
+      return {kOrtxErrorInvalidArgument,
+              "[AudioFeatures]: hann_win size (" + std::to_string(fft_win_.size()) +
+              ") is smaller than frame_length (" + std::to_string(frame_length_) + ")"};
+    }
+
     return {};
   }
```
kunal-vaishnavi, line 63: From Copilot:

Could we validate the PCM tensor shape and length before computing `n_batch` here?

`SpeechLibSTFTNorm` assumes a rank-2 tensor and that `pcm_length >= frame_length_`. With malformed or too-short input, `(pcm_length - frame_length_) / hop_length_ + 1` can become non-positive, which may lead to invalid matrix dimensions or an exception instead of a clean `ORT_INVALID_ARGUMENT` response.

Please add validation for the expected PCM rank and minimum length before calculating `n_batch`.

test/pp_api_test/test_tokenizer_impl.cc
```diff
@@ -944,4 +944,25 @@ TEST(OrtxTokenizerProfileTest, DISABLED_MemoryUsage) {
...
+// token ID buffer. Previously the inner loop used the total tensor element count
+// instead of the per-row sequence length as its bound.
+TEST(OrtxTokenizerTest, BpeDecoderBatchedDetokenize) {
```
kunal-vaishnavi, line 952: From Copilot:

Could we assert the exact expected output for the second row instead of only checking that it differs from the first?

`EXPECT_NE(out_text[0], out_text[1])` would pass for any arbitrary, incorrect string as long as it is different. Please assert the expected detokenized value for `out_text[1]`—for example, `"Hello world!"`—or compare both rows against known tokenizer-generated outputs.

shared/api/speech_features.hpp
sayanshaw24: done.

shared/api/speech_features.hpp
sayanshaw24: nit - this is a valid hardening suggestion but it's in SpeechLibSTFTNorm::Compute, a different function from the Init-time validation this PR fixes. Happy to address in a follow-up, but it's out of scope for this bounds-check PR.

test/pp_api_test/test_tokenizer_impl.cc
sayanshaw24: Already addressed in the latest commit; the test now uses two different rows with EXPECT_EQ(out_text[0], \"This is a test.\") and EXPECT_NE to verify row boundaries.

apsonawane (approved):

PR #1099, opened 2026-08-06
Fix Marian Unicode case restoration

mustjab:
## Problem

Marian tokenization through ORT Extensions produces the same token IDs as SentencePiece, but detokenization can lose capitalization for non-ASCII scripts. In downstream translation package validation, all 36 language directions matched SentencePiece token IDs exactly (553/553 cases per direction), while decoded text was lowercased where SentencePiece restored the original capitalization. This blocks the strict tokenizer quality gate and would otherwise ship visibly degraded translation output.

For example, Marian's case encoder lowercases text and records the original casing with `T`, `U`, and `A` markers. The decoder sees those markers, but currently decides whether to apply them with `std::iswalpha` and maps case with `std::towupper`. Those APIs depend on the process locale. Under the C locale commonly used by hosted Linux and Windows environments, Bashkir and other non-ASCII letters may not be recognized as letters, so the marker is ignored and the output remains lowercase.

## Fix

- classify letters with ORT Extensions' locale-independent Unicode category table
- restore uppercase and titlecase with its Unicode case mappings
- use `char32_t` for decoded codepoints so behavior does not depend on the platform width of `wchar_t`
- preserve the existing Marian marker state machine and leave non-Marian UGM decoding unchanged

This does not change encoded token IDs. It fixes Marian decoding so its case markers are applied consistently across locales and scripts.

## Regression coverage

- verifies Bashkir and accented Latin capitalization is restored
- verifies unmarked lowercase Bashkir and Latin remain lowercase
- verifies uncased text, punctuation, and digits remain unchanged
- retains the existing cross-piece, mid-piece, and marker-reset regressions

## Testing

- `pp_api_test.exe --gtest_filter=MarianId2TokenTest.*` (6 passed)
- `pp_api_test.exe --gtest_filter=*Tokenizer*:*Ugm*:*UGM*` (61 passed)

### DIFF
--- a/test/pp_api_test/test_tokenizer_capi.cc
+++ b/test/pp_api_test/test_tokenizer_capi.cc
@@ -373,6 +373,22 @@ TEST_F(MarianId2TokenTest, CombinedBugs) {
             "THIS iPhone costs PPV-mp only");
 }
 
+// Non-ASCII letters must not depend on the process locale. Hosted Linux and
+// Windows test environments commonly use the C locale, where iswalpha and
+// towupper only handle ASCII reliably.
+TEST_F(MarianId2TokenTest, UnicodeCaseRestoration) {
+  ASSERT_EQ(tokenizer_.Code(), kOrtxOK) << "Failed to create tokenizer.";
+  EXPECT_EQ(RoundTrip(u8"Башҡортостан Республикаһы"),
+            u8"Башҡортостан Республикаһы");
+  EXPECT_EQ(RoundTrip(u8"École Über"), u8"École Über");
+}
+
+TEST_F(MarianId2TokenTest, UnicodeCasePreservesUnmarkedText) {
+  ASSERT_EQ(tokenizer_.Code(), kOrtxOK) << "Failed to create tokenizer.";
+  EXPECT_EQ(RoundTrip(u8"башҡорт теле; école über; 中文 123"),
+            u8"башҡорт теле; école über; 中文 123");
+}
+
 // ============================================================================
 // Transformers v5 format tests
 // ============================================================================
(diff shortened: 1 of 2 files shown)


sayanshaw24 (dismissed):

test/pp_api_test/test_tokenizer_capi.cc
```diff
@@ -373,6 +373,22 @@ TEST_F(MarianId2TokenTest, CombinedBugs) {
             "THIS iPhone costs PPV-mp only");
 }
 
+// Non-ASCII letters must not depend on the process locale. Hosted Linux and
+// Windows test environments commonly use the C locale, where iswalpha and
+// towupper only handle ASCII reliably.
```
kunal-vaishnavi, line 436: From Copilot:

Could we make this regression test explicitly run under the `"C"` locale, or otherwise use a locale-independent harness that reproduces the original failure?

As written, the test verifies the corrected output but relies on the host process locale to exercise the old `std::iswalpha`/`std::towupper` behavior. On environments with a UTF-8 locale, the previous implementation may already pass, allowing the regression to go undetected. If the test changes the process-global locale, please restore it safely and account for parallel test execution.

kunal-vaishnavi (dismissed):
Approving this PR. Leaving it up to you on whether you want to resolve the Copilot comment or not.

test/pp_api_test/test_tokenizer_capi.cc
mustjab: Addressed in da52082. The two Unicode regressions now run under an explicit per-thread `"C"` locale. On POSIX the guard uses `newlocale`/`uselocale`; on Windows it enables per-thread locale mode before calling `setlocale`. The prior thread locale is restored through RAII, so concurrently running tests are not affected.

I also mutation-tested the regression by temporarily restricting decoder letter recognition to ASCII. `UnicodeCaseRestoration` then failed for both Bashkir and accented Latin, and passed again after restoring the Unicode category logic. Validation: 6/6 Marian tests and 61/61 tokenizer/UGM tests pass.

kunal-vaishnavi (approved):

PR #1100, opened 2026-08-11
Align BertTokenizer offsets with returned token IDs

danielsongmicrosoft:
## Summary

- Emit exactly one offset pair for each returned token ID.
- Keep the final separator mapping when truncation removes tokens.
- Remove the second sequence's duplicate start-token mapping for paired inputs.
- Apply the same output normalization to `BertTokenizer` and `HfBertTokenizer`.
- Add Hugging Face parity coverage for truncated and paired inputs.

## Problem

Each tokenized input builds an offset list with synthetic start and separator entries. Two cases could make the flattened offset count differ from the returned token count:

- `max_length` truncation shortened `input_ids` after offsets had already been generated.
- Paired inputs contributed a second synthetic start mapping even though the combined output has only one `[CLS]` token.

The previous implementation allocated from `input_ids.size()` and then wrote every generated offset, which could write past the output buffer. Simply allocating the larger offset count avoids the write overflow but leaves offsets that do not correspond to returned token IDs.

## Fix

Normalize the per-sequence offset lists to the actual combined token output before allocation. The helper drops the duplicate paired-input start mapping, preserves the final separator mapping after truncation, and fails if the final counts still disagree.

## Validation

- Rebuilt the native Python extension from the PR source.
- `python -m pytest -q test/test_bert_tokenizer.py` - 1 passed.
- Truncated input matches Hugging Face exactly: 5 token IDs and 5 offset rows.
- Paired input matches Hugging Face exactly: 15 token IDs and 15 offset rows.
- `ruff check`, `ruff format --check`, and `git diff --check` pass.

## Compatibility

The operator signatures are unchanged. Non-truncated single-input output remains unchanged; paired offsets now align with the existing combined token sequence.


sayanshaw24:
Looks good, thanks for adding this! Could you please remove the python test and instead add a native C++ test, preferably in [`test/pp_api_test/test_tokenizer_capi.cc`](https://github.com/microsoft/onnxruntime-extensions/blob/main/test/pp_api_test/test_tokenizer_capi.cc)?.

Since this fix is about offset alignment, the test should validate `offset_mapping` behavior directly:
- `offset_mapping` row count matches output token count
- special tokens use `(0, 0)`
- paired + truncated input stays correctly aligned

sayanshaw24 (approved):

PR #1114, opened 2026-09-01
adds jinja sort filter

prathikr:
## Summary

Fixes chat-template rendering for templates that use Jinja's `sort` filter with an attribute, such as:

```jinja
message['content'] | sort(attribute='type')
```

The Minja builtins registered `dictsort`, but not `sort`. As a result, models with multimodal templates using `sort(attribute='type')` failed at runtime with `Value is not callable: null`.

This change registers a stable `sort` filter that supports sorting arrays by an `attribute` and supports `reverse=true`.

## Validation

Adds a regression test that sorts unordered image/text content blocks by `type` and verifies the rendered result is `image,text`.

### DIFF
--- a/shared/api/minja.hpp
+++ b/shared/api/minja.hpp
@@ -3788,8 +3788,10 @@ namespace minja
     }
     auto attribute = args.get<std::string>("attribute", "");
     std::stable_sort(result.begin(), result.end(), [&attribute](const Value &left, const Value &right) {
-      Value left_key = attribute.empty() ? left : left.get(attribute);
-      Value right_key = attribute.empty() ? right : right.get(attribute);
+      Value left_item = left;
+      Value right_item = right;
+      Value left_key = attribute.empty() ? left_item : left_item.get(attribute);
+      Value right_key = attribute.empty() ? right_item : right_item.get(attribute);
       return left_key < right_key;
     });
     if (args.get<bool>("reverse", false)) {
--- a/shared/api/minja.hpp
+++ b/shared/api/minja.hpp
@@ -3777,6 +3777,25 @@ namespace minja
       res.push_back(Value::array({key, value.at(key)}));
     }
     return res; }));
+    globals.set("sort", simple_function("sort", {"items", "reverse", "case_sensitive", "attribute"}, [](const std::shared_ptr<Context> &, Value &args)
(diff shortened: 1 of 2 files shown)


test/pp_api_test/test_tokenizer_chat.cc
```diff
@@ -2564,6 +2564,16 @@ TEST(OrtxTokenizerTest, MinjaUndefinedAwareFiltersPreserveExplicitNull) {
...
 
+TEST(OrtxTokenizerTest, MinjaSortSupportsAttribute) {
+  OrtxObjectPtr<OrtxTokenizer> tokenizer(OrtxCreateTokenizer, "data/phi-4-base");
+  ASSERT_EQ(tokenizer.Code(), kOrtxOK) << OrtxGetLastErrorMessage();
+
+  EXPECT_EQ(RenderMinjaExpr(tokenizer.get(),
+                            "{{ [{'type': 'text'}, {'type': 'image'}] | sort(attribute='type') | "
+                            "map(attribute='type') | join(',') }}"),
+            "image,text");
+}
```
sayanshaw24, line 2611: in addition to the Copilot comment above, can we add one regression for `sort` without `attribute` (e.g., sorting a simple list of strings or ints)? this PR validates the attribute path, but the plain `sort` path is also a common template usage and currently untested.

and could we also add a small negative test that verifies non-array input to `sort` fails with the expected error?  
that would lock in the contract for invalid input and prevent accidental behavior drift later.

shared/api/minja.hpp
```diff
@@ -3777,6 +3777,27 @@ namespace minja
       res.push_back(Value::array({key, value.at(key)}));
     }
     return res; }));
+    globals.set("sort", simple_function("sort", {"items", "reverse", "case_sensitive", "attribute"}, [](const std::shared_ptr<Context> &, Value &args)
+                                        {
+    auto & items = args.at("items");
+    if (!items.is_array()) throw std::runtime_error("sort expects an array");
```
sayanshaw24: can we make this error message include the received value/type (or a short dump)?  

`"sort expects an array"` is correct, but including the actual input will make template debugging much faster in real model traces.

sayanshaw24 (approved):

PR #1116, opened 2026-09-09
Harden tokenizer offset mappings and charsmap parsing

sayanshaw24:
### Summary

This PR fixes offset mapping generation for BPE and SentencePiece-backed BPE tokenizers when an input contains added or special tokens. It also rejects malformed precompiled charsmap replacement data before normalization. It builds on the BERT offset alignment work in #1100 by addressing the separate BPE tokenizer paths.

### Problem

`KernelBpeTokenizer::Tokenize` previously created a separate offset list for every ordinary text segment produced by splitting around added or special tokens. Each segment added its own BOS/EOS mappings, while added-token segments emitted a token ID without a corresponding mapping.

As a result, the number and ordering of offset pairs could diverge from the emitted token IDs. `KernelBpeTokenizer::Compute` allocated the offset tensor from the token output shape but copied every accumulated mapping, which could write beyond the allocated output when multiple segments were present.

### Changes

- Build one offset list per input instead of one list per ordinary text segment.
- Apply the same one-list invariant to the SentencePiece-backed BPE path.
- Add BOS and EOS offset pairs only when their corresponding token IDs are emitted.
- Add a `(0, 0)` mapping for emitted added or special tokens.
- Preserve global byte offsets across segments split around added tokens.
- Preserve the existing CLIP all-whitespace behavior while emitting matching offset pairs.
- Keep offset mappings grouped by input and zero-fill padding within each batch row.
- Add capacity checks before writing offset pairs to the output tensor.
- Reserve fixed-length output capacity for EOS and cap the CLIP all-whitespace fast path.
- Precompute UTF-8 byte offsets once per input to keep offset bookkeeping linear.
- Require nonempty precompiled charsmap replacement regions to end with a NUL terminator.
- Replace unbounded replacement-string reads with bounded terminator searches in both Unigram normalization paths.
- Avoid indexing one past the charsmap buffer when the replacement region is empty.

### Testing

Added native C++ coverage in `test/pp_api_test/test_tokenizer_impl.cc`:

- `BpeOffsetMappingMatchesTokenIds` loads the existing RoBERTa tokenizer configuration.
- Exercises an input containing an interspersed `<mask>` token.
- Verifies the `<mask>` mapping, the global byte offsets for the following token, output shape alignment, and zero-filled padding for a shorter second batch row.
- `SpmOffsetMappingMatchesTokenIds` loads the existing Phi-3 tokenizer configuration and verifies an interspersed `<|endoftext|>` token has a matching zero offset while following text retains global offsets.
- `RejectsUnterminatedCharsmapReplacement` verifies a malformed charsmap blob is rejected with `kOrtxErrorCorruptData` during tokenizer loading.
- `RejectsCharsmapWithoutReplacementRegion` verifies nonempty XCDA data without replacement bytes is rejected safely.
- `BpeFixedLengthReservesSpecialTokens` covers EOS reservation and the one-slot CLIP whitespace path.

Focused native tests pass in a Windows Debug build:

- `OrtxTokenizerTest.BpeOffsetMappingMatchesTokenIds`
- `OrtxTokenizerTest.SpmOffsetMappingMatchesTokenIds`
- `OrtxTokenizerTest.ClipTokenizer`
- `OrtxTokenizerTest.CodeGenTokenizer`
- `OrtxTokenizerTest.Phi3Tokenizer`
- `OrtxTokenizerTest.RejectsUnterminatedCharsmapReplacement`
- `OrtxTokenizerTest.SpmUgmTokenizer`
- `OrtxTokenizerTest.T5Tokenizer`
- `OrtxTokenizerTest.RejectsCharsmapWithoutReplacementRegion`
- `OrtxTokenizerTest.BpeFixedLengthReservesSpecialTokens`


operators/tokenizer/bpe_kernels.cc
```diff
@@ -609,19 +617,18 @@ std::vector<int64_t> KernelBpeTokenizer::Tokenize(ustring& input, int64_t max_le
...
-      // Add offset mappings for input in this instance to list of offset mappings for all inputs
-      offset_map.emplace_back(offset_mapping);
-    }
   }
 
   if (add_eos_token && add_special_tokens) {
     // Add EOS token to result
     res.push_back(eos_token_id_);
+    if (compute_offset_mapping) {
+      offset_mapping.emplace_back(0, 0);
```
apsonawane, line 626: When ordinary token generation has already filled max_length, this block still appends EOS, making both res and offset_mapping contain max_length + 1 elements. With an offset output, the new size check returns an internal error for an otherwise valid fixed-length truncation request; without an offset output, the token-writing loop can write past the allocated row. Could we reserve one slot for EOS while generating ordinary tokens, or otherwise apply the intended truncation policy before appending it? A regression test with fixed padding_length, EOS enabled, and an overlong input would cover this.

operators/tokenizer/bpe_kernels.cc
```diff
@@ -555,17 +562,18 @@ std::vector<int64_t> KernelBpeTokenizer::Tokenize(ustring& input, int64_t max_le
...
-
-    if (compute_offset_mapping) {
-      if (add_bos_token && add_special_tokens) {
-        // Add offset mapping for BOS token
-        offset_mapping.push_back(std::make_pair(0, 0));
-      }
+    if (compute_offset_mapping && !seg_id.first.empty()) {
+      const size_t prefix_length = static_cast<size_t>(seg_id.first.data() - input.data());
+      std::string prefix;
+      ustring::ToUTF8Into(std::u32string_view(input.data(), prefix_length), prefix);
```
apsonawane, line 575: This converts the entire prefix from the start of input for every ordinary segment. With many interspersed added/special tokens, progressively longer prefixes are re-encoded and temporary strings are repeatedly allocated, making offset bookkeeping O(n²). Since tokenizer input may be user-controlled, could we instead maintain a running UTF-8 byte position or precompute UTF-8 byte offsets once per input? The same pattern also appears in SpmTokenize.

operators/tokenizer/bpe_kernels.cc
```diff
@@ -688,9 +703,11 @@ std::vector<int64_t> KernelBpeTokenizer::SpmTokenize(ustring& input, int64_t max
     spm_splitter.Set(std::u32string_view(ustr));
 
     size_t offset = 0;
-    OffsetMappingType offset_mapping;
-    if (compute_offset_mapping && add_bos_token_.value_or(true) && add_special_tokens) {
-      offset_mapping.push_back(std::make_pair(0, 0));  // initial offset
+    if (compute_offset_mapping && !seg_id.first.empty()) {
+      const size_t prefix_length = static_cast<size_t>(seg_id.first.data() - input.data());
```
apsonawane, line 707: This has the same quadratic prefix re-encoding issue as the BPE path. Each segment converts the complete preceding prefix to UTF-8 again. Please use shared running/precomputed UTF-8 byte offsets for both paths.

operators/tokenizer/ugm_kernels.hpp
```diff
@@ -114,6 +114,11 @@ struct SpmUgmTokenizer {
       // replacement strings for prefixes matched by the XCDA.
       prefix_replacements_ = reinterpret_cast<const char*>(&charsmap_data_[charsmap_offset]);
       prefix_replacements_size_ = charsmap_data_.size() - charsmap_offset;
+      if ((xcda_array_size_ > 0 && prefix_replacements_size_ == 0) ||
```
apsonawane, line 117: The empty-region case is checked here, but the preceding assignment evaluates &charsmap_data_[charsmap_offset] first. When charsmap_offset == charsmap_data_.size(), operator[] is out of bounds, so malformed input can trigger undefined behavior before this check rejects it. Please construct the pointer with charsmap_data_.data() + charsmap_offset instead, and add a test with a nonempty XCDA section followed by zero replacement bytes.

test/pp_api_test/test_tokenizer_impl.cc
```diff
@@ -946,6 +947,103 @@ TEST(OrtxTokenizerProfileTest, DISABLED_MemoryUsage) {
...
+  const auto pad_id = static_cast<int64_t>(tokenizer.GetTokenId("<pad>"));
+  bool found_mask = false;
+  bool found_world = false;
+  for (int64_t i = 0; i < row_length; ++i) {
+    if (ids[i] == mask_id) {
+      found_mask = true;
+      EXPECT_EQ(offsets[i * 2], 0);
+      EXPECT_EQ(offsets[i * 2 + 1], 0);
+    }
+    if (offsets[i * 2] == 13 && offsets[i * 2 + 1] == 18) {
```
apsonawane, line 984: This assertion accepts any token carrying (13, 18) as evidence that world is aligned. A shifted mapping could therefore still pass. Could this check also verify the token ID at ids[i], or preferably assert the complete expected non-padding ID/offset sequence?

test/pp_api_test/test_tokenizer_impl.cc
```diff
@@ -946,6 +947,103 @@ TEST(OrtxTokenizerProfileTest, DISABLED_MemoryUsage) {
...
+  bool found_special = false;
+  bool found_following_text = false;
+  for (int64_t i = 0; i < input_ids.NumberOfElement(); ++i) {
+    if (ids[i] == special_id) {
+      found_special = true;
+      EXPECT_EQ(offsets[i * 2], 0);
+      EXPECT_EQ(offsets[i * 2 + 1], 0);
+      continue;
+    }
+    if (found_special && offsets[i * 2] >= 18) {
```
apsonawane, line 1039: This only proves that some token after the special token has a start offset of at least 18; it does not prove that the mapping remains aligned with the expected world token. Could the test tie the expected offset to the corresponding token ID, or assert the complete ID/offset sequence? Including multibyte Unicode around the special token would also verify that these remain UTF-8 byte offsets.

operators/tokenizer/bpe_kernels.cc
sayanshaw24: Content generation now reserves one slot for EOS when enabled, and EOS is appended only when capacity remains.

operators/tokenizer/bpe_kernels.cc
sayanshaw24: UTF-8 byte offsets are now precomputed once per input, so each segment offset uses an O(1) lookup instead of re-encoding progressively longer prefixes.

operators/tokenizer/bpe_kernels.cc
sayanshaw24: Updated the SPM path to use the same precomputed UTF-8 byte-offset table as the regular BPE path.

operators/tokenizer/ugm_kernels.hpp
sayanshaw24: The pointer now uses charsmap_data_.data() + charsmap_offset, which safely permits a one-past-end pointer without dereferencing it. I also added a native test for nonempty XCDA data with no replacement region.

test/pp_api_test/test_tokenizer_impl.cc
sayanshaw24: The test now requires the expected Ġworld token ID with the exact (13, 18) offset, instead of accepting any token carrying that offset.

test/pp_api_test/test_tokenizer_impl.cc
sayanshaw24: The test now asserts the complete expected token-ID and offset sequences, including multibyte é, the interspersed special token, and the following ▁world token.

apsonawane (approved):

PR #1117, opened 2026-09-11
Preserve raw tool schemas in chat templates

tianleiwu:
## Description

Preserve OpenAI tool schemas for chat templates that serialize tool definitions directly with `tojson`. This prevents lossy normalization from dropping constraints such as `required`, `enum`, `items`, nested properties, and nullable types before they reach Qwen-style prompts.

### Chat Template Handling

| File | Change |
|------|--------|
| `shared/api/chat_template.cc` | Detect templates that consume raw tool objects through `tool.function`, `tool \| tojson`, or `tools \| tojson`, and bypass lossy normalization for those templates. |

### Tests

| File | Change |
|------|--------|
| `test/pp_api_test/test_tokenizer_chat.cc` | Update Qwen expected prompts to retain OpenAI schemas and add regression coverage for required fields, enum constraints, and array item schemas. |

## Testing

- `cmake --build out/Linux/RelWithDebInfo --target pp_api_test --parallel 4`
- `out/Linux/RelWithDebInfo/bin/pp_api_test --gtest_brief=1`
- Result: 138 enabled tests passed; 2 existing tests remain disabled.

## Motivation and Context

This ports the ONNX Runtime Extensions portion of [microsoft/onnxruntime-genai#2537](https://github.com/microsoft/onnxruntime-genai/pull/2537). `NormalizeTools()` is intended for templates using the flat Phi-4/Minja representation, but Qwen-style templates serialize each supplied OpenAI tool object verbatim. Normalizing those objects changes the model prompt and can cause generated arguments to violate the caller's schema.

## Checklist

- [x] Tests added/updated
- [x] No breaking API changes
- [x] Documentation not required


kunal-vaishnavi (dismissed):

shared/api/chat_template.cc
```diff
@@ -326,6 +328,37 @@ static json NormalizeTools(const char* tools_str) {
...
+/*
+ * Reports whether a chat template consumes tool definitions in their raw OpenAI shape.
+ *
+ * NormalizeTools() rewrites tools into the flat Phi-4/Minja shape, which is lossy: it
+ * unwraps {"type":"function","function":{...}}, drops "required", "enum", "items" and any
+ * nested property schema, and renames the "string" type to "str". That is only safe for
+ * templates written against the flat shape. Two families need the raw objects instead:
+ *
+ *   - Harmony/GPT-OSS templates, which reach into `tool.function` themselves.
+ *   - Qwen-style templates, which serialize each tool verbatim with `tool | tojson`.
```
kunal-vaishnavi, line 340: Do older Qwen models also use `tool | tojson` or only newer Qwen models? We want to ensure that tool calling works across the Qwen model family (Qwen-2.5, Qwen-3, Qwen-3.5, Qwen-3.6, Qwen-3.8, etc).

tianleiwu: I checked the official tokenizer configs across the family:

- `Qwen2-7B-Instruct` has no native tools block in its chat template, so this change does not alter its existing path.
- `Qwen2.5-7B-Instruct` and `Qwen3-8B` both use the executable `{% for tool in tools %} ... {{ tool | tojson }}` pattern.
- Every official Qwen3.5, Qwen3.6, and Qwen3.8 checkpoint currently listed under the Qwen organization that I queried uses the same `tool | tojson` pattern.
- `Qwen3-Coder-30B-A3B-Instruct` uses `tool.function` instead, which is covered by the existing raw-tool detection; Qwen3-VL and Qwen2.5-Coder use `tool | tojson`.

The detection is intentionally based on template syntax rather than model name/version. That covers these families and future/custom templates using either raw-tool pattern, while older or custom flat-format templates without those patterns continue through `NormalizeTools()`.

I also added `Qwen3PreservesRawToolSchema` in commit `7d226bd`, using the repository's real Qwen3 template fixture to verify that the OpenAI wrapper plus `required`, `enum`, and `items` survive rendering. The full native suite now passes 138/138 enabled tests.

kunal-vaishnavi (approved):

PR #1120, opened 2026-09-15
Support Jinja is true and is false predicates in chat templates

baijumeswani:
## Summary
 
 Add strict Jinja boolean identity predicates to the Minja chat-template evaluator:
 
 - `is true`
 - `is false`
 - `is not true`
 - `is not false`
 
 These predicates match only actual boolean values. Integers, strings, collections, null, and undefined values do not match, preserving Jinja identity semantics rather than applying truthiness.
 
 ## Motivation
 
 Qwen3.8 chat templates use expressions such as:
 
 ```jinja
 {% if enable_thinking is undefined or enable_thinking is true %}

ORT Extensions previously rejected this expression with:

 Unknown type for 'is' operator: true

As a result, applications could not explicitly pass  {"enable_thinking": false}  through  chat_template_kwargs .


tianleiwu (approved):
Non-blocking:

`x is True` still throws, because parseIdentifier accepts True as a name and there's no entry for it. That's actually consistent with Jinja (its test registry has only lowercase true/false), so leaving it as-is is defensible — but the error message is opaque if a template ever hits it.

Related gap. Jinja's general identity test is `sameas` is still unsupported. Out of scope here, but it's the same dispatch table if it ever comes up.

baijumeswani:
This makes sense. I'll leave True and False unsupported.
Will add a more complete support in a follow up PR (including sameas, divisibleby and so on that jinja supports).

kunal-vaishnavi (approved):

PR #1121, opened 2026-09-17
Support Jinja  sameas  singleton predicates in chat templates

baijumeswani:
Add targeted Minja support for Jinja  sameas  expressions used by chat templates:

 -  value is sameas true|false|none 
 -  value is not sameas true|false|none 

The implementation uses strict singleton semantics rather than truthiness or structural equality. Booleans match only the corresponding Boolean value, explicit null matches only  none , and undefined or other value types do not match. Callable values are evaluated as non-singletons instead of being lifted into callable results.

The parser intentionally accepts only bare `true`, `false`, and `none` operands. General parameterized Jinja tests, arbitrary object identity, and parenthesized `sameas(...)` syntax remain out of scope.

### Context

Chat templates commonly use  sameas  to require an exact Boolean or null value rather than generic truthiness. For example:

```
 {%- if enable_thinking is defined and enable_thinking is sameas true %}
 <think>
 {%- endif %}
```

A template with a default thinking policy may look like:

```
 {%- if enable_thinking is undefined or enable_thinking is sameas true %}
 {{- '<think>\n' }}
 {%- else %}
 {{- '<answer>\n' }}
 {%- endif %}
```

Negation can distinguish an explicit opt-out:

```
 {%- if enable_thinking is not sameas false %}
 {{- '<think>\n' }}
 {%- endif %}
```

The null singleton can be checked similarly:

```
 {%- if tool_choice is sameas none %}
 {{- 'No tool was selected.' }}
 {%- endif %}
```

These are deliberately bare-operand forms supported by this change. For example, use `value is sameas true`, not `value is sameas(true)`.


tianleiwu (approved):
Reviewed the parser/evaluator changes and focused regression coverage. The implementation preserves strict Boolean/null singleton semantics, keeps undefined distinct from null, and correctly treats callable values as non-singletons. I also rebuilt `pp_api_test` and ran all five affected Minja tests successfully. No blocking or non-blocking findings.

sayanshaw24 (approved):

PR #1123, opened 2026-09-17
Adding word level timestamps to Nemotron Streaming Pipeline Extensions

themason2011:
## Summary

Adds metadata-aware incremental detokenization to support word-level timestamp alignment for streaming speech models such as Nemotron.

The new API emits completed words alongside their contributing token spans while preserving the exact decoded text, including whitespace and punctuation. Existing `OrtxDetokenizeCached` behavior remains unchanged.

This work unblocks [onnxruntime-genai#2591](https://github.com/microsoft/onnxruntime-genai/pull/2591), which combines these token spans with acoustic timing information to produce streaming word- and segment-level timestamps across the C, C#, and Python APIs.

## Changes

- Adds `OrtxDetokenizeCachedWithMetadata`, which returns:
  - The decoded text fragment.
  - Zero or more completed word events.
  - Half-open token spans (`[start_token_index, stop_token_index)`) for each word.
  - The earliest token index still required by pending decoder output.
- Adds `OrtxFinalizeDetokenizeCachedWithMetadata` to emit the final pending word at the end of a stream without producing duplicate events.
- Introduces tokenizer-aware streaming word grouping for:
  - SentencePiece tokenizers.
  - Prefix- and suffix-boundary BPE tokenizers.
  - Unigram tokenizers.
  - Special tokens, punctuation, whitespace, and buffered byte-fallback output.
- Preserves exact transcript reconstruction from emitted word events, including leading/trailing whitespace and punctuation.
- Supports multiple completed words from a single token.
- Prevents mixing legacy text-only and metadata-aware decoding on the same detokenizer cache, avoiding inconsistent decoder state.
- Exports the new C API entry points on Windows.

## API behavior

Metadata returned by each decode call is event-based rather than cumulative. A call may return no completed words or multiple completed words. Returned data is owned by the detokenizer cache and remains valid until the next operation on that cache.

Existing users of `OrtxDetokenizeCached` are unaffected.

## Validation

Adds coverage for:

- Identical decoded text between legacy and metadata-aware paths.
- Exact transcript reconstruction and token-span tracking.
- Finalization of trailing words.
- Rejection of mixed cache modes.
- SentencePiece and Unigram tokenizers.
- Multiple words completed by one token.
- Empty buffered decoder output and byte-fallback spans.
- Whitespace and punctuation grouping behavior.


shared/api/c_api_tokenizer.cc
```diff
@@ -485,16 +492,84 @@ extError_t ORTX_API_CALL OrtxDetokenizeCached(const OrtxTokenizer* tokenizer, Or
 
   // If skip_special_tokens option exists, use its value, otherwise use default (true)
   bool skip_special_tokens = ParseBoolOption(token_ptr->GetOption("skip_special_tokens"), true);
+  TokenizerWordPieceInfo piece_info;
```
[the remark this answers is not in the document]
themason2011: Good catch. I added an explicit mode to DetokenizerCache. The first cached detokenization call locks the cache to either text-only or metadata mode. Calling the other API afterward—including metadata finalization on a text-only cache—now returns kOrtxErrorInvalidArgument and instructs the caller to destroy and recreate the cache to switch modes. The mismatch is rejected before consuming the token. Tests cover switching in both directions.

shared/api/tokenizer_word_grouping.cc
```diff
@@ -0,0 +1,128 @@
...
+
+void TokenizerWordGroupingState::Finalize() {
+  completed_words_.clear();
+  if (pending_delimiter_.active) {
+    Append(pending_word_, pending_delimiter_.text,
+           pending_delimiter_.start_token_index, pending_delimiter_.stop_token_index);
+    pending_delimiter_ = {};
+  }
+  CompleteWord();
+}
```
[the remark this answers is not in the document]
themason2011: Good catch. Finalize() now clears the buffered output start index because trailing empty decoder output cannot contribute to any future text after the stream is finalized. I added a regression test verifying that FirstPendingTokenIndex() advances past that token.

include/ortx_tokenizer.h
```diff
@@ -214,6 +230,27 @@ extError_t ORTX_API_CALL OrtxDetokenize1D(const OrtxTokenizer* tokenizer, const
 extError_t ORTX_API_CALL OrtxDetokenizeCached(const OrtxTokenizer* tokenizer, OrtxDetokenizerCache* cache,
                                               extTokenId_t next_id, const char** text_out);
 
+/** \brief Detokenize one token and return completed word events.
+ *
+ * The decoded fragment is identical to OrtxDetokenizeCached. Word events preserve exact decoded
+ * whitespace and punctuation and use half-open token-index spans. The event array may be empty or
+ * contain multiple words when one token spans multiple boundaries. Returned words are owned by the
+ * cache and remain valid until the next operation on that cache.
+ */
```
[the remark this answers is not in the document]
themason2011: Documented the enforced cache-mode contract. The first cached detokenization call selects text-only or metadata mode; mixing the APIs returns an error, and callers must destroy and recreate the cache before switching modes.

operators/tokenizer/ugm_kernels.hpp
```diff
@@ -763,10 +763,17 @@ class SpmUgmDecoder {
...
     tokenizer_add_space_prefix_ = tokenizer.tokenizer_add_space_prefix_;
+    tokenizer_treat_whitespace_as_suffix_ = tokenizer.tokenizer_treat_whitespace_as_suffix_;
     case_encoding_ = tokenizer.case_encoder_ != nullptr;
     return {};
   }
 
+  std::string_view GetEncodedPiece(extTokenId_t id) const {
+    return static_cast<size_t>(id) < vocab_.size() ? std::string_view{vocab_[id]} : std::string_view{};
+  }
+  bool IsSpecialToken(extTokenId_t id) const { return special_token_ids_.count(id) != 0; }
```
sayanshaw24, line 774: why do we need this? we recently added a special token API in GenAI, could this possible be done there and handled in your GenAI PR: https://github.com/microsoft/onnxruntime-genai/pull/2591

shared/api/c_api_tokenizer.cc
```diff
@@ -11,13 +11,26 @@
 
 using namespace ort_extensions;
 
+struct StreamingWordGroupingCache {
+  TokenizerWordGroupingState state;
+  std::vector<OrtxDetokenizedWord> word_views;
+};
+
+enum class DetokenizerCacheMode {
```
sayanshaw24, line 19: let's put these enums/structs elsewhere (ideally existing audio files) and keep the shared C API clean.

sayanshaw24:
is this thread safe? Extensions tokenizers are currently thread safe and we want to keep it that way.

cmake/ext_python.cmake
```diff
@@ -31,7 +31,8 @@ include(pybind11)
 target_include_directories(extensions_pydll PRIVATE
   ${pybind11_INCLUDE_DIRS}
   $<TARGET_PROPERTY:Python3::Module,INTERFACE_INCLUDE_DIRECTORIES>
-  $<TARGET_PROPERTY:ocos_operators,INTERFACE_INCLUDE_DIRECTORIES>)
+  $<TARGET_PROPERTY:ocos_operators,INTERFACE_INCLUDE_DIRECTORIES>
+  ${PROJECT_SOURCE_DIR}/shared/api)
```
themason2011: I thought this got removed because it was fixed with an include statement in the right location, we shouldn't be adding the /shared/api folder generally

sayanshaw24:
Can we use `OrtxUpdateTokenizerOptions` + setting/updating options at tokenization time for this to avoid introducing a new Detokenizer API? You can look at the recent chat template kwargs addition as a reference - here is the original PR for it that introduced a new `OrtxApplyChatTemplateWithOptions` API here: https://github.com/microsoft/onnxruntime-extensions/pull/1102, and the updated one after my requesting of using `OrtxUpdateTokenizerOptions` in this PR: https://github.com/microsoft/onnxruntime-extensions/pull/1108

include/ortx_tokenizer.h
```diff
@@ -214,6 +234,31 @@ extError_t ORTX_API_CALL OrtxDetokenize1D(const OrtxTokenizer* tokenizer, const
...
+ * The decoded fragment is identical to OrtxDetokenizeCached. Word events preserve exact decoded
+ * whitespace and punctuation and use half-open token-index spans. The event array may be empty or
+ * contain multiple words when one token spans multiple boundaries. Returned words are owned by the
+ * cache and remain valid until the next operation on that cache.
+ *
+ * The first call to OrtxDetokenizeCached or OrtxDetokenizeCachedWithMetadata locks the cache
+ * to that mode. The two functions cannot be mixed on the same cache. Destroy and recreate the
+ * cache to switch modes.
+ */
+extError_t ORTX_API_CALL OrtxDetokenizeCachedWithMetadata(const OrtxTokenizer* tokenizer,
```
themason2011, line 248: Move OrtxDetokenizeMetadata into the Tokenizer Options API (OrtxTokenizer can allocate it) so we don't have to have a separate helper method like this, we can remove it (and OrtxFinalizeDetokenizeCachedWithMetadata) and just call OrtxDetokenizeCached, track the Metadata through the Tokenizer Options API, and buffer it there. Also, make sure to retain thread-safe abilities. Sayan left a separate comment for this, link these together.

shared/api/c_api_tokenizer.cc
themason2011: Addressed in 11fa9f6. Moved DetokenizerCache, its mode enum/handling, and timestamp-state management into shared/api/tokenizer_stream_state.h/.cc. The grouping state, C word views, and timestamp payload are encapsulated in a lazily allocated per-cache TimestampState; the C API now handles validation, option retrieval, decoder calls, and returning the borrowed result.

I used a tokenizer-specific helper rather than the existing Nemo audio files because this operates on decoded token text, not PCM/mel features, and should remain available in tokenizer-only builds. WordBoundaryStyle and TokenizerWordPieceInfo now live in operators/tokenizer/tokenizer_common.h, while tokenizer_word_grouping retains the grouping algorithm. BPE and Unigram each provide GetWordPieceInfo(), and TokenizerImpl just dispatches to them.

Validated the focused metadata/grouping tests and the shared/Python extension builds.

cmake/ext_python.cmake
themason2011: Confirmed this workaround is no longer in the PR: cmake/ext_python.cmake has no diff against the PR base. The decoder headers now include tokenizer_common.h for the shared word-piece types, so they no longer need the shared/api grouping header either. No broad shared/api include-path addition is needed for the Python target; extensions_pydll builds successfully.

include/ortx_tokenizer.h
themason2011: Updating this earlier proposal after separating configuration from per-stream output. Tracking is now enabled through OrtxUpdateTokenizerOptions using track_timestamp_metadata (default false), captured at the first metadata decode for each cache.

I am retaining the metadata decode/finalize entry points in this revision. The options API accepts string configuration, not typed output, and placing mutable metadata on a shared OrtxTokenizer would mix results from independent streams. Results instead belong to each detokenizer cache. OrtxMetadata is a typed C-compatible wrapper with timestampMetadata, allowing future metadata members without another decode method. Finalization flushes a pending word without injecting a synthetic token.

This addresses configuration through options, but does not fully implement the original suggestion to remove the new decode API. Keeping the existing ordinary decode signature while exposing metadata would still require a separate cache-result retrieval contract. Related discussion: https://github.com/microsoft/onnxruntime-extensions/pull/1123#issuecomment-5736920311 and https://github.com/microsoft/onnxruntime-extensions/pull/1123#issuecomment-5736734210.

operators/tokenizer/ugm_kernels.hpp
themason2011: The standalone IsSpecialToken() getter is now removed in 11fa9f6. Its existing classification logic is part of GetWordPieceInfo() in the Unigram decoder, alongside the equivalent BPE implementation.

The classification is still needed here: empty decoded output can mean either a skipped special token or an ordinary token whose bytes are buffered. The latter must retain its contributing token index for word/timestamp alignment; a skipped leading special token must not become the start of that word. The named BOS/EOS/PAD/BOT/EOT/BOR/EOR getters in GenAI are not an equivalent arbitrary-ID query against the complete special-token set in the decoder. Also, Extensions cannot call back into GenAI without reversing the dependency.

Added SkippedSpecialTokenDoesNotExtendBufferedWordSpan to verify that a skipped special token is excluded from the start of the word span while ordinary buffered byte tokens remain included. No special-token decoding behavior was changed.

themason2011:
@sayanshaw24 Regarding thread safety: https://github.com/microsoft/onnxruntime-extensions/pull/1123#issuecomment-5736734210

The new tracking state is per detokenizer cache, not on the shared tokenizer. Independent caches can decode concurrently using one tokenizer. Tokenizer option reads/updates use the existing mutex; track_timestamp_metadata is captured on the first metadata decode for each cache, so later updates do not change an in-flight stream. An existing but unused cache sees the option value at its first metadata decode, not at allocation.

Operations on the same cache, and reads of its borrowed text/metadata, must be externally serialized. Returned pointers expire on the next cache operation or disposal; this does not promise concurrent access to a single cache. The public header now states this explicitly.

Coverage includes independent concurrent caches, disabled tracking, and ConcurrentOptionUpdatesPreserveStreamTrackingSettings (two streams with different captured settings while another thread updates options). All 17 focused Extensions tests passed. This is functional concurrency coverage, not a ThreadSanitizer verification.

themason2011:
@sayanshaw24 Regarding the tokenizer-options suggestion: https://github.com/microsoft/onnxruntime-extensions/pull/1123#issuecomment-5736920311

Tracking now uses OrtxUpdateTokenizerOptions with track_timestamp_metadata (default false). I retained the metadata decode/finalize APIs because configuration and per-stream output have different ownership requirements: options are string-valued settings on the shared tokenizer, whereas word records and pending-token indices belong to an individual decoding cache.

Unlike chat_template_kwargs, this feature needs to return additional typed data, not just change how existing inputs are processed. Putting output buffers on the shared tokenizer would mix concurrent streams. Keeping the ordinary decode signature unchanged would instead require a cache-result accessor and an explicit flush/finalization mechanism.

The current compromise is one typed OrtxMetadata wrapper returned by the metadata path, with a nullable timestampMetadata member. Future metadata can extend the wrapper without new decode methods. Disabled timestamp tracking leaves that member null and skips word-piece extraction, grouping, and timestamp-view updates. Finalization emits pending word records without injecting a synthetic token.

This is a justification for retaining the API additions, not a claim that the original no-new-decode-API suggestion is fully implemented. I also updated the earlier inline proposal here: https://github.com/microsoft/onnxruntime-extensions/pull/1123#discussion_r4087871261.

